PR #456, opened 2022-01-20
Leila tb shine

Leila-Uwais:
Changing treatment for children with minimal TB from 6-months to 4-months


tbhallett:
@tdm32  -- removing this PR from our project tracker for now. Let me know if I should reinstate / put in queue to review, or if we should close it.

PR #821, opened 2023-01-23
Fixed incidence TB model with updated HIV 

tdm32:
This is the current PR that replaces the closed PR #752 
It uses a fixed TB incidence model with treatment coverage as a modifying factor for incidence

image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

### DIFF
--- a/src/scripts/hiv/projections_jan2023/calibration_script.py
+++ b/src/scripts/hiv/projections_jan2023/calibration_script.py
@@ -39,13 +39,13 @@ import pandas as pd
 from tlo import Date, logging
 from tlo.methods import (
     demography,
-    hiv_tb_calibration,
     enhanced_lifestyle,
     epi,
     healthburden,
     healthseekingbehaviour,
     healthsystem,
     hiv,
+    hiv_tb_calibration,
     simplified_births,
     symptommanager,
     tb,


src/tlo/methods/tb.py
```diff
@@ -1690,14 +1753,16 @@ def apply(self, person_id, squeeze_factor):
...
-        self.sim.modules["HealthSystem"].schedule_hsi_event(
-            hsi_event=hiv.HSI_Hiv_TestAndRefer(
-                person_id=person_id, module=self.sim.modules["Hiv"], referred_from='Tb'
-            ),
-            priority=1,
-            topen=self.sim.date,
-            tclose=None,
-        )
+        # do not run if already HIV diagnosed or had test in last week
+        if not person["hv_diagnosed"] or (person["hv_last_test_date"] >= (now - DateOffset(days=7))):
```
EvaJanouskova, line 1757: Do I understand this correctly that if any of the conditions, 1) being already HIV diagnosed or 2) being tested in last week, is satisfied; the person shouldn't be referred for HIV testing?

According to the code, if one of the conditions is true, than the person is referred for HIV testing. Hence, they are referred for testing if:
- diagnosed but more than a week ago,
- not diagnosed yet but the test was performed less than a week ago,
- not diagnosed and test was performed more than a week ago.

If the person should only be referred for testing when not diagnosed **and** the test was performed more than a week ago, then there needs to be `and` instead of `or` between the conditions. Or am I misunderstanding the comment above?

EvaJanouskova (commented):
Hi @tdm32 , while reading your updates related to equipment, I noticed a comment which is not in agreement with the following code, if I'm not misunderstanding it.

src/tlo/methods/tb.py
tdm32: Hi @EvaJanouskova , the comment is correct. The person is referred for testing if they're not already diagnosed or if they had a test in the last week. This allows repeat testing for known cases but prevents retesting on the same week through multiple routes which was occurring. 

So they can (re-)test if they:
- are diagnosed but it was >1 week ago
- are not diagnosed but their last test was >1 week ago

EvaJanouskova: Hi @tdm32, But the comment says "**do not run** if already HIV diagnosed or had test in last week" -> which should be interpreted as "**do run** if not HIV diagnosed and had not test in last week"  which is not in agreement with the following code.

But from what you write, the code is correct, but the comment above should then say "**do not run** if already HIV diagnosed **and** had test in last week"

PR #826, opened 2023-01-24
Update the APPT_FOOTPRINT of FamPlan

BinglingICL:
In this small PR, we try to reduce the use of FamPlan from TLO output by changing the appt footprint depending on the contraception method and the number of times run.

Subsequent scenario run experiments are to be conducted.

Builds on:
* #818

### DIFF
--- a/tests/test_contraception.py
+++ b/tests/test_contraception.py
@@ -474,7 +474,7 @@ def test_record_of_appt_footprint_for_switching_to_methods(tmpdir, seed):
         )
 
     assert is_list_longer_than_length_of_one_and_with_first_element_nonblank_and_subsequent_blank(
-        get_appt_footprints(switch_from='not_using', switch_to='pill', consumables_available=False)
+        get_appt_footprints(switch_from='not_using', switch_to='female_sterilization', consumables_available=False)
     )
     assert is_list_longer_than_length_of_one_and_with_first_element_nonblank_and_subsequent_blank(
         get_appt_footprints(switch_from='not_using', switch_to='pill', consumables_available=False)


src/tlo/methods/contraception.py
```diff
@@ -958,14 +961,29 @@ def __init__(self, module, person_id, new_contraceptive):
...
         self.TREATMENT_ID = "Contraception_Routine"
-        self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({'FamPlan': 1})
         self.ACCEPTED_FACILITY_LEVEL = _facility_level
 
+    @property
+    def EXPECTED_APPT_FOOTPRINT(self):
+        """Return the expected appt footprint based on contraception method and whether the HSI has been rescheduled."""
+        person_id = self.target
+        current_method = self.sim.population.props.loc[person_id].co_contraception
+        if self._number_of_times_run > 0:  # if it is to re-schedule due to unavailable consumables
```
EvaJanouskova, line 971: This will be positive not only if the appt is to re-schedule but also if it was a rescheduled appt. But in that case, the contraceptive could have been administrated this time, hence the appt footprint shouldn't be empty. Am I right?

EvaJanouskova (commented):
Hi @tbhallett and @BinglingICL, I think we overlooked this:

src/tlo/methods/contraception.py
EvaJanouskova: There is also smt going on with the minor surgery appt footprint which I can't explain. When I add a condition to add equipment `if _new_contraceptive == 'female_sterilization'` this equipment is then in logs never logged along with `['MinorSurg', 1]` appt footprint, but only with empty footprint, which may be just because the above. 

BUT, it is logged with `['MinorSurg', 1]` appt footprint when the equipment for sterilisation is not added, which can only happen if consumables for sterilisation are not available, or different method is used. But according of  `def EXPECTED_APPT_FOOTPRINT` I would say that in such cases the footprint should be never MinorSurg. I have no idea how this could be happening.
If I set some of the equipment for sterilisation as unavailable, so these hsi will not even start to run, the logges of equipment used with sterilisation (with empty footprint) disappear as expected, however the logs with `['MinorSurg', 1]` appt footprint when the equipment for sterilisation is not added, are still there.

EvaJanouskova: Btw, it also means that with this logic, for example, Minor Surgery will be logged for the first attempt to get female sterilisation, but if it is never performed (I think there is only limited number of attempts, isn't it?), it was still logged as minor surgery once event that no surgery was performed at the end. This applies to all other appts with same logic for appt_footprint.

PR #949, opened 2023-04-20
Switch to using Conda lock files to specify requirements and update dependencies

matt-graham:
Switches from using `pip-compile` and `requirements.txt` files to specify Python dependencies to be installed using `pip` to using [`conda-lock`](https://github.com/conda/conda-lock) to generate lock files specifying (platform specific) Conda environments with both Python and non-Python dependencies.

Also currently set to update dependencies to Python 3.11 and Pandas 2.0.0 - we could potentially pin to the current Python (v3.8) / Pandas (v1.2.2) versions instead to split the change over how we're handling dependencies and the actual updating of dependencies.


matt-graham:
I don't think we want to go forward with this one any more so closing.

PR #974, opened 2023-05-22
contraception health impacts: scenario

EvaJanouskova:
**Scenario to run some first jobs:**

All loggings, modules, and draw_parameters as in Joe's [baseline_scenario](https://github.com/UCL/TLOmodel/blob/301e9419ebfa83ee220002cc1fb58513dcd715b2/src/scripts/maternal_perinatal_analyses/scenario_files/baseline_scenario.py).

Except, I use `cons_availability='all'` for now because we don't have availability data for most of the consumables we use for contraceptives and we assume that all consumables are essential, thus it won't be very realistic yet anyway.
_________________________________________
_We may use the default availability once the issue #975 is resolved._

_note: The description at the top of the scenario file is not updated._


EvaJanouskova:
This is too old version. I created a new one from current master. See #1099

Issue #686, opened 2022-08-09
Tracking profiling run results


matt-graham:
We would like to be able to track how the the timings measured in profiling runs of the `src/scripts/profiling/scale_run.py` script changes as new pull-requests are merged in. This would help identifying when PRs lead to performance regressions and allow us to be more proactive in fixing performance bottlenecks.

Ideally this should be as automated using GitHub Actions workflows. Triggering the workflow on pushes to `master` would give the most detail in terms of giving a direct measurement of the performance differences arising from a particular PR, but when lots of PRs are going in could potentially create a large backlog of profiling runs, so an alternative would be to run on a schedule (for example nightly) using the `cron` event.  It would probably be worth also allowing triggering either using the `workflow_dispatch` event or using the comment-triggered workflow functionality to allow manually triggering in PRs that it is thought might have a significant effect on performance before merging.

Key questions to be resolved are what profiling outputs we want to track (for example at what level of granularity, using which profiling tool) and how we want to visualize the outputs. One option would be to [save the profiler output as a workflow artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). While this would be useful in allowing access to the raw profiling data, the only option for accessing workflow artifacts appears to be downloading the artifact as a compressed zip file so this is not necessarily itself that useful for visualizing the output. One option for visualizing the profiling results would be to use the [GitHub Actions job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) which allows using Markdown to produce customized output showed on the job summary page. Another option would be to output the profiling results to HTML files and then [deploy these to either a GitHub Pages site](https://github.com/actions/deploy-pages) or potentially to [a static site on Azure storage](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-static-site-github-actions?tabs=userlevel).

## Potentially useful links
 
The [airspeed velocity package](https://asv.readthedocs.io/en/stable/) allows tracking the results of benchmarks of Python packages overtime and visualizing the results as plots in a web interface. While focused on suites of benchmarks it does also have support for [running single benchmarks with profiling](https://asv.readthedocs.io/en/stable/commands.html#cmd-asv-profile).

[htmlpreview](https://github.com/htmlpreview/htmlpreview.github.com) allows directly previewing HTML files in a GitHub repository as _GitHub forces them to use the "text/plain" content-type, so they cannot be interpreted_

tamuri:
[...]

PR #1012, opened 2023-06-29
Profiling workflow on `workflow_dispatch`

willGraham01:
## Short Description

Provides the "skeleton" for implementation of profile tracking of runs across the repository history (#686).

Adds the `run_profiling.yml` workflow which profiles the `scale_run.py` script and exports the resulting `.pyisession` file to the [`TLOmodel-profiling` repository](https://github.com/UCL/TLOmodel-profiling) - repo is under the UCL organisation so everyone should be able to see it, but I can fix that if this isn't the case. Workflow runs on self-hosted runners and uses a `tox` environment to run the profiling script.

## Full Changes

- Adds `.github/workflows/run_profiling.yml` workflow.
  - This workflow can be requested on dispatch
  - This workflow creates results in a tree structure: `year/month/day/HHMM/{files.out}` on the [`TLOmodel-profiling` repository](https://github.com/UCL/TLOmodel-profiling). Deployment targets the `results` branch (so that in future we can trigger rendering on `push`es to this branch).

- `src/scripts/profiling` directory: 
  - `scale_run.py` has the "main" refactored so that it can be imported from the profiling script
  - `_paths.py` file to fix some important paths to avoid filesystem errors
  - `_parameters.py` file to reliably the track the parameters that should be passed to the models that we are profiling. **NOTE:** For the purposes of testing new features, the profiling only runs a month-long simulation. This can be edited once we're happy with the workflow and how the outputs are presented/moved to the output repo.
  - `run_profiling.py`; new script that can be run from the command-line. Profiles the `scale_run` model and writes the outputs to either/both HTML and JSON.

- `vscode`'s linter has also zealously run over a couple of other files like `shared.py`, which is bloating the diff.


willGraham01:
This OK to be merged in? (I don't have merge access to the TLO repository, but if there's a team I can join let me know!)

willGraham01:
@tamuri This OK to be merged in? (I don't have merge authority to `master`, but if there's a team I can join let me know)

PR #1044, opened 2023-07-28
Nic503/impact tb dah

nchagoma503:
Here is the new branch: reference files include: tb_DAH_scenarios.py and analysis_tb_DAH_scenarios.py. Expectation is that overriding availability of CXR and Xpert should increase DALYs and mortality while introduction of outreach and CXR scale up should do the opposite


tbhallett:
Shall we close this one? Or is it still being worked on?

tdm32:
> Shall we close this one? Or is it still being worked on?

Can be closed

PR #1053, opened 2023-08-02
Add autoscale formula to terminate nodes when no pending tasks

### LINKED ISSUE #678: Azure Batch: remove VMs from pool as tasks complete
Currently, under some circumstances, a pool retains idle VMs despite no tasks left to run. An autoscale rule should help i.e. [removing VMs if there aren't tasks to run](https://docs.microsoft.com/en-us/azure/batch/batch-automatic-scaling#pending-tasks).

tamuri:
Resolves #678 

Formula taken verbatim from https://learn.microsoft.com/en-us/azure/batch/batch-automatic-scaling#pending-tasks

There might be a better one, will explore

### DIFF
--- a/src/scripts/dev/scenarios/playing_22.py
+++ b/src/scripts/dev/scenarios/playing_22.py
@@ -1,3 +1,5 @@
+import time
+
 from tlo import Date, logging
 from tlo.methods import (
     demography,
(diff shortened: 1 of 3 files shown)


tamuri:
I think this formula will do for now - it works.

PR #1067, opened 2023-08-15
nic503/impact tb DAH

nchagoma503:
@Tara-mades changes to the class HSI_Tb_Xray_level1b and 2 to incorporate "probability_access_to_xray" as its previously defined as parameter but not used in the code. I just commented out the previous block of code but still hanging in there


giordano:
There seems to be something wrong with this PR: it's merging branch `master` onto `nic503/impact_Tb_DAH`, maybe it was supposed to be the opposite?

PR #1081, opened 2023-08-23
update parameters to enable interventions: a simpler and more reader-friendly way

EvaJanouskova:
`update_params_for_interventions()` fnc is re-written so that all the calculations are not done again and among those calculations multiplied by the interventions multipliers, instead the final probs without interventions are multiplied by those multipliers. 

The initiation probs 'p_stop_per_month' with intervention calculated with the old version versus new version of `contraception_initiation_with_interv` are the same if up to 16 decimals are considered.
(I have also removed the calculations for the years before the interventions started, as they are not needed at that time.)

The initiation probs after birth 'p_start_after_birth_below30' and 'p_start_after_birth_30plus' calculated with the old version versus new version of `contraception_initiation_after_birth_with_interv` are the same if up to 5 decimals are considered.

### DIFF
--- a/src/tlo/methods/contraception.py
+++ b/src/tlo/methods/contraception.py
@@ -596,122 +596,37 @@ class Contraception(Module):
 
         processed_params = self.processed_params
 
-        def expand_to_age_years(values_by_age_groups, ages_by_year):
-            _d = dict(zip(['15-19', '20-24', '25-29', '30-34', '35-39', '40-44', '45-49'], values_by_age_groups))
-            return np.array(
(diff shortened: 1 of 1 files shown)


tbhallett:
@EvaJanouskova  -- we have a conflict on this one, I see. Please could you resolve?

EvaJanouskova:
@tbhallett , I rebase the branch on master, it should help.

tbhallett:
@EvaJanouskova  -- the changes per se here look great. But there is a conflict to resolve before we can we merge. Please could you sort out the conflict? Thanks

tbhallett (approved):

PR #1087, opened 2023-08-31
Update calibrations stats for COPD

tbhallett:


mnjowe:
@tbhallett I'm re-running this file now. I will confirm the results after it has finished. Thanks

andrew-phillips-1:
@tbhallett Would you like me to review the updated graphs to check I am happy with them ?  Happy to do so if you show me the link to them @mnjowe

mnjowe:
@andrew-phillips-1, Yes I would be happy if you take a look at them. I will share you the link once I'm done running the simulation. Thanks

mnjowe:
@tbhallett and @andrew-phillips-1 . I've made some minor changes to both the calibration and analyses file - see the commit above.  I'm also attaching the COPD writeup with updated plots. The plots haven't changed that much. 

@tbhallett hope when we agree on the updated plots you will also update the write-up on tlo website. Thanks
[Copd.docx](https://github.com/UCL/TLOmodel/files/14193993/Copd.docx)

mnjowe:
Just seeing that for relative rate of death according to COPD  we are using 2023 data(Mannino et al 2003 ) . Correcting calibration plot on relative deaths now

mnjowe:
@andrew-phillips-1 and @tbhallett . Kindly find attached, the updated write-up. Thanks
[Copd.docx](https://github.com/UCL/TLOmodel/files/14195241/Copd.docx)

andrew-phillips-1:
Thanks @mnjowe (and @tbhallett) These generally look good.  For the last two graphs, have the results changed much from the ones we were looking at when this was being developed ?  If so, I can't recall our discussions on how we reconciled ourselves to the large differences between data and model.  The difference goes one way in the second last graph and the other way in the last graph.  (Incidentally, if the last graph is "relative rate" then the units should not include the "per 100 person years".  It would be helpful if the footnote for each graph mentions the reference for the data source (if you still have this).

mnjowe:
Hi @andrew-phillips-1 . Thanks for your feedback on this. These plots are almost the same with the last plots  when COPD was being developed. For those two last calibration plots, we agreed to just to include them in the write-up BUT trust the modal GBD plots on deaths if we're looking at deaths calibration (@tbhallett correct me if not).

andrew-phillips-1:
Ah yes @mnjowe thanks.  I now recall that discussion.  That seems fine. For the last graph since it is a relative rate perhaps we could plot on a log scale ?

mnjowe (approved):
I agree to the outputs. I think we can now merge this

Issue #1089, opened 2023-09-04
[BUG] `update_call` assigns work to officers with no time available


willGraham01:
The current version of the model throws a `KeyError` when attempting to run the `scale_run` profiling script.

Discovered after returning to issue #686's profiling PR (#1012) and the error being raised after a rebase onto `aace4d1`.

## To Reproduce:

In a terminal at the root of the TLOmodel repository:

```bash
$ mamba create -y -n tlo-test python=3.8
$ mamba activate tlo-test
$ (tlo-test) git status
On branch 'master'
Your branch is up-to-date with 'origin/master'.
$ (tlo-test) pip install -r requirements/dev.txt
...
$ (tlo-test) pip install -e .
$ (tlo-test) python src/scripts/profiling/scale_run.py <PARAMETERS HERE>
  File "src/scripts/profiling/scale_run.py", line 201, in <module>
    sim.simulate(end_date=end_date)
  File "/home/ccaegra/Documents/TLO/TLOmodel/src/tlo/simulation.py", line 234, in simulate
    self.fire_single_event(event, date)
  File "/home/ccaegra/Documents/TLO/TLOmodel/src/tlo/simulation.py", line 279, in fire_single_event
    event.run()
  File "/home/ccaegra/Documents/TLO/TLOmodel/src/tlo/events.py", line 66, in run
    self.apply(self.target)
  File "/home/ccaegra/Documents/TLO/TLOmodel/src/tlo/methods/healthsystem.py", line 2607, in apply
    self.process_events_mode_2(hold_over)
  File "/home/ccaegra/Documents/TLO/TLOmodel/src/tlo/methods/healthsystem.py", line 2459, in process_events_mode_2
    set_capabilities_still_available.remove(officer)
KeyError: 'FacilityID_91_Officer_Pharmacy'
```

The parameters passed to the `scale_run.py` script to reproduce this bug are:
```bash
python src/scripts/profiling/scale_run.py \
    --years 0 \
    --months 1 \
    --initial-population 50000 \
    --tlo-dir . \
    --output-dir ./outputs \
    --log-filename scale_run_profiling \
    --log-level DEBUG \
    --show-progress-bar \
    --seed 0 \
    --mode-appt-constraints 2
```

and the KeyError is thrown approximately 35% of the way through the simulation.

## Other Information

This bug is not present on commit `267c1e85f`, so was introduced at some point between then and `aace4d1`.

[...]

Issue #1090, opened 2023-09-04
actual_appt_footprint of HSI_Tb_Xray_level2 in conflict with mode_appt_constraints = 2


marghe-molaro:
In HSI_Tb_Xray_level2, if consumables are not available the HSI defaults to a clinical diagnosis, with adds another clinical appt footprint ("Under5OPD" : 1) to the expected one. Under5OPD relies on different medical officers than the expected_appt_footprint ("DiagRadio": 1);  an issue in mode_appt_constraints = 2 therefore occurs when capabilities for the officers required for the Under5OPD have already been exhausted before the HSI is ran, since we cannot know that those officers would be required until the HSI is actually ran. This is what leads to issue #1089.

There are two possible solutions:
1. Since this seems to be quite an infrequent occurrence (~3 times a month in a 50,000 population according to analysis in #1089, although this would depend on prioritisation policy) we could ignore the exhausted capabilities issue for the particular instance of the Under5OPD appt for this HSI.
2. For the specific case of HSI_Tb_Xray_level2, when checking if capabilities are still available we can, in addition to capabilities required for the expected_footprint, check whether there are also capabilities for an Under5OPD appt, whether those would actually be required when running the event or not.
3. Modify the HSI s.t. instead of simply adding the Under5OPD footprint, an additional clinical HSI is scheduled.

@tdm32, what would you recommend?

tdm32:
Hi @marghe-molaro, thanks for flagging this. 

I suppose if the consumables are not available for xray we should amend the footprint to be blank as the DiagRadio appt will not occur. I think what would likely happen is that they are referred back for a clinical diagnosis which would require another under5OPD. We might need to write a separate clinical HSI for this though as we don't need them to go through the whole triage process plus HIV testing in HSI_Tb_ScreeningAndRefer. It could even be a 0.5 Under5OPD appt with a pharmacy appt attached. 

I can do this as a separate PR if helpful?

marghe-molaro:
Hi @tdm32, I think a separate clinical HSI would fix this issue quite nicely, as it would require no "compromise" with the mode_appt_constraints assumptions (since we can check that all required officers are available to perform this clinical HSI before it is actually ran). A separate PR would be perfect, thank you!

tdm32:
@marghe-molaro I'm on it :-) 

Just a note to add the priority ranking for the new treatment id - if you're happy to do that. I don't want to mess with your system.

marghe-molaro:
[...]

PR #1091, opened 2023-09-05
Safely catch HSIs with different officers required for actual_appt_footprint in mode_appt_constraints = 2

marghe-molaro:
Fixes https://github.com/UCL/TLOmodel/issues/1089

The HSI likely to have caused issue #1089 has been identified in issue #1090, and this is currently being addressed. However more HSIs - whether already in the code or introduced in the future by new modules - may similarly clash with mode_appt_constraints = 2 and lead to a KeyError if their actual_appt_footprint requires different medical officers than their expected one. Because this clash will only manifest itself under specific and random circumstances (i.e. all officers required for the expected_footprint still have capabilities but those for the actual_footprint do not), some such HSIs may go undetected for a long time. It would therefore be preferable to ensure that these can be safely identified in future runs in mode_appt_constraints = 2 without leading to error, but flagging the issue as a warning, as done in this PR. As the incidence of this error at present appears to be very low, this is unlikely to significantly affect results relying on this warning.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -2456,7 +2456,14 @@ class HealthSystemScheduler(RegularEvent, PopulationScopeEventMixin):
                             # remove them from list of available officers.
                             for officer, call in updated_call.items():
                                 if capabilities_monitor[officer] <= 0:
-                                    set_capabilities_still_available.remove(officer)
+                                    if officer in set_capabilities_still_available:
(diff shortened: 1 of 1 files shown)


willGraham01 (approved):

tbhallett:
Have we confirmed that this fixes https://github.com/UCL/TLOmodel/issues/1089 and https://github.com/UCL/TLOmodel/issues/1090....?

tbhallett (approved):

PR #1094, opened 2023-09-05
Change Referral Pathway for TB when consumables not available: Try X-ray at level 1b and then 2, and then clinical diagnosis

### LINKED ISSUE #1090: actual_appt_footprint of HSI_Tb_Xray_level2 in conflict with mode_appt_constraints = 2

tdm32:
Fixes #1090

This fix addresses issue #1090 where the actual footprint returned by the TB xray HSIs can vary in terms of officer types required. This becomes an issue under mode_appt_constraints = 2 when officer time is constrained.

The TB diagnostic triage process has changed slightly to accommodate this:
- eligible children under 5 are first scheduled HSI_Tb_Xray_level1b
- if consumables are not available, a blank footprint is returned and person is referred to HSI_Tb_Xray_level2
- if no consumables available at level 2, a blank footprint is returned and person is referred to HSI_Tb_ClinicalDiagnosis 
- HSI_Tb_ClinicalDiagnosis is a new HSI which is a short (0.5* Under5OPD) appt. This appt does not repeat all the triage process of the full HSI_Tb_ScreeningAndRefer, which includes HIV test etc. but acts as a final diagnostic step if all other tests were not available. It refers a person to start treatment if the dx_test[tb_clinical] returns positive. 

test_tb.py tests all passing
treatment coverage rates consistent with version in master

Remaining issue is to add the new treatment ID to the policy lists (TREATMENT_ID = "Tb_Test_Clinical")

@marghe-molaro please review and let me know if this fixes the issue.


marghe-molaro:
Based on a local long_run_all_diseases.py run under mode_appt_constraints = 2, this PR appears to solve issue #1090 previously reported. 
The missing treatment_ID (Tb_Test_Clinical) has now been added to all policies and assigned the same priority assumed for Tb_Test_Xray.

marghe-molaro (approved):

tbhallett:
@tdm32  - there's a change in here for the Malaria write-up. I presume this is accidental so am rolling it back.

src/tlo/methods/tb.py
```diff
@@ -1933,6 +1933,74 @@ def apply(self, person_id, squeeze_factor):
...
+        # if none of the above conditions are present, no further action
+        persons_symptoms = self.sim.modules["SymptomManager"].has_what(person_id)
+        if not any(x in self.module.symptom_list for x in persons_symptoms):
+            return self.make_appt_footprint({})
+
+        # check for presence of all symptoms (clinical diagnosis)
+        if all(x in self.module.symptom_list for x in persons_symptoms):
+            test_result = self.sim.modules["HealthSystem"].dx_manager.run_dx_test(
+                dx_tests_to_run="tb_clinical", hsi_event=self
+            )
```
tbhallett, line 1984: do you want to do anything if a person has some but not all of the symptoms?

tdm32:
> @tdm32 - there's a change in here for the Malaria write-up. I presume this is accidental so am rolling it back.

Yes that's not intentional - thanks for changing

src/tlo/methods/tb.py
```diff
@@ -1933,6 +1933,74 @@ def apply(self, person_id, squeeze_factor):
...
+    def apply(self, person_id, squeeze_factor):
+        """ Do the screening and referring process """
+
+        df = self.sim.population.props
+        now = self.sim.date
+        person = df.loc[person_id]
+        test_result = None
+
+        # If the person is dead or already diagnosed, do nothing do not occupy any resources
+        if not person["is_alive"] or person["tb_diagnosed"]:
```
tbhallett, line 1967: The none / all logic could be accomplished more efficiently used sets, I think:

i.e. 
```python
if not any(x in self.module.symptom_list for x in persons_symptoms)
```
and 
```python
if all(x in self.module.symptom_list for x in persons_symptoms)
```

becomes:
```python
set_of_symptoms_that_indicate_tb = set(self.module.symptom_list)
if not set_of_symptoms_that_indicate_tb.intersect(persons_symptoms):
```
and 
```python
if set_of_symptoms_that_indicate_tb.issubset(persons_symptoms):
```

src/tlo/methods/tb.py
tdm32: No, I think the clinical criteria require all 4 symptoms to be present for a clinical diagnosis

tbhallett: OK!

tbhallett (approved):

PR #1096, opened 2023-09-05
Ensure mode_appt_constraints shift even under naive policy

marghe-molaro:
Before, we assumed a mode_appt_constraints = 2 under a "naive" policy from 2010, and switched the policy in 2023. If in 2023 we wanted to continue with a "naive" policy, there was therefore no need to schedule a policy change in 2023.
However we now assume a mode_appt_constraints = 1 until 2023; this means that the policy change (which also takes care of switching the mode) should be scheduled even in the case of the "naive" policy, to ensure that the mode_appt_constraints shift takes when evaluating a "naive" policy too. 
We have fixed this + changed the name of the policy-change function to more accurately reflect that it also takes care of changing mode of appt delivery.


src/tlo/methods/healthsystem.py
```diff
@@ -2782,21 +2781,29 @@ def apply(self, population):
...
+        # Change mode_appt_constraints
         self.module.mode_appt_constraints = self.module.parameters["mode_appt_constraints_postSwitch"]
-        if self.module.priority_policy != "":
-            self.module.load_priority_policy(self.module.priority_policy)
+
+        # If policy has changed, update it
+        if self.module.parameters["policy_name"] != self.module.parameters["policy_name_post_switch"]:
+            self.module.priority_policy = self.module.parameters["policy_name_post_switch"]
+            if self.module.priority_policy != "":
+                self.module.load_priority_policy(self.module.priority_policy)
```
tbhallett, line 2800: if the new priority policy is "" is it right to do nothing? What if the initial policy with something different? Don't we always need to actively set the policy to something.

Also,  I think the handling of the special case of "" should be inside `load_priority_policy` rather than repeated in the two places where the call is made.

src/tlo/methods/healthsystem.py
```diff
@@ -2782,21 +2781,29 @@ def apply(self, population):
...
+        print("Switched to ", self.module.mode_appt_constraints)
```
tbhallett, line 2801: Removing the bare print
```suggestion
```

marghe-molaro: Sorry!

src/tlo/methods/healthsystem.py
marghe-molaro: Mhm yes I'd say if the policy is "" it is right to do nothing, because in that case the policy will be ignored anyway, so it should be consistent. Will move the check to the function.

tbhallett: But what if we want:
Period 1: Prioritise RMNCH
Period 2: Natural policy (i.e. "").

Will that be achieved?

marghe-molaro: On line 1294
1294         if self.priority_policy != "":
1295             # Look-up priority ranking of this treatment_ID in the policy adopted
1296             priority = self.enforce_priority_policy(hsi_event=hsi_event)

This ensures that we will only look up a policy if a policy is not "". So if in Period 2 we set a policy "", this means for that period the priority assigned to the HSIs will not be looked up via the "enforce_priority_policy" fnc, but rather it will default to whatever was set by the disease modules themselves.

marghe-molaro: Hence it is fine for no policy to have been loaded in that period.

tbhallett (approved):

PR #1097, opened 2023-09-05
Run profiling on comment-trigger, and cron

willGraham01:
Updates the profiling workflow to:
- Run as a `cron` job once a week (Sundays at 00:00).
- Be triggered by comments on pull-requests.

Updates the `run-on-comment` workflow:
- Can be optionally passed a directory and tag to upload as an artifact after the run completes.

## Comment Triggering

Profiling can be triggered by commenting on a PR with
```
/run profiling
```
provided the commenting user has the necessary permissions to trigger a workflow run.

Our current comment-triggered workflow was only capable of running `shell` commands, and for the profiling runs we needed to trigger a workflow (since running the profiling alone is not enough, as we need to export and upload the artifacts). 

The solution suggested below is to adapt `run-on-comment` to allow export of an artifact. This means that, on comment-triggered runs we perform the `profile-on-comment` job, which passes the shell commands needed to trigger profiling to the `run-on-comment` workflow and requests an artifact to be produced with the profiling results. Conversely on workflows not triggered by comments, we can just run an alternative job (`profile-on-dispatch`) with the profiling commands and then upload an artifact with the same naming convention. The workflow will then check that at least one of `profile-on-comment` or `profile-on-dispatch` succeeded, and then push the results to the profiling repository and trigger a rebuild of the results website.

### DIFF
--- a/.github/workflows/run-on-comment.yml
+++ b/.github/workflows/run-on-comment.yml
@@ -195,6 +195,6 @@ jobs:
       if: "${{ inputs.artifact-path != '' }}"
       uses: actions/upload-artifact@v3
       with:
-        name: ${{ inputs.artifact-tag }}
+        name: ${{ inputs.artifact-name }}
         path: ${{ inputs.artifact-path }}
-        retention-days: ${{ inputs.artifact-days }}
+        retention-days: ${{ inputs.artifact-retention-days }}
--- a/.github/workflows/run-profiling.yaml
+++ b/.github/workflows/run-profiling.yaml
@@ -32,10 +32,10 @@ jobs:
     name: Create unique output file identifier and artifact name
     runs-on: ubuntu-latest
     outputs:
-      profiling-filename: ${{ steps.set-output-id.outputs.name }}
+      profiling-filename: ${{ steps.set-profiling-filename.outputs.name }}
       artifact-name: $${{ steps.set-artifact-name.outputs.name }}
     steps:
-      - id: set-output-id
+      - id: set-profiling-filename
         name: Set profiling output file name
         run: |
           echo "name=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA}" >> "${GITHUB_OUTPUT}"


.github/workflows/trigger-profiling.yaml
```diff
@@ -0,0 +1,154 @@
...
+            const run_test_commands_steps = job.steps.filter(
+              step => step.name == "Run test command(s)"
+            );
+            // There should only be one step with name "Run test command(s)"
+            if (run_test_commands_steps.length > 1) {
+              throw "Multiple steps with name 'Run test command(s)'";
+            }
+            step = run_test_commands_steps[0];
+            got_completed_step_info = (step.status == "completed");
+            attempts += 1;
```
matt-graham, line 137: I think the logic here will no longer work for the updated workflow as this specifically looks for a step with name `Run test command(s)` which will not exist for this workflow. Nearest equivalent is `Trigger profiling workflow` step but I'm not sure if the subsequent logic will follow through for this as its calling another reusable workflow rather than running a step directly (in particular the `step.status` check).

matt-graham (commented):
Making `run-profiling.yaml` define a reusable workflow seems like a good idea, but I wonder if we can somehow avoid the duplication of comment triggered workflow code by abstracting the resuable workflow in `run-on-comment.yaml` rather than partially copying and pasting. 

I think we could either go the route of adding the option of uploading artifacts from a comment triggered workflow or allow optionally calling a composite action instead of running a shell command from the comment triggered workflow.

For the former I think the easiest route would be to factor out the `upload-results` workflow job in to a reusable workflow and then have two workflow files calling into the reusable workflows - `comment-triggered-profiling.yaml` with contents something like

```YAML
on:
  issue_comment:
    types:
      - created
jobs:
  run-profiling:
    with:
      runs-on: self-hosted
      keyword: profiling
      commands: |
          UNIQUE_ID=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA}
          tox -vv -e profile -- --output_name ${UNIQUE_ID}
      description: Profiled run of model
      timeout-minutes: 8640
      application-organization: UCL
      artifact-tag: profiling
      artifact-path: profiling_results
      artifact-retention-days: 1
    secrets:
      application-id: ${{ secrets.COMMENT_BOT_APP_ID }}
      application-private-key: ${{ secrets.COMMENT_BOT_APP_PRIVATE_KEY }}
  upload-results:
    uses: ./github/workflows/upload-profiling-results.yml
    needs: run-profiling
    with:
      run-number: ${{ github.run-number }}
    secrets:
      profiling-repo-access-token: ${{ secrets.PROFILING_REPO_ACCESS }}
```

and `scheduled-and-workflow-dispatch-profiling.yaml` with contents something like

```YAML 
on:
  workflow_dispatch:
  schedule:
    - cron: 0 0 * * 6
jobs:
  run-profiling:
    runs-on: self-hosted
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3
      - name: Run profiling
        run: |
          UNIQUE_ID=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA}
          tox -vv -e profile -- --output_name ${UNIQUE_ID}
      - name: Save results as artifact
        uses: actions/upload-artifact@v3
        with:
          name: profiling_${{ github.run-number }}
          path: profiling_results/
          retention-days: 1
  upload-results:
    uses: ./github/workflows/upload-profiling-results.yml
    needs: run-profiling
    with:
      run-number: ${{ github.run-number }}
    secrets:
      profiling-repo-access-token: ${{ secrets.PROFILING_REPO_ACCESS }}
```

This would still duplicate the profiling run command and logic around naming of artifact, but would be less repetition overall. 

Not sure that this is better than your proposed solution given extra complexity but thought I'd mention as something to discuss!

willGraham01:
> but I wonder if we can somehow avoid the duplication of comment triggered workflow code by abstracting the resuable workflow in run-on-comment.yaml rather than partially copying and pasting.

Yeah I'm definitely not going to defend that the copy/paste was the best design choice on my part. :sweat_smile: 

I think I like option 1: we give `run-on-comment` some additional (optional) inputs; `artifact-name`, `artifact-path`, etc, and an additional (optional) extra step right at the end that uploads the artifact. We can abstract the `upload-profiling-results` to another workflow as you say too.

Then we only actually need one workflow file for the profiling itself, unless you'd prefer to explicitly put comment-triggered and non-comment-triggered into their own separate workflow files:

```yaml
name: Run Profiling

on:
  workflow_dispatch:
  # Allow profiling to be triggered by comments on pull requests
  # Trigger is /run profiling
  issue_comment:
    types:
      - created
  # Profile the model every Sunday at 00:00, 
  # on the HEAD of master
  schedule:
    - cron: 0 0 * * 6

jobs:
  profile-on-comment:
    if: github.event_name == 'issue_comment'
    # what you have above, passing to run-on-comment.yml, 
    # with artifact upload option enabled
  
  profile-on-dispatch:
    if: github.event_name != 'issue_comment'
    # explicit tox call, followed by upload step
  
  upload-profiling-results:
    needs: [profile-on-dispatch or profile-on-comment] # TODO: check syntax!
    uses: ./.github/workflows/upload-results.yaml
    # suitable upload workflow that runs on ubuntu-latest, 
    # and fetches the artifact that was created by whichever job above was actually performed
```

In fact, I don't think we'd even to factor out `upload-results` into another workflow here, as we'd only be calling it from within this file (unless again, you think having separate workflows for the two profiling triggers would be better)?

matt-graham:
> Then we only actually need one workflow file for the profiling itself, unless you'd prefer to explicitly put comment-triggered and non-comment-triggered into their own separate workflow files:
>  ...
> In fact, I don't think we'd even to factor out upload-results into another workflow here, as we'd only be calling it from within this file (unless again, you think having separate workflows for the two profiling triggers would be better)?

Agree that what you propose here would be better but I think it's not possible for a step to depend on one of a set of previous steps rather than all of set of previous steps that is I don't think the

```YAML
needs: [profile-on-dispatch or profile-on-comment] # TODO: check syntax!
```
syntax exists - [it looks like we can only specify an intersection not a union](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

We could potentially have two jobs which each depend on just one of the previous steps, but then we get back to duplicating code unless we factor out to a reusable workflow. It might be possible to do something similar to `needs` with an `if` conditional, but I'm not sure.

willGraham01:
> Agree that what you propose here would be better but I think it's not possible for a step to depend on one of a set of previous steps rather than all of set of previous steps 

We might be able to get around this with a slight hack (I think `brainglobe` uses this hack for something similar but would need to double check) - we have an intermediate job/step that checks that at least one of `profile-on-comment` or `profile-on-dispatch` ran successfully. Something like

```yaml
upload-profiling-results:
  if: |
    jobs.profile-on-comment.result == 'success' || jobs.profile-on-dispatch.result == 'success'
  uses: ./.github/workflows/upload-results.yaml
  # suitable upload workflow that runs on ubuntu-latest, 
  # and fetches the artifact that was created by whichever job above was actually performed
```

.github/workflows/trigger-profiling.yaml
willGraham01: I think this is fixed by our choice to recycle the `run-on-comment` workflow now

.github/workflows/run-profiling.yaml
```diff
@@ -1,77 +1,78 @@
...
+  profile-on-comment:
+    name: Comment triggered profiling
+    if: github.event_name == 'issue_comment'
+    uses: ./.github/workflows/run-on-comment.yml
+    with:
+      runs-on: self-hosted
+      keyword: profiling
+      commands: |
+        UNIQUE_ID=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA}
+        tox -vv -e profile -- --output_name ${UNIQUE_ID}
```
matt-graham, line 40: To guard against `UNIQUE_ID` becoming out of sync with the corresponding value passed to `--output_name` below then we could potentially have a preceding job which both `profile-on-comment` and `profile-on-dispatch` depend on which sets the identifier and artifact name as outputs and then uses these outputs in the dependent jobs, something like:

```YAML
  set-variables:
    name: Create unique identifier and artifact name for profiling outputs
    runs-on: ubuntu-latest
    outputs:
        unique-id: ${{ steps.set-unique-id.outputs.id }}
        artifact-name: ${{ steps.set-artifact-name.outputs.name }}
    steps:
      - id: set-unique-id
        name: Set identifier output
        run: echo "id=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
      - id: set-artifact-name
        name: Set artifact name output
        run: echo "name=profiling_results_${GITHUB_RUN_NUMBER}" >> "$GITHUB_OUTPUT"

  profile-on-comment:
    needs: set-variables
    ...
    with:
      ...
      commands: tox -vv -e profile -- --output_name ${{ needs.set-variables.unique-id }}
      ...
      artifact-name: ${{ needs.set-variables.artifact-name }}
      ...
 
  profile-on-dispatch:
    needs: set-variables
    ...
    steps:
      ...
      - name: Run profiling in dev environment
        run: tox -vv -e profile -- --output_name ${{ needs.set-variables.unique-id }}
      ...
      - name: Save results as artifact
        uses: actions/upload-artifact@v3
        with:
          name: ${{ needs.set-variables.artifact-name }}
          ...

  upload-profiling-results:
    ...
    needs: [set-variables, profile-on-comment, profile-on-dispatch]
    ...
    steps:
      - name: Download the profiling results
        uses: actions/download-artifact@v3
        with:
          name: ${{ needs.set-variables.artifact-name }}
          path: .
      ...
```

Having written that all out it possibly seems a bit too complex for a minimal avoidance of repetition 😅 so probably not worthwhile!

matt-graham (changes requested):
Thanks @willGraham01 this is looking great. Have suggested a few possible minor naming changes and a small typo fix in one key but otherwise looks almost ready to merge.

.github/workflows/run-profiling.yaml
willGraham01: I think we should have the `set-variables` step regardless - I'm not sure how GitHub handles evaluation of variables like ${GITHUB_RUN_NUMBER} (I presume in the sensible way) but this at least means we don't have to worry about it.

It's only adding ~7 extra lines and it does mean there's one fewer place to make a typo in the file names we want to produce!

.github/workflows/run-profiling.yaml
```diff
@@ -1,77 +1,78 @@
...
+    # because we need Docker for the workflow call below,
+    # and because we're now just pushing results files to another repo
     runs-on: ubuntu-latest
-    needs: run-profiling
+    # only run this step if exactly one of the previous two jobs ran successfully
+    if: |
+      always() &&
+      (jobs.profile-on-comment.result == 'success' || jobs.profile-on-comment.result == 'skipped') &&
+      (jobs.profile-on-dispatch.result == 'skipped' || jobs.profile-on-dispatch.result == 'success') &&
+      !(jobs.profile-on-comment.result == 'skipped' && jobs.profile-on-dispatch.result == 'skipped')
```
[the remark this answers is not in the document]
willGraham01: Agreed it should be equivalent (although I've just noticed `jobs` needs to swap to `needs`) so I'll change it and see whether GitHub complains.

.github/workflows/run-on-comment.yml
```diff
@@ -165,3 +190,11 @@ jobs:
...
             body: `## ${{inputs.description }} ${result}\n${details}`,
           });
+
+    - name: Upload artifact if requested
+      if: "${{ inputs.artifact-path != '' }}"
+      uses: actions/upload-artifact@v3
+      with:
+        name: ${{ inputs.artifact-tag }}
+        path: ${{ inputs.artifact-path }}
+        retention-days: ${{ inputs.artifact-days }}
```
matt-graham, line 200:
```suggestion
        name: ${{ inputs.artifact-name}}
        path: ${{ inputs.artifact-path }}
        retention-days: ${{ inputs.artifact-retention-days }}
```

.github/workflows/run-profiling.yaml
```diff
@@ -9,39 +9,106 @@ name: Run profiling
...
-  run-profiling:
-    name: Setup developer environment
+  set-variables:
+    name: Create unique output file identifier and artifact name
+    runs-on: ubuntu-latest
+    outputs:
+      profiling-filename: ${{ steps.set-output-id.outputs.name }}
+      artifact-name: $${{ steps.set-artifact-name.outputs.name }}
+    steps:
+      - id: set-output-id
```
matt-graham, line 38:
```suggestion
      profiling-filename: ${{ steps.set-profiling-filename.outputs.name }}
      artifact-name: $${{ steps.set-artifact-name.outputs.name }}
    steps:
      - id: set-profiling-filename
```
Unnecessary but just to keep things consistent

matt-graham (changes requested):
Looks great - thanks for all your work on this @willGraham01 and for putting up with my overly long comments 😁. One small change needed I think to ensure all internal references to input names are consistent but otherwise looks good to merge.

.github/workflows/run-on-comment.yml
willGraham01: VSCode's `CRTL+F` has betrayed me :cry:

matt-graham:
@tamuri did you also want to review this? From my perspective this is good to merge.

Issue #988, opened 2023-05-31
Scenario Definitions Cannot Over-write Parameters Defined as a `pd.Series` or a `pd.DataFrame`


tbhallett:
The below exemplar scenario definition -- wherein we attempt to over-write in `draw_parameters` a parameter that is defined either as a `pd.Series` or a `pd.DataFrame` -- fails with this error:

```TypeError: Object of type DataFrame is not JSON serializable```

But, many parameters that need to be overwritten are in the format!

```python

from pathlib import Path
import pandas as pd
from tlo import Date, logging
from tlo.methods.fullmodel import fullmodel
from tlo.scenario import BaseScenario


class TestScenario(BaseScenario):
    def __init__(self):
        super().__init__()
        self.seed = 0
        self.start_date = Date(2010, 1, 1)
        self.end_date = Date(2020, 1, 1)
        self.pop_size = 100_000
        self.number_of_draws = 1
        self.runs_per_draw = 5

    def log_configuration(self):
        return {
            'filename': 'effect_of_each_treatment',
            'directory': Path('./outputs'),
            'custom_levels': {
                '*': logging.WARNING,
                'tlo.methods.demography': logging.INFO,
                'tlo.methods.demography.detail': logging.WARNING,
                'tlo.methods.healthburden': logging.INFO,
                'tlo.methods.healthsystem.summary': logging.INFO,
            }
        }

    def modules(self):
        return fullmodel(resourcefilepath=self.resources)

    def draw_parameters(self, draw_number, rng):
        return {
            'Module': {
                'param_defined_as_series': pd.Series(['a', 'b', 'c']),
                'param_defined_as_dataframe': pd.DataFrame([[0, 1, 2], [3, 4, 5]]),
            },
        }


if __name__ == '__main__':
    from tlo.cli import scenario_run

    scenario_run([__file__])

```

tamuri:
See #392

PR #1098, opened 2023-09-06
The `Equipment` Class

EvaJanouskova:
main changes:
* modified `healthsystem` module to allow tracking and logging of equipment used

minor changes:  
* added some TODOs (hoping at least some of them will be resolved before this PR closes)

TODO
-----
- [x] decide on empty declaration appearance
- [x] logging of equipment counts of use by HSI event & facility level
- [x] script to catalogue equipment associated with each HSI event & catalogue equipment by treatment id and facility levels
- [x] add item codes to RF
- [x] log item codes instead of names
- [x] add possibility of adding equipment packages (not just items)
- [x] make possible equipment pkgs to be set as essential equipment
- [x] <del>test essential equipment set for all HSI events</del>
        if essential equipment setting is missing, set it to empty set and add HSI event name to warning
- [x] <del>test essential equipment is in RF_Equipment</del>
        <del>test equipment potentially used is in RF_Equipment (if possible)</del>
        if any equipment item or pkg name provided in the code is not included in RF_Equipment, ignore the item(s) and 
add the item/pkg name to warning
- [x] stop HSI events from running if essential equipment not available
- [x] add possibility to check equip availability within an hsi event
- [x] add switcher: no X all X default equipment available
- [x] equip availability columns by facility levels in RF_Equipment with dummy values (True for all)
- [x] availability of items based on RF_Equipment (now, all available except a few item codes)
- [x] tests
- [x] add instructions for equipment integration to the wiki
- [x] make the script code_to_items_list to be a function in the utils module

before merge:
- [x] rm the dummy examples before merge
- [x] what equip availability default values for the merge?

move to next PR:
- [x] open issue to include essential equipment if any in `TLOmodel/src/tlo/analysis/hsi_events.py` script to be shown on https://www.tlomodel.org/hsi_events.html


src/scripts/healthsystem/equipment/equipment_catalogue.py
```diff
@@ -0,0 +1,102 @@
...
+from pathlib import Path
+
+import pandas as pd
+
+from tlo.analysis.utils import extract_results
+
+
+def get_annual_equipment_declarations_by_levels(results_folder: Path) -> pd.DataFrame:
+    """Return pd.DataFrame gives the simulated annual equipment declaration by facility levels for each simulated
+    year."""
```
tbhallett, line 11: Perhaps also say that we require the `healthsystem.summary` logger to have been set at the level of `logger.INFO` or higher.

src/tlo/analysis/utils.py
```diff
@@ -282,6 +282,9 @@ def get_multiplier(_draw, _run):
     # get number of draws and numbers of runs
     info = get_scenario_info(results_folder)
 
+    def is_number(element):
```
tbhallett, line 285: good idea.

src/tlo/methods/breast_cancer.py
```diff
@@ -575,6 +575,7 @@ class BreastCancerMainPollingEvent(RegularEvent, PopulationScopeEventMixin):
     def __init__(self, module):
         super().__init__(module, frequency=DateOffset(months=1))
         # scheduled to run every 3 months: do not change as this is hard-wired into the values of all the parameters.
+        # TODO: Is it? There is 1 month as the frequency, isn't it?
```
tbhallett, line 578: Good spot. Please update the comment!

src/tlo/methods/breast_cancer.py
```diff
@@ -646,6 +647,7 @@ def __init__(self, module, person_id):
         self.TREATMENT_ID = "BreastCancer_Investigation"
         self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({"Over5OPD": 1, "Mammography": 1})
         self.ACCEPTED_FACILITY_LEVEL = '3'  # Mammography only available at level 3 and above.
+        # TODO: what this means, should be the mammography done within this event, or the biopsy, or both?
```
tbhallett, line 650: Not sure about this. Better ask Andrew.

src/tlo/methods/breast_cancer.py
```diff
@@ -666,6 +668,8 @@ def apply(self, person_id, squeeze_factor):
 
         # Use a biopsy to diagnose whether the person has breast Cancer:
         # todo: request consumables needed for this
```
tbhallett, line 670: Adding consumbales can be a future issue.

src/tlo/methods/breast_cancer.py
```diff
@@ -666,6 +668,8 @@ def apply(self, person_id, squeeze_factor):
...
+        self.used_equipment = {'Slice Master sample processing Unit', 'Paraffin Dispense', 'Whatever used with biopsy',
```
tbhallett, line 671: As the HSI event will already have `self.used_equipment` as an empty set. I think we should encourage users to use `self.used_equipment.update(['Slice Master sample processing Unit', 'Paraffin Dispense'])`

tbhallett: I think a good style would be:
1) If the items to be used in the HSI are always the same (most cases), do the delcaration in `__init__` alongside other declarations: 
`self.equipment = {'item1', 'item2', etc.}` 
2) If the items may change during the HSI, then the declaration can be made/updated during the `apply` function:
`self.equipment = {'item1', 'item2', etc.}  # over-writes what was in the __init__` 
OR: `self.equipment.update(['item1', 'item2', etc.])  # add to what was in the __init__` 

In both cases, we tell the user that this isn't "read" until the HSI has run.

src/tlo/methods/healthsystem.py
```diff
@@ -182,6 +182,7 @@ def __init__(self, module, *args, **kwargs):
         self._received_info_about_bed_days = None
         self.expected_time_requests = {}
         self.facility_info = None
+        self.used_equipment = set()
```
tbhallett, line 185: to be consistent with other declarations that the HSI makes about what it requires, I think we should use capitals. I also think we don't need "used" as it's implicit.
```sugggetion:
self.EQUIPMENT = set()
```

src/tlo/methods/rti.py
```diff
@@ -4665,6 +4665,8 @@ def apply(self, person_id, squeeze_factor):
             # administer antibiotic
             get_item_code("Ampicillin injection 500mg, PFR_each_CMST"): 1,
             # equipment used by surgeon, gloves and facemask
+            # TODO: As we now consider both consumables and equipment, using 'equipment' when meaning consumables is
```
tbhallett, line 4668: yes, this could have said "consumables used by surgeon...""

src/tlo/methods/newborn_outcomes.py
```diff
@@ -410,6 +410,8 @@ def get_and_store_newborn_item_codes(self):
         get_list_of_items = pregnancy_helper_functions.get_list_of_items
 
         # ---------------------------------- IV DRUG ADMIN EQUIPMENT  -------------------------------------------------
+        # TODO: As we now consider both consumables and equipment, using 'equipment' when meaning consumables is
```
tbhallett, line 413: Yes, could correct the comment to say 'consumable'

src/tlo/methods/contraception.py
```diff
@@ -744,6 +744,7 @@ def get_item_code_for_each_contraceptive(self):
         get_items_from_pkg = self.sim.modules['HealthSystem'].get_item_codes_from_package_name
 
         _cons_codes = dict()
+        # items for each method that requires an HSI to switch to
```
tbhallett, line 747: These comments are useful (and other typos picked-up appreciated), but in general we should try to make the changes in a PR specific to it's aim and as small as possible for that aim. (Comments and other things you want to add incidentally, can be put into a seperate "omnibus comment changes" PR.)

src/tlo/methods/contraception.py
```diff
@@ -1285,6 +1288,13 @@ def apply(self, person_id, squeeze_factor):
                              )
 
             _new_contraceptive = self.new_contraceptive
+
+            # Record used equipment when needed
+            if _new_contraceptive == 'female_sterilization':
+                self.used_equipment = {'Smt used to sterilize a woman'}
```
tbhallett, line 1294:
```suggestion
                self.used_equipment.update({'Smt used to sterilize a woman'})
```

src/tlo/methods/contraception.py
```diff
@@ -1285,6 +1288,13 @@ def apply(self, person_id, squeeze_factor):
...
+            elif _new_contraceptive == 'IUD':
+                self.used_equipment = {'Equipment used when performing IUD'}
```
tbhallett, line 1296:
```suggestion
                self.used_equipment.update({'Equipment used when performing IUD'})
 ```

File: src/tlo/methods/healthsystem.py [hunk 13]
```diff
@@ -2738,6 +2748,14 @@ def write_to_log_and_reset_counters(self):
             },
         )
 
+        logger_summary.info(
+            key="Equipment",
+            description="Sets of used equipment for each facility level in this calendar year.",
+            data={
+                "Equipment_By_Level": self._equip_by_level,
```
tbhallett, line 2755: As sets aren't ordered, but we want to the results of two runs of the model to be identical (if same random seed and same parameters), I propose that we use `sorted` that will turn the `set` into `list` that is sorted. 


```suggestion
                "Equipment_By_Level": sorted(self._equip_by_level),
```

tbhallett (changes requested):
Great start on this Eva. Only minor comments. 

Do we have an idea about how to provide people with a catalogue of equipment they can choose from when making these declarations?

src/tlo/methods/contraception.py
EvaJanouskova: I have reverted those commits and created separate PR as you suggested: #1105.

src/tlo/methods/breast_cancer.py
tbhallett: Looks perfect.

src/tlo/methods/contraception.py
tbhallett: Thanks for that!

tbhallett:
> Hi @EvaJanouskova (and @tbhallett) It's been a while since I worked on the module so I have forgotten all the reading I did at the time. I'm fairly sure that for some of your questions I didn't know the answer to even at that time. When I return I can try to spend some time over the next few weeks trying to read up about it again if that seems something worth me prioritising at this time.

Thanks @andrew-phillips-1. I don't think a wholesale update or anything like that is needed at the moment (although we have something along these lines slated for version 2, when we have someone available to support). For now, we just need to know the equipment that (implicitly) is being used. In the coming weeks, Eva will gather a catalogue of items and ask all disease module leads to identify the ones used in each HSI. This is needed for the costing model and to allow us to look at some analyses that MOH/CHAI are asking about.

andrew-phillips-1:
Thanks @tbhallett - that seems good and I'm happy to try to do that for the modules I developed.  @EvaJanouskova Do you know how other module leads are going about checking on equipment use for their modules ?  There are the 2015 clinical guidelines, and I guess the service assessment survey is of some use, but do you know of other sources, that I have perhaps paid insufficient attention to ?

src/tlo/methods/breast_cancer.py
andrew-phillips-1: Hi @EvaJanouskova Thanks - I've looked back at this.  It looks like you are right that we moved over to a 1 month update.  So the comment can be updated or removed.   I also see that the write-up needs updating so I will try to do this when I can.

src/tlo/methods/breast_cancer.py
andrew-phillips-1: @EvaJanouskova I don't know for sure but I think we should consider this as just biopsy.  As you saw in the guidelines, population screening for mammography is not done.  So I think the mammography can be removed here.

src/tlo/methods/breast_cancer.py
```diff
@@ -646,6 +647,9 @@ def __init__(self, module, person_id):
         self.TREATMENT_ID = "BreastCancer_Investigation"
         self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({"Over5OPD": 1, "Mammography": 1})
         self.ACCEPTED_FACILITY_LEVEL = '3'  # Mammography only available at level 3 and above.
+        # TODO: what this means, should be the mammography done within this event, or the biopsy, or both?
+        self.EQUIPMENT = {'Slice Master sample processing Unit', 'Paraffin Dispense', 'Whatever used with biopsy',
+                          'Mammography maybe?'}  # biopsy and ?mammography always performed with this HSI
```
andrew-phillips-1, line 652: let's take mammography out

src/tlo/methods/breast_cancer.py
```diff
@@ -759,6 +763,10 @@ def apply(self, person_id, squeeze_factor):
         df.at[person_id, "brc_date_treatment"] = self.sim.date
         df.at[person_id, "brc_stage_at_which_treatment_given"] = df.at[person_id, "brc_status"]
 
+        # Update equipment used with treatment
+        # NB. read only with HSI run and healthsystem.summary logger set at the level INFO or higher
+        self.EQUIPMENT.update({'Anything used for mastectomy as I guess this is about'})
```
andrew-phillips-1, line 768: yes surgery for mastectomy.  Radiotherapy is not at all widely available yet I believe.  Chemotherapy is used but not equipment needed as far as I am aware.

src/tlo/methods/breast_cancer.py
andrew-phillips-1: @EvaJanouskova @tbhallett 
Just trying to look back at this BreastCancer module and understand the code.  On line 669 I think we should have this        

df.at[person_id, 'brc_breast_lump_discernible_investigated'] = True

instead of 

df.brc_breast_lump_discernible_investigated = True

EvaJanouskova:
> Great start on this Eva. Only minor comments.
> 
> Do we have an idea about how to provide people with a catalogue of equipment they can choose from when making these declarations?

It depends on the purpose and scope of the catalogue. Are we looking for a simple list for users to reference, where they manually copy equipment names as needed? Should it include only equipment currently modelled, meaning that when a new piece of equipment is added to the model, it must also be added to the catalogue?

What functionalities or requirements on the catalogue did you have in mind?

src/tlo/methods/breast_cancer.py
EvaJanouskova: I've update the comment from `# Mammography only available at level 3 and above.` to `# Biopsy only available at level 3 and above.`, is that correct?

EvaJanouskova:
> Thanks @tbhallett - that seems good and I'm happy to try to do that for the modules I developed. @EvaJanouskova Do you know how other module leads are going about checking on equipment use for their modules ? There are the 2015 clinical guidelines, and I guess the service assessment survey is of some use, but do you know of other sources, that I have perhaps paid insufficient attention to ?

Hi @andrew-phillips-1, I've just received a draft of the new 2023 version of the Malawi standard treatment guidelines. I will make these guidelines and the survey easily available to everyone. Additionally, we will create an equipment catalogue, allowing everyone to find the equipment needed for their module(s) and ensuring consistency by using the same names if used by multiple leaders.

andrew-phillips-1:
Sounds good @EvaJanouskova

src/tlo/methods/breast_cancer.py
EvaJanouskova: I think you are right. @tbhallett, could you confirm this? Should this be changed in small separate PR?

EvaJanouskova:
_Note: rebased on master_

EvaJanouskova:
_Note: rebased on master_

EvaJanouskova:
_Note: rebased on master_

EvaJanouskova:
_rebased on master_

EvaJanouskova:
Hi @matt-graham, I've done the changes  as we agreed. Could you please have a look, and let me know if I can call this done and start working on the corresponding changes for all the modules in the PR [equipment: integration into modules](https://github.com/UCL/TLOmodel/pull/1124#top)
#1124?

EvaJanouskova:
Now the equipment item codes are logged instead of equipment item names.

src/tlo/methods/breast_cancer.py
EvaJanouskova: Hi @tbhallett, Should we do this change here or in a separate PR?

tbhallett: We can do it here as it so small.

src/tlo/methods/healthsystem.py
```diff
@@ -337,6 +343,45 @@ def make_appt_footprint(self, dict_of_appts):
...
+                "equipment item names from ResourceFile_Equipment.csv."
+            )
+        # from the set of equip item names create a set of item codes
+        # this function is calling parameters from this
+        equip_codes = set(self.get_equip_item_code_from_item_name(
+            self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups'], item_name
+        ) for item_name in set_of_equip
+        )
+        self.EQUIPMENT.update(equip_codes)
+        self.EQUIPMENT.discard('')
```
matt-graham, line 383: What is the reason for removing any `''` empty string entry?

src/tlo/methods/healthsystem.py
```diff
@@ -337,6 +343,45 @@ def make_appt_footprint(self, dict_of_appts):
...
+                "Argument to add_equipment should be a non-empty set of strings of "
+                "equipment item names from ResourceFile_Equipment.csv."
+            )
+        # from the set of equip item names create a set of item codes
+        # this function is calling parameters from this
+        equip_codes = set(self.get_equip_item_code_from_item_name(
+            self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups'], item_name
+        ) for item_name in set_of_equip
+        )
+        self.EQUIPMENT.update(equip_codes)
```
matt-graham, line 382: Is there a reason why `self.EQUIPMENT` contains equipment item codes while `self.ESSENTIAL_EQUIPMENT` contains equipment item names?

src/tlo/methods/healthsystem.py
```diff
@@ -337,6 +343,45 @@ def make_appt_footprint(self, dict_of_appts):
...
+        # Set EQUIPMENT if the given set_of_equip in correct format, ie a set of strings or an empty set
+        if not isinstance(set_of_equip, set) or any(not isinstance(item, str) for item in set_of_equip):
+            raise ValueError(
+                "Argument to set_equipment_essential_to_run_event should be an empty set or a set of strings of "
+                "equipment item names from ResourceFile_Equipment.csv."
+            )
+
+        self.ESSENTIAL_EQUIPMENT  = set_of_equip
+
+    def add_equipment(self, set_of_equip):
```
matt-graham, line 364:
```suggestion
    def add_equipment(self, set_of_equip: Set[str]) -> None:
```

Not essential but additing a type hint like this makes it explicit what the expected type of the argument is (and that the method doesn't return anything). Would also need to add a `from typing import Set` statement.

src/tlo/methods/healthsystem.py
```diff
@@ -337,6 +343,45 @@ def make_appt_footprint(self, dict_of_appts):
             "values"
         )
 
+    def get_equip_item_code_from_item_name(self, lookup_df: pd.DataFrame, equip_item_name: str) -> int:
+        """Helper function to provide the equip_item_code (an int) when provided with the equip_item_name of the item"""
+        return int(pd.unique(lookup_df.loc[lookup_df["Equip_Item"] == equip_item_name, "Equip_Code"])[0])
+
+    def set_equipment_essential_to_run_event(self, set_of_equip):
```
matt-graham, line 350:
```suggestion
    def set_equipment_essential_to_run_event(self, set_of_equip: Set[str]) -> None:
```

Not essential but additing a type hint like this makes it explicit what the expected type of the argument is (and that the method doesn't return anything). Would also need to add a `from typing import Set` statement.

src/tlo/methods/healthsystem.py
```diff
@@ -182,6 +183,11 @@ def __init__(self, module, *args, **kwargs):
         self._received_info_about_bed_days = None
         self.expected_time_requests = {}
         self.facility_info = None
+        # self.set_equipment_essential_to_run_event({''})  # HSI needs this attribute, but it is not defined in the Base
+        #                                                    class to allow verification of its existence as a test for
+        #                                                    each HSI event, showing that equipment setup was thought
+        #                                                    through for the event.
```
matt-graham, line 189:
```suggestion
        # HSI_Event instances needs the essential equipment to run event to be set using
        # the set_equipment_essential_to_run_event method, but it is not called with
        # a default value in the base class to allow explicit checking if the method
        # was called as a test for each HSI_Event, showing that equipment setup was 
        # thought through for the event.
```

I would avoid having commented code. For a long comment documenting the expectations of how subclasses use the base class like this, I would say possibly makes more sense to add to docstring for the class (multiline string after `class ...:` line, starting `Base HSI event class, from which all others inherit.` above.

matt-graham: Just realised this is following pattern `# self.EXPECTED_APPT_FOOTPRINT ... ` comment above, so feel free to ignore.

src/tlo/methods/healthsystem.py
```diff
@@ -337,6 +343,45 @@ def make_appt_footprint(self, dict_of_appts):
...
+            raise ValueError(
+                "Argument to add_equipment should be a non-empty set of strings of "
+                "equipment item names from ResourceFile_Equipment.csv."
+            )
+        # from the set of equip item names create a set of item codes
+        # this function is calling parameters from this
+        equip_codes = set(self.get_equip_item_code_from_item_name(
+            self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups'], item_name
+        ) for item_name in set_of_equip
+        )
```
matt-graham, line 381:
```suggestion
        equip_codes = set(
            self.get_equip_item_code_from_item_name(
                self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups'], 
                item_name
            ) for item_name in set_of_equip
        )
```
Having the nested parentheses at the same indentation level makes it a bit difficult to see the block structure here.

Also is there a reason to have `self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups']` passed as an argument to `get_equip_item_code_from_item_name` method rather than just directly accessed within that method using `self` argument, as I think that would help make the code a bit more easily readable here?

EvaJanouskova: I think `self` directs to the HSI event here not to the HealthSystem module, as if I use only `self.parameters['equip_item_and_package_code_lookups']` instead of `self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups']` I get an `AttributeError: 'HSI_Contraception_FamilyPlanningAppt' object has no attribute 'parameters'`

src/tlo/methods/healthsystem.py
EvaJanouskova: No reason in particular, only that I need the `self.EQUIPMEN` to be codes for the logging to be substantially shorter, but I don't log the `self.ESSENTIAL_EQUIPMENT` hence don't need to be codes. Do you think it would be better if these are both codes anyway?

src/tlo/methods/healthsystem.py
EvaJanouskova: Good catch, it shouldn't be there anymore, it was only needed when the item names were saved in `self.EQUIPMENT`.

EvaJanouskova (commented):
Thank you @matt-graham, please see the comments:

src/tlo/methods/healthsystem.py
EvaJanouskova: I've rm-ed this.

src/tlo/methods/healthsystem.py
matt-graham: Yeah I'd say probably best to be consistent just to avoid people accidentally assuming these are coded the same.

src/tlo/methods/healthsystem.py
matt-graham: Sorry here I meant as the `self` argument in the `get_equip_item_code_from_item_name` method could also be used to access `self.sim.modules['HealthSystem'].parameters['equip_item_and_package_code_lookups']` I don't think you need to have it passed in to the method at all so you could call it as `self.get_equip_item_code_from_item_name(item_name)`

src/tlo/methods/healthsystem.py
EvaJanouskova: Done.

src/tlo/methods/healthsystem.py
EvaJanouskova: Ah, got it now. Yes, that makes sense. Done.

EvaJanouskova:
The option of using also equipment packages is in.

EvaJanouskova:
Hi @matt-graham, I did run tox check on my laptop and it used to take ~65 sec, but now it was ~85 sec, should I be worried, does it mean smt?
`(86.70=setup[12.62]+cmd[0.01,1.20,72.67,0.06,0.13] seconds)`

src/tlo/analysis/codes_to_items_list.py
```diff
@@ -0,0 +1,75 @@
+"""
```
tamuri, line 1: Is this a file processing script? Should probably we reworked to be a function in the utils module.

src/tlo/analysis/utils.py
```diff
@@ -293,7 +296,10 @@ def generate_series(dataframe: pd.DataFrame) -> pd.Series:
                 df: pd.DataFrame = load_pickled_dataframes(results_folder, draw, run, module)[module][key]
                 output_from_eval: pd.Series = generate_series(df)
                 assert pd.Series == type(output_from_eval), 'Custom command does not generate a pd.Series'
-                res[draw_run] = output_from_eval * get_multiplier(draw, run)
+                if output_from_eval.apply(is_number).all():
```
tamuri, line 299: Is this to check that the Series is float or int dtype? Can you use \[is\_numeric\_dtype\](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_numeric_dtype.html)?

src/tlo/analysis/codes_to_items_list.py
EvaJanouskova: Okay, I've added this among TODO list above. As this want change anything regarding the equipment structure, could be done later.

src/tlo/analysis/utils.py
EvaJanouskova: Yes, it is. I use the suggested fnc now, thx!

EvaJanouskova:
@tamuri, Is this it or you still need to finish the review?

src/tlo/analysis/utils.py
EvaJanouskova: Btw, the `is_numeric_dtype` is underlined in pycharm when imported, and if I hold the mouse above it, it says:
`'is_numeric_dtype' is not declared in __all__`

Should I worry about it or it's fine?

EvaJanouskova:
_Eva's questions regarding the following tasks from TODO list:_
- [A ] test essential equipment set for all HSI events
- [B] test essential equipment is in RF_Equipment
- [C] test equipment potentially used is in RF_Equipment (if possible)

[A], I would like to test within the test_healthsystem whether the essential equipment is set for each HSI event. If the fnc self.set_equipment_essential_to_run_event is not used in __init__ of an HSI event, it won’t set up the self.ESSENTIAL_EQUIPMENT, and hence self.ESSENTIAL_EQUIPMENT won’t exist.
[B], I also want to check that all the items set as the essential equipment are included in the RF_Equipment.
[A & B] Should I use the functions from TLOmodel/src/tlo/analysis/hsi_events.py script to get all the names of HSI events and verify the existence of  self.ESSENTIAL_EQUIPMENT and presence of all items in RF for each event name? 

Is this: https://www.tlomodel.org/hsi_events.html the latest version from master?
Do we want to add there also the sets of essential equipment for the HSI events if any?
Should be the verification of self.ESSENTIAL_EQUIPMENT existence and presence in the RF done in test_healthsystem, or if we will add the ess. equip. into the hsi_event script anyway it could be done within the script?

[C] Could we anyhow test whether the equipment items potentially used (set within the apply with add_equipment or add_equipment_from_pkg fnc) are included in the RF_Equipment? Originally, we discussed with Matt to have a warning at the end of simulation about the items which were logged but not present in the RF. However, some changes have been made since then and now we reach the RF when updating the self.EQUIPMENT and self.ESSENTIAL_EQUIPMENT (to set them as the item codes, while the item names are provided in the modules), hence the simulation fails if a provided item name is not in the RF. But Tim H would prefer if it could be tested instead of letting the simulation fail. But I can't think of a way to do it. Do you have any idea?

EvaJanouskova:
_Points decided on hybrid teams & in-person meeting with Asif, Tim H, Eva, and Joe:_

> * [A ] test essential equipment set for all HSI events
> * [B] test essential equipment is in RF_Equipment
> * [C] test equipment potentially used is in RF_Equipment (if possible)
> 
> [A], I would like to test within the `test_healthsystem` whether the essential equipment is set for each HSI event. If the fnc `self.set_equipment_essential_to_run_event` is not used in `__init__` of an HSI event, it won’t set up the `self.ESSENTIAL_EQUIPMENT`, and hence `self.ESSENTIAL_EQUIPMENT` won’t exist.

--> _Both `self.EQUIPMENT` and `self.ESSENTIAL_EQUIPMENT` will be attributes of `HSI_Event` class, if `self.ESSENTIAL_EQUIPMENT` is still None when initialised, it will be silently set to empty set & warning and log will be executed at the end of simulation, informing on the names of the hsi events with missing settings of `self.ESSENTIAL_EQUIPMENT`.

> [B], I also want to check that all the items set as the essential equipment are included in the RF_Equipment.

--> If an equipment item name provided is not in RF_Equipment, it will be ignored (not used in simulation)  & warning and log will be executed at the end of simulation, informing on the equipment item names attempted to be used, but missing in the RF_Equipment.

> Is this: https://www.tlomodel.org/hsi_events.html the latest version from master? Do we want to add there also the sets of essential equipment for the HSI events if any? 

--> @tbhallett, I actually forgot to discuss this question during the meeting. I think it's question to you. If you like to have there the info on essential equipment as well, or not.

> [C] Could we anyhow test whether the equipment items potentially used (set within the apply with add_equipment or `add_equipment_from_pkg` fnc) are included in the RF_Equipment? Originally, we discussed with Matt to have a warning at the end of simulation about the items which were logged but not present in the RF. However, some changes have been made since then and now we reach the RF when updating the `self.EQUIPMENT` and `self.ESSENTIAL_EQUIPMENT` (to set them as the item codes, while the item names are provided in the modules), hence the simulation fails if a provided item name is not in the RF. But Tim H would prefer if it could be tested instead of letting the simulation fail. But I can't think of a way to do it. Do you have any idea?

--> Same solution as for [B].

tbhallett:
> Is this: https://www.tlomodel.org/hsi_events.html the latest version from master? Do we want to add there also the sets of essential equipment for the HSI events if any?

Yes, I think that would be nice, but not essential!

EvaJanouskova:
TODOs done and pushed in here:
- [x] <del>test essential equipment set for all HSI events</del>
        if essential equipment setting is missing, set it to empty set and add HSI event name to warning
- [x] <del>test essential equipment is in RF_Equipment</del>
        <del>test equipment potentially used is in RF_Equipment (if possible)</del>
        if any equipment item or pkg name provided in the code is not included in RF_Equipment, ignore the item(s) and add the item/pkg name to warning

EvaJanouskova:
Hi @tamuri, all we discussed is here now. Could you please review this? Thank you!

tamuri:
I think this looks fine. I checked whether the sets were sorted before logging/processing, which I think you've done everywhere.

tbhallett (changes requested):
Thanks Eva --

Here are some initial comments.

I will also push shortly a proposal for a set of tests, and a possible refactoring.

src/scripts/healthsystem/equipment/equipment_catalogue.py
tbhallett, line 1: I think this function should be in analysis.utils , as it won't be a script but a function

tbhallett:
Hi @EvaJanouskova 

Thanks for this.

I think we should work now to get the tests working. This will include making those scripts that read the logs into proper functions that we can call to work out what equipment has been used.

I think we should remove the declarations of equipment in the modules (and do that all in the subsequent PRs).

Once we have the tests (and so the whole workflow) going, then I think we'll do a refactor to put this all into one place (an Equipment class) so that the logic is all in one place.

src/tlo/methods/healthsystem.py
```diff
@@ -1006,6 +1073,42 @@ def get_beds_availability(self) -> str:
...
+        return _equip_availability
+
+    def get_equip_item_availability(self, equip_item_code: str) -> bool:
+        # TODO: update with implementation of essential equipment availability for the HSI event to run
+        #  for now, always available
+        if equip_item_code in [243, 41]:  # 243 = Pulse oximeter, 41 = 'Lamp, Anglepoise'
+            return False
+        return True  # True of False
+
+    def get_essential_equip_availability(self, essential_equip_set: Set[int]) -> bool:
```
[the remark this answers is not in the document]
EvaJanouskova: good idea

tests/test_healthsystem.py
```diff
@@ -1283,6 +1285,7 @@ def apply(self, population):
             _params['capabilities_coefficient'] = hs.capabilities_coefficient
             _params['cons_availability'] = hs.consumables.cons_availability
             _params['beds_availability'] = hs.bed_days.availability
+            _params['equip_availability'] = hs.equip_availability
```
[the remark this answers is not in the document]
EvaJanouskova: okay

EvaJanouskova (commented):
Hi @tbhallett,

Thank you for the review. See the responses to your comments:

tbhallett:
Thanks @EvaJanouskova  --- I've made updates to the overall structure to streamline and factorise into it's own class.

@willGraham01  -- as discussed Friday, I'd be very grateful for your review on this, particularly with an eye to making this not  incur any speed penalties to the overall simulation.

tbhallett (approved):

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,254 @@
...
+            pass
+        elif availability == "all":
+            df = (df + 1).clip(upper=1.0)  # All probabilities -> 1.0
+        elif availability == "none":
+            df = df.mul(0.0)  # All probabilities -> 0.0
+        else:
+            raise KeyError(f"Unknown equipment availability specified: {availability=}")
+
+        # Save
+        self._probabilities_of_items_available = df
```
willGraham01, line 149: Suggested changes in https://github.com/UCL/TLOmodel/pull/1342.

It looks like every time the equipment availability is updated, we are going to rebuild this dataframe from scratch which would likely slow us down a bit. 

In two of the cases ("all" and "none") we actually just want all the probabilities to be 1/0 respectively, and in the third case we can get away with computing the probabilities once (say at the start of the simulation) and then storing them rather than overwriting them with 1s and 0s.

So I think it might be better for us to store the "availability" (that gets passed on `__init__`) and slightly rework the probability method to check against this variable before trying to lookup equipment:

```python

def __init__(self, ...):
  # ...
  self.availabilty = "all"/"none"/"default"
  self._equip_probabilities_by_default = _set_equipment_items_available()
  # ...

def probability_all_equipment_available(self, ...):
  if self.availability = "all":
    return 1
  elif self.availability = "none":
    return 0
  else:
    return p # the value we currently compute, from looking up all availability probabilities and multiplying, etc.
```

We might also save a little bit of time in "all"/"none" by not having to query the dataframe (if it is big) too.

tbhallett:
Thanks again for really useful review @willGraham01 

@matt-graham  ---- please could you take one last look at this, especially on the 'interface'. I'd like to merge it so that we can finalise the next PR (#1341) which begins to use this for real.

src/tlo/methods/healthsystem.py
```diff
@@ -193,6 +194,27 @@ class HealthSystem(Module):
             "Availability of beds. If 'default' then use the availability specified in the ResourceFile; if "
             "'none', then let no beds be  ever be available; if 'all', then all beds are always available. NB. This "
             "parameter is over-ridden if an argument is provided to the module initialiser."),
+        'EquipmentCatalogue': Parameter(
+            Types.DATA_FRAME, "Data on equipment items and packages."),
+        'equipment_availability_estimates': Parameter(
+            Types.DATA_FRAME, "Data on the availability of equipment items and packages."
+        ),
```
willGraham01, line 201: Just another comment from me - having these two DataFrames as parameters means we're actually saving them twice (since we pass the DataFrames into the Equipment constructor, which then takes it's own copy of the DF).

I've opened https://github.com/UCL/TLOmodel/pull/1343 which removes these two parameters and instead just gets the Equipment class to read from the files directly. It also removes a few other methods from the HealthSystem class that aren't needed with this change.

Though if having these dataframes as parameters is required by the logging or output, feel free to ignore this suggestion!

matt-graham: > Just another comment from me - having these two DataFrames as parameters means we're actually saving them twice (since we pass the DataFrames into the Equipment constructor, which then takes it's own copy of the DF).

Is this definitely the case? As far as I can see the call to the `Equipment` initialiser just passes references to the parameter dictionary entries and the `Equipment` initialiser itself doesn't directly make any copies, it just attaches the passed references as attributes of the class? I think having these values as parameters of the `HealthSystem` module is consistent with the pattern generally used which is that any values read in from external resource files are parameters (and could feasibly be swapped for other values by changing the resource files or overriding the parameter values in a scenario (though currently the latter case isn't possibly because of the constraint to scalar valued parameters - see #988).

tbhallett: Yes, I was thinking the same as @matt-graham  here -- that we want all the reading to happen in `read_parameters` so that the `Scenario` class can overwrite anything afterwards.

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+
+        assert facility_id in self._all_fac_ids, f"Unrecognised facility ID: {facility_id=}"
+        assert item_codes.issubset(self._all_item_codes), f"At least one item code was unrecognised: {item_codes=}"
+
+        if self.availability == "all":
+            return 1.0
+        elif self.availability == "none":
+            return 0.0
+        return self._probabilities_of_items_available.loc[
+            (facility_id, list(item_codes))
```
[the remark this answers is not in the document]
matt-graham: While above comment is still correct (I think) because of how[ indexing / subscripting of classes in Python is handled](https://stackoverflow.com/a/1685450) I now realise that `_probabilities_of_items_available` is a multi indexed series and so the tuple here is possibly meant to specifically hint that we're passing in a multi index key rather than index / column specifiers (as for a `DataFrame.loc`), so possibly keeping parenthesis makes sense!

matt-graham (commented):
Adding some initial comments, will continue reviewing tomorrow!

src/tlo/methods/hsi_event.py
```diff
@@ -41,6 +41,7 @@ class HSIEventDetails(NamedTuple):
     facility_level: Optional[str]
     appt_footprint: Tuple[Tuple[str, int]]
     beddays_footprint: Tuple[Tuple[str, int]]
+    equipment: set
```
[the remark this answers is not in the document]
tbhallett: you're right. I had to change what I actually put in it to be a tuple, but didn't change the definition. Committing your change.

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+        dat = dat.groupby("Facility_ID").transform(lambda x: x.fillna(x.mean()))
+        # ... and also impute availability for any facility_ids for which no data, based on all other facilities
+        dat = dat.groupby("Item_Code").transform(lambda x: x.fillna(x.mean()))
+
+        # Check no missing values
+        assert not dat.isnull().any()
+
+        return dat
+
+    def parse_items(self, items: Union[int, str, Iterable[int | str]]) -> Set[int]:
```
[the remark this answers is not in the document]
tbhallett: Good to know, thanks

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+
+    :param: `rng`: The Random Number Generator object to use for random numbers.
+
+    :param: `availability`: Determines the mode availability of the equipment. If 'default' then use the availability
+     specified in the `data_availability`; if 'none', then let no equipment be ever be available; if 'all', then all
+     equipment is always available.
+
+    :param `: `master_facilities_list`: The pd.DataFrame with the line-list of all the facilities in the HealthSystem.
+
+    """
```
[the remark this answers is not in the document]
tbhallett: Thanks VERY much for that. Looking forward to seeing how it renders in the docs.

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def __init__(
+        self,
+        catalogue: pd.DataFrame,
+        data_availability: pd.DataFrame,
+        rng: np.random,
+        master_facilities_list: pd.DataFrame,
+        availability: Optional[Literal["all", "default", "none"]] = "default",
```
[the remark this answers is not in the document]
tbhallett: You're right, we can't let it be `None`. It should be:

```suggestion
availability: Literal["all", "default", "none"] = "default"
```

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+        items_are_ints = all(isinstance(element, int) for element in items)
+
+        if items_are_ints:
+            check_item_codes_recognised(items)
+            # In the return, any unrecognised item_codes are silently ignored.
+            return items.intersection(self._all_item_codes)
+        else:
+            check_item_descriptors_recognised(items)  # Warn for any unrecognised descriptors
+            # In the return, any unrecognised descriptors are silently ignored.
+            return set(filter(lambda item: item is not None, map(self._item_code_lookup.get, items)))
```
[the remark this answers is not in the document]
tbhallett: Happy for that. I had some misplaced feeling that using map might be faster than list comprehension, but I realise (thanks [stack overflow](https://stackoverflow.com/questions/1247486/list-comprehension-vs-map)) that the difference (especially on list of length 1 and 2) is microscopic.

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+
+        # Check no missing values
+        assert not dat.isnull().any()
+
+        return dat
+
+    def parse_items(self, items: Union[int, str, Iterable[int | str]]) -> Set[int]:
+        """Parse equipment items specified as an item_code (integer), an item descriptor (string), or an iterable of
+         either, and return as a set of item_code (integers). For any item_code/descriptor not recognised, a
+         `UserWarning` is issued."""
```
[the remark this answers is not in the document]
tbhallett: You're reading is correct --- if it's an interable, its elements have to all be the same type.

src/tlo/methods/equipment.py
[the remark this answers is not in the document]
tbhallett: Indeed, that was the intention. In data frame indexing, one has to use the tuple for MultiIndex, otherwise it's ambiguous whether you want to refer to a column... so I did the same thing here for readability (for myself). My preference would be to keep the parantheses.

src/tlo/methods/equipment.py
```diff
@@ -0,0 +1,278 @@
...
+        self, availability: Literal["all", "default", "none"]
+    ) -> None:
+        """
+        Update the availability of equipment.
+
+        This is expected to be called midway through the simulation if the
+        assumption of the equipment availability is changed.
+        """
+        assert availability in ["all", "none", "default"], f"New availability parameter {availability} not recognised."
+        self.availability = availability
```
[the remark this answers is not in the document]
tbhallett: The property method seems neater, I agree.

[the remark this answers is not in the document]
tbhallett: implementing that in next commit (hopefully done as intended). In init, I use the setter (`self.availablility = availability`) in order to get the benefit of the checks in the setter (rather than, as you say, storing to the private variable, `self._availability`) -- is that ok?

src/tlo/methods/equipment.py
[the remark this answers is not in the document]
tbhallett: done in next commit

src/tlo/methods/equipment.py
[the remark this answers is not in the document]
matt-graham: Yep that makes sense!

src/tlo/methods/equipment.py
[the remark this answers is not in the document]
matt-graham: The property implementation looks all correct and you're right that you would set using `self.availablility = availability` to make sure setter method is used rather than using underlying attribute `self._availability`. I was originally thinking here for this to replace the `update_availability` method though maybe that is not necessary as might involve a lot of downstream changes. If keeping the separate `update_availability` method then the separate assertion there is not needed as we will already hit the assert statement in the setter method before changing the value.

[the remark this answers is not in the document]
willGraham01: > In init, I use the setter (self.availablility = availability) in order to get the benefit of the checks in the setter (rather than, as you say, storing to the private variable, self._availability) -- is that ok?

Yep, should behave as intended :+1:

<details>
<summary>(Quick trial by testing)</summary>

```python

class Foo:

    _my_prop: int

    @property
    def my_prop(self) -> int:
        return self._my_prop

    @my_prop.setter
    def my_prop(self, val: int) -> None:
        assert isinstance(val, int) and 0 <= val < 10
        self._my_prop = val

    def __init__(self, prop_val: int = 3):
        self.my_prop = prop_val

f = Foo(prop_val=4) # Sets fine
ff = Foo(prop_val=40) # Assertion error

```
</details>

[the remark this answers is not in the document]
willGraham01: > `update_availability` method though maybe that is not necessary as might involve a lot of downstream changes

`update_availability` used to be [called here](https://github.com/UCL/TLOmodel/blob/3be83633970bd06ee609015c15c901af2b47543e/src/tlo/methods/healthsystem.py#L2781) but now we're referring to the `availability` property directly, we don't seem to need the method? CTRL+F on the codebase doesn't return any hits for `update_availability` anymore.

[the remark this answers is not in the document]
tbhallett: commit just pushed deletes it... 

(Sorry, my mistake. I realised we had deprecated it and used the new setter instead, but forgot to remove it).

tests/test_equipment.py
```diff
@@ -0,0 +1,414 @@
...
+
+    # Checks on parsing equipment items
+    # - using single integer for one item_code
+    assert {1} == eq_default.parse_items(1)
+    # - using list of integers for item_codes
+    assert {1, 2} == eq_default.parse_items([1, 2])
+    # - using single string for one item descriptor
+    assert eq_default.parse_items('ItemOne')
+    # - using list of strings for item descriptors
+    assert eq_default.parse_items(['ItemOne', 'ItemTwo'])
```
matt-graham, line 71: The assertions here seem to be just that the values returned by `parse_items` are not 'Falsey' (for example not an empty set). Could we not instead check that we get the expected set of item codes (that is I think `{1}` and `{1, 2}` respectively if I'm understanding how `parse_items` works correctly)?

tests/test_equipment.py
```diff
@@ -0,0 +1,414 @@
...
+from tlo.analysis.utils import parse_log_file
+from tlo.events import IndividualScopeEventMixin
+from tlo.methods import Metadata, demography, healthsystem
+from tlo.methods.equipment import Equipment
+from tlo.methods.hsi_event import HSI_Event
+
+resourcefilepath = Path(os.path.dirname(__file__)) / "../resources"
+
+
+def test_core_functionality_of_equipment_class(seed):
```
matt-graham, line 20: As a general comment on this test function - the checks look sensible here, but generally I would say better to have lots of smaller test functions with only a few assert statements rather than one big test function with lots of asserts as this means you get less granular information on test failure (as if one of the earlier asserts fails none of the subsequent asserts are run which means you have less information available for finding out the source of errors). To avoid having to repeatedly construct the `Equipment` class instance you could make this (and its arguments) fixtures, that is something like the following

```Python
@pytest.fixture
def catalogue():
    return pd.DataFrame(
        [
            {"Item_Description": "ItemZero", "Item_Code": 0, "Pkg_Name": float('nan')},
            {"Item_Description": "ItemOne", "Item_Code": 1, "Pkg_Name": float('nan')},
            {"Item_Description": "ItemTwo", "Item_Code": 2, "Pkg_Name": 'PkgWith2+3'},
            {"Item_Description": "ItemThree", "Item_Code": 3, "Pkg_Name": 'PkgWith2+3'},
        ]
    )

@pytest.fixture
def data_availability():
    return pd.DataFrame(
        # item 0 is not available anywhere; item 1 is available everywhere; item 2 is available only at facility_id=1
        # No data for fac_id=2
        [
            {"Item_Code": 0, "Facility_ID": 0, "Pr_Available": 0.0},
            {"Item_Code": 0, "Facility_ID": 1, "Pr_Available": 0.0},
            {"Item_Code": 1, "Facility_ID": 0, "Pr_Available": 1.0},
            {"Item_Code": 1, "Facility_ID": 1, "Pr_Available": 1.0},
            {"Item_Code": 2, "Facility_ID": 0, "Pr_Available": 0.0},
            {"Item_Code": 2, "Facility_ID": 1, "Pr_Available": 1.0},
        ]
    )

@pytest.fixture
def master_facilities_list():
    return pd.DataFrame(
        [
            {'District': 'D0', 'Facility_Level': '1a', 'Region': 'R0', 'Facility_ID': 0, 'Facility_Name': 'Fac0'},
            {'District': 'D0', 'Facility_Level': '1b', 'Region': 'R0', 'Facility_ID': 1, 'Facility_Name': 'Fac1'},
            {'District': 'D0', 'Facility_Level': '2', 'Region': 'R0', 'Facility_ID': 2, 'Facility_Name': 'Fac2'},
        ]
    )

@pytest.fixture
def rng(seed):
    return np.random.RandomState(np.random.MT19937(np.random.SeedSequence(seed)))

@pytest.fixture
def equipment_default(catalogue, data_availability, master_facilities_list, rng):
    return Equipment(
        catalogue=catalogue,
        data_availability=data_availability,
        rng=rng,
        master_facilities_list=master_facilities_list,
        availability="default",
    )

@pytest.fixture
def equipment_default(catalogue, data_availability, master_facilities_list, rng):
    return Equipment(
        catalogue=catalogue,
        data_availability=data_availability,
        rng=rng,
        master_facilities_list=master_facilities_list,
        availability="default",
    )

@pytest.fixture
def equipment_all(catalogue, data_availability, master_facilities_list, rng):
    return Equipment(
        catalogue=catalogue,
        data_availability=data_availability,
        rng=rng,
        master_facilities_list=master_facilities_list,
        availability="all",
    )

def test_parse_single_item_code(equipment_default):
    assert {1} == equipment_default.parse_items(1)

def test_parse_iterable_of_item_codes(equipment_default):
    assert {1, 2} == equipment_default.parse_items([1, 2])

def test_parse_single_item_descriptor(equipment_default):
    assert {1} == equipment_default.parse_items('ItemOne')

def test_parse_iterable_of_item_descriptors(equipment_default):
    assert {1, 2} == equipment_default.parse_items(['ItemOne', 'ItemTwo'])

@pytest.mark.parametrize("container_type", [list, tuple, set])
def test_parse_empty_items(equipment_default, container_type):
    assert set() == eq_default.parse_items(container_type())

def test_parse_unrecognised_item_code_warns(equipment_default):
    with pytest.warns():
        equipment_default.parse_items(10001)

def test_parse_unrecognised_item_descriptor_warns(equipment_default):
    with pytest.warns():
        equipment_default.parse_items('ItemThatIsNotDefined')

...
```

Possibly not worth the effort doing this here to avoid delaying getting this PR merged in further, but thought it might be useful to note for future!

matt-graham (changes requested):
Interface generally looks good to me. I've added some further suggested changes and comments below, but the only one I think is vital is around changing `HSI_Event._EQUIPMENT` from being a class attribute to an instance attribute unless the former was specifically intended.

src/tlo/methods/hsi_event.py
```diff
@@ -85,6 +86,12 @@ class HSI_Event:
     # which have been loaded.
     BEDDAYS_FOOTPRINT: Dict[str, Union[float, int]]
 
+    _EQUIPMENT: Set[int] = set()  # The set of equipment that is used in the HSI. If any items in this set are not
+    #                               available at the point when the HSI will be run, then the HSI is not run, and the
+    #                               `never_ran` method is called instead. This is a declaration of resource needs, but
+    #                               is private because users are expected to use `add_equipment` to declare equipment
+    #                               needs.
+
```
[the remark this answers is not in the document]
tbhallett: Thanks. We absolutely don't want that behaviour. I'm making the change and will also add a test to check that the equipment list for each HSI_Event is kept separate.

tests/test_equipment.py
tbhallett: Thanks Matt - that's really useful to know. As you guessed, I don't think I want to make that change to these tests now (though I am tempted as you've made it look nice and easy), in the interest of getting this in and carrying on with defining the actual items... but it's certainly noted for next time. I can certainly see the advantages of doing it this way.

tbhallett:
bahhhh-- checks failing due to `ruff` errors on file I haven't touched in this PR.... 🤪

matt-graham:
> bahhhh-- checks failing due to `ruff` errors on file I haven't touched in this PR.... 🤪

I'll take a look - if specific rules are causing false positives I can set them to be ignored.

tbhallett:
> > > bahhhh-- checks failing due to `ruff` errors on file I haven't touched in this PR.... 🤪
> > 
> > 
> > I'll take a look - if specific rules are causing false positives I can set them to be ignored.
> 
> Looks like this is due a new `pylint` release yesterday (version 3.2.0) which [added a new `possibly-used-before-assignment` check](https://github.com/pylint-dev/pylint/releases/tag/v3.2.0) as the [scheduled checks workflow on `master` also failed overnight](https://github.com/UCL/TLOmodel/actions/runs/9087784671/job/24976165364). I'll check whether any of the occurences flagged are actually a problem and either fix or disable the rule.

That makes sense -- thanks very much for this.

matt-graham:
@tbhallett #1349 should fix the `pylint` errors when running checks on both `master` and here

matt-graham (approved):
Given the failing checks here are not to do with anything in this PR as they are also failing on `master` and should be fixed by #1349, I would say this is good to go in now.

Edit: A but we won't be able to do without disabling the status check, so maybe we do need to merge in #1349 first.

tbhallett:
Thanks Matt. I’ll merge it in.

Sent from Outlook for iOS<https://aka.ms/o0ukef>

EvaJanouskova:
Thank you all for your amazing work! This looks so much better than it ever did. I love it!

Thank you.

PR #1099, opened 2023-09-06
contraception health impacts: scenario

EvaJanouskova:
**Scenario to run some first jobs:**

All loggings, modules, and draw_parameters as in Joe's [baseline_scenario](https://github.com/UCL/TLOmodel/blob/301e9419ebfa83ee220002cc1fb58513dcd715b2/src/scripts/maternal_perinatal_analyses/scenario_files/baseline_scenario.py).

Only, I have removed arguments for HealthSystem: mode_appt_constraints=1, cons_availability='default'. As those are default values anyway.

### DIFF
--- a/src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
+++ b/src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
@@ -1,0 +1,124 @@
+"""
+This file defines a batch run to get sims results to be used by an analysis script
+( # TODO: create one and add its name here
+).
+Run on the remote batch system using:
+```tlo batch-submit src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py```
+or locally using:
+```tlo scenario-run src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py```
+
+# TODO: update
+SCENARIO SETTINGS
+-----------------
+used modules:
+...
+
+# TODO: update
+FP INTERVENTIONS' IMPLICATIONS FOR MNCH PAPER (Eva J et al. 2024):
+------------------------------------------------------------------
+...
+"""
+
+from tlo import Date, logging
+from tlo.methods import (
+    alri,
+    cardio_metabolic_disorders,
+    care_of_women_during_pregnancy,
+    contraception,
+    demography,
+    depression,
+    diarrhoea,
+    enhanced_lifestyle,
+    epi,
+    healthburden,
+    healthseekingbehaviour,
+    healthsystem,
+    hiv,
+    labour,
+    malaria,
+    newborn_outcomes,
+    postnatal_supervisor,
+    pregnancy_supervisor,
+    stunting,
+    symptommanager,
+    tb,
+    wasting,
+)
+from tlo.scenario import BaseScenario
+
+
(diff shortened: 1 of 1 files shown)


EvaJanouskova:
@joehcollins, Feel free (once you are available, finished with your thesis) to go through the scenario, and if you agree there is everything we need to run some first jobs - please merge into our latest common branch `origin/co_health_impact/2023_09`.

src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
```diff
@@ -0,0 +1,124 @@
+"""
+This file defines a batch run to get sims results to be used by an analysis script
+( # TODO: create one and add its name here
```
tbhallett, line 3: Todo.....?

src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
```diff
@@ -0,0 +1,124 @@
...
+                "tlo.methods.newborn_outcomes": logging.INFO,
+                "tlo.methods.care_of_women_during_pregnancy": logging.INFO,
+                "tlo.methods.pregnancy_supervisor": logging.INFO,
+                "tlo.methods.postnatal_supervisor": logging.INFO,
+                "tlo.methods.tb": logging.INFO,
+            }
+        }
+
+    def modules(self):
+        return [
```
tbhallett, line 86: Why not use fullmodel ?

tbhallett (approved):

tbhallett:
@EvaJanouskova is this done?
Do you want this Pr to go into master? Or shall we close it...?

src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
EvaJanouskova: we didn't start working on the analysis script yet, and now this project was postponed

src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
EvaJanouskova: I think @joehcollins  mentioned that we need pretty large population to be able to see smt, so at least for preliminary results it didn't make sense to use fullmodel, is that correct @joehcollins?

EvaJanouskova:
> @EvaJanouskova is this done? Do you want this Pr to go into master? Or shall we close it...?

This was awaiting review from @joehcollins. We still can save it for future work, ie put it in master, if @joehcollins is able to look at this, shouldn't take much time.

src/scripts/contraception/scenarios/run_analysis_co_health_impacts_no_diseases.py
joehcollins: i think this was a direct copy from the scripts i was using for my PhD - we didnt use the full model then because of the run time. I think going forward with this analysis we would want to use the full model

joehcollins:
@EvaJanouskova I would maybe lean towards just closing this for now instead of just adding to master until were are sure we're going to do this analysis?

tbhallett:
Let's make it a draft then and not put it in any merging queues, for now.

PR #1100, opened 2023-09-07
Delegate upload of profiling results to GH runners

willGraham01:
The self-hosted runners do not have `Docker` installed on the system, which the action for uploading the profiling results to the profiling repository depends on.

This changes the `run-profiling` workflow to have two composite jobs. 
1. The first runs on our self-hosted runners, and runs the simulation to be profiled. Once the profiling run has been performed, the results are uploaded as an artifact.
2. Once step 1 completes, a second job is triggered on a GH-actions `ubuntu` runner to pull the artifact just created, then move it to the profiling results repository.

This means that the profiling workflow releases the self-hosted machines slightly earlier, and we don't waste some Azure compute time handling the transfer of files between repositories. The created artifacts all have unique names, and are set to auto-delete after 1 day.


.github/workflows/run-profiling.yaml
```diff
@@ -24,6 +24,25 @@ jobs:
           UNIQUE_ID=${GITHUB_EVENT_NAME}_${GITHUB_RUN_NUMBER}_${GITHUB_SHA} >> $GITHUB_ENV
```
matt-graham, line 24: Realise this is not a change from this PR, but I don't think the `>> $GITHUB_ENV` is necessary here. It looks like you only need this environment variable locally in this run step. The `GITHUB_ENV` variable points to a file that can be used to share variables [across job steps](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable) and as its a path to a file the syntax would be
```
echo "{environment_variable_name}={value}" >> "$GITHUB_ENV"
```
I don't think that assigning to a variable outputs anything to `stdout` so I think the above will currently have no effect (appending an empty string to `$GITHUB_ENV` file).

matt-graham (approved):
Looks like a good solution to problem to me! Just one minor comment with regards to use of `$GITHUB_ENV`.

PR #1101, opened 2023-09-07
malaria hiv interactions

tdm32:
This branch incorporates interactions between malaria and HIV. 

Specific changes are:

- Include property for cotrimoxazole (BOOL) so we can include protective effects of this on malaria
- Edit ART dispensation to track cotrim as well
- Cotrim is optional item in consumables check
- Only give cotrim if not on IPTp already (pregnant)
- Only give IPTp if not on cotrim - add in check
- Add linear models to incorporate individual risk for clinical and severe malaria
- Add effects of HIV to malaria incidence draws
- Add effects of HIV to malaria severity draws
- cotrim and IPTp now give partial protection against clinical/severe malaria through linear model
- End protection of IPTp after 6weeks - change property
- Edit malaria tests to pass under new edits

Plots below are small population (2000) and show some stochastic noise but are within reasonable tolerance of reported data.

image
image
image
image


tbhallett:
Re-running tests, as they failed on the last run (but I think due to Azure side things)

tdm32:
> Wow. It did!

Ah great thanks Tim.

tdm32:
@tbhallett I'm just working on this - in another branch - where I'm decoupling malaria from HIV and TB. Any conditions which are reliant on properties set by module Hiv now have checks, so you can run malaria with or without Hiv

tbhallett:
Thanks. So this is ready for review and merging?

tdm32:
> Thanks. So this is ready for review and merging?

Yes please

src/tlo/methods/hiv.py
```diff
@@ -2534,13 +2541,15 @@ def do_at_initiation(self, person_id):
         df = self.sim.population.props
         person = df.loc[person_id]
 
-        # Check if drugs are available, and provide drugs:
+        # Check if drugs are available, and provide drugs
```
tbhallett, line 2544: It would be safer for this to be dict or named tuple so that we don't have to remember the special ordering. 
Ie for example return
{"art": True, "contrim": False}

src/tlo/methods/malaria.py
```diff
@@ -1334,6 +1491,39 @@ def apply(self, population):
...
+                df[df.is_alive & (df.ma_inf_type != 'none') & df.hv_inf]
+            )
+
+            # hiv population
+            hiv_infected = len(
+                df[df.is_alive & df.hv_inf]
+            )
+
+            # prevalence of malaria in HIV population
+            prev_malaria_in_hiv_population = coinfection_num / hiv_infected
```
tbhallett, line 1508: Careful about division of values than could be 0. Use safety around this: eg

X if y > 0 else 0

src/tlo/methods/malaria.py
```diff
@@ -1334,6 +1491,39 @@ def apply(self, population):
...
+
+            # proportion of malaria cases with concurrent HIV infection
+            malaria_infected = len(
+                df[df.is_alive & (df.ma_inf_type != 'none')]
+            )
+
+            prop_malaria_cases_with_hiv = coinfection_num / malaria_infected
```
tbhallett, line 1515: Same as above and elsewhere

tbhallett (changes requested):

src/tlo/methods/hiv.py
tdm32: I've changed this

src/tlo/methods/malaria.py
tdm32: changed

src/tlo/methods/malaria.py
tdm32: changed here and wherever appropriate

tdm32:
@tbhallett the suggested changes have been made and this is ready for final review. Thanks very much!

tbhallett:
Thanks @tdm32  --- there's a conflict on this branch now. (ResourceFile_malaria.xlsx ... so it's hard for me to work out which version to choose). Please could you resolve through merging-in `master`?

tdm32:
> Thanks @tdm32 --- there's a conflict on this branch now. (ResourceFile_malaria.xlsx ... so it's hard for me to work out which version to choose). Please could you resolve through merging-in `master`?

Hi @tbhallett, I've merged in master this morning. The conflict arose I think because the file location moved. The correct version is in this branch.

tbhallett (approved):

PR #1103, opened 2023-09-11
Add clarifying comment and extra test to `mix_scenarios`

tbhallett:
Following a comment by @marghe-molaro , it seems like the comments on this function should be more clear about how the situation is handled when a parameter is to be over-written more than once.


src/tlo/analysis/utils.py
```diff
@@ -1218,7 +1218,8 @@ def construct_multiindex_if_implied(df):
 
 def mix_scenarios(*dicts) -> Dict:
     """Helper function to combine a Dicts that show which parameters should be over-written.
-     * Warnings are generated if a parameter appears in more than one Dict with a different value;
+     * If a parameter appears in more than one Dict, the value in the "right-most" dict is taken, and a UserWarning
```
tamuri, line 1221: I know I'm being unnecessarily pedantic, but this should really be "last added" to the dictionary, not  right-most.

marghe-molaro (approved):

PR #1105, opened 2023-09-12
omnibus comment changes

EvaJanouskova:
Fixed a few typos I came across.

Updated/moved some comments.

### DIFF
--- a/src/scripts/data_file_processing/formatting_demography_data.py
+++ b/src/scripts/data_file_processing/formatting_demography_data.py
@@ -69,7 +69,7 @@ a1.loc[region_names, 'Region'] = region_names
 a1['Region'] = a1['Region'].ffill()
 a1 = a1.drop(region_names)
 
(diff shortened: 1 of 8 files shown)


tbhallett:
Thanks for this @EvaJanouskova  -- I'll leave this open a little longer in case other small things need to be changed.

tbhallett:
> @tbhallett, The description at the top of `hsi_generic_first_appts.py` seems outdated, doesn't it?

Yes it is a bit, you’re right. It should explain that those appointments can run at different levels.

tbhallett:
thanks for all these @EvaJanouskova --- is it ready to go in?

EvaJanouskova:
> thanks for all these @EvaJanouskova --- is it ready to go in?

There will always be smt, but it is all I have right now, so I think it can go in.

tbhallett (approved):

PR #1107, opened 2023-09-13
BUG FIX [not urgent]: severe malaria cases may not be resolved if mortality rate changes

tdm32:
This bug does not arise in any of the current model configurations and does not require an urgent fix.

Malaria module:
If in future the case fatality rate or mortality adjustment parameter is modified such that not all severe malaria cases are routed through the MalariaDeathEvent, any cases which do not have a scheduled death will not have their symptoms resolved after treatment. They will remain indefinitely on treatment. 

Currently, through the combination of cfr and mortality_adjustment, which are calibrated to match WHO estimates, all severe malaria cases have a scheduled death and this is modified by appropriate treatment in MalariaDeathEvent. If an individual does not die, they will be cured (symptoms resolved and treatment status reset to 'none'). 

If we change the cfr, for example, to have higher rates for children and lower rates for adults, some severe cases will not be routed through the MalariaDeathEvent and these cases will not have symptoms resolved at any point. 

The solution is to change the property ma_tx to a categorical variable indicating which treatment regimen a person is using, then add an additional treatment resolution code snippet to allow any severe cases who are not routed through the death event, to have their symptoms resolved. 

This bug will not affect any of the current runs and the fix does not affect any of the outputs. Results from a short run are below. 

image


tbhallett (approved):

tbhallett:
@tdm32  - please could you merge-in `master` to this one and resolve conflicts

tbhallett (approved):

PR #1109, opened 2023-09-14
Proposed Fix to Runtime Warnings from `np.log`

### LINKED ISSUE #1108: RuntimeWarnings
I'm getting (in the stderr.txt with master version before 4th September) ~300 of RuntimeWarnings per one run of long_run_all_diseases:
```
/usr/local/lib/python3.8/site-packages/pandas/core/arraylike.py:358: RuntimeWarning: divide by zero encountered in log
  result = getattr(ufunc, method)(*inputs, **kwargs)
```

A numpy error?
Asif mentioned that only np.log operation is in `stunting`.
Tim H guesses that changing the value to 0.9999 in clip at line 487 in `stunting.py` would solve it.

tbhallett:
(Hopefully) Fixes #1108

### DIFF
--- a/src/tlo/methods/stunting.py
+++ b/src/tlo/methods/stunting.py
@@ -484,8 +484,7 @@ class StuntingPollingEvent(RegularEvent, PopulationScopeEventMixin):
         df = self.sim.population.props
         rng = self.module.rng
 
-        annual_prob = model.predict(df.loc[mask]).clip(upper=0.99999)  # Prevents the np.log call below raising warning
-        #                                                                if it encounters np.log(0.0).
+        annual_prob = model.predict(df.loc[mask]).clip(upper=1.0)
         cum_prob_over_days_exposed = 1.0 - (1.0 - annual_prob) ** (days_exposed_to_risk / DAYS_IN_YEAR)
 
         assert pd.notnull(cum_prob_over_days_exposed).all()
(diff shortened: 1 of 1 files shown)


src/tlo/methods/stunting.py
```diff
@@ -484,7 +484,8 @@ def apply_model(self, model, mask, days_exposed_to_risk):
         df = self.sim.population.props
         rng = self.module.rng
 
-        annual_prob = model.predict(df.loc[mask]).clip(upper=1.0)
+        annual_prob = model.predict(df.loc[mask]).clip(upper=0.99999)  # Prevents the np.log call below raising warning
+        #                                                                if it encounters np.log(0.0).
         cum_prob_over_days_exposed = 1.0 - np.exp(np.log(1.0 - annual_prob) * days_exposed_to_risk / DAYS_IN_YEAR)
```
matt-graham, line 489: Is there a reason for favouring the form here over the more direct and I think equivalent

```suggestion
        cum_prob_over_days_exposed = 1.0 - (1.0 - annual_prob) ** (days_exposed_to_risk / DAYS_IN_YEAR)
```
(as $\exp(\log(1-p) \times r) \equiv (1-p)^r$).

Switching to the above form I think would eliminate the need to clip at 0.99999 as we would not be computing a logarithm, and I think should be relatively numerically stable given `annual_prob` is by definition in range `[0, 1]`.

tbhallett: Good point. I think it is actually more correct.

matt-graham (approved):
Assuming tests pass this looks good to merge from my perspective :) !

PR #1110, opened 2023-09-15
Capture profiling statistics

willGraham01:
Almost closes issue #686 - we just need to decide on the parameters that we should pass to the `scale_run` script itself, and to capture the extra statistics that are listed in the issue itself.

## `ipysession` files are no longer produced

This alleviates the issue of filesizes of the `.pyisession` outputs. They are no longer pushed to the profiling repository - instead we push the `stats.json` files (discussed below) and optionally the rendered HTML files of the session results.

The HTML outputs are significantly smaller than the raw `.pyisession` outputs (of the order 10/100s kBs rather than 100s/1000s MBs).

## Profiling captures additional information on top of the profiling output

The profiling runs are now setup to output a `.stats.json` file upon completion, which captures information from the profiling session as well as additional information about the simulation itself. The output file is just a `json` file and can be parsed as such. Currently we are capturing:
- Start time of the profiling session
- Duration of the profiling session
- CPU time used by the profiled program
- Information about the final state of the population DataFrame from the simulation:
  - Number of rows and columns
  - Size in MBs
  - Number of times it was extended
- Disk I/O stats:
  - Number of read/writes
  - Size (in MB) of read/writes
  - Time spent reading/writing from disk
- The name of the HTML file produced, provided it was produced

There is scope to include additional statistics that are of interest, which can be done by adding additional fields to the dictionaries that are produced in the `record_XXX_statistics` functions.

The `run_profiling.py` script can also be passed the `--html` flag to also produce a HTML file containing the results of the profiling session (as rendered by pyinstrument) if we so choose.
Additionally, `run_profiling.py` can also be called with an `--additional-stats` flag, which can be used to pass shell or workflow variables to the program which will then also be included in the `.stats.json` output. For example,
```bash
python run_profiling.py --additional-stats foo="bar bar" "fave number"=5
```
will result in the following key/value pairs appearing in the output:
```json
"foo": "bar bar",
"fave number": "5",
```

## Sample output:

Run on commit `3f3507a23483fa5e373d9b518968aefb8c84cbff`, manually passing the `sha` and `trigger` keys.
```bash
tox -vv -e profile -- --additional-stats \
          sha=3f3507a23483fa5e373d9b518968aefb8c84cbff \
          trigger="manual trigger"
```
producing the `.stats.json` file:
```json
{
  "sha": "3f3507a23483fa5e373d9b518968aefb8c84cbff",
  "trigger": "manual test",
  "html_output": "manual_test_0_3f3507a23483fa5e373d9b518968aefb8c84cbff.html",
  "start_time": 1695219849.2109582,
  "duration": 425.1599588394165,
  "cpu_time": 424.63759634,
  "pop_df_rows": 51000,
  "pop_df_cols": 448,
  "pop_df_mem_mb": 107.226341,
  "pop_df_times_extended": 1,
  "disk_reads": 96,
  "disk_writes": 6097,
  "disk_read_MB": 5.726208,
  "disk_write_MB": 441.458688,
  "disk_read_s": 0.031,
  "disk_write_s": 2.343
}
```


src/scripts/profiling/run_profiling.py
```diff
@@ -43,10 +89,12 @@ def run_profiling(
         output_pyis_file = output_dir / "output.pyisession"
         output_html_file = output_dir / "output.html"
         output_json_file = output_dir / "output.json"
+        output_stat_file = output_dir / "output.stats.json"
```
matt-graham, line 92: Can't suggest change here as not all lines touched in this PR, but could avoid the repetition / need to keep the extensions in sync here by doing something like
```Python
output_stem = "output" if output_name is None else output_name.stem
output_pyis_file = output_dir / f"{output_stem }.pyisession"
output_html_file = output_dir / f"{output_stem }.html"
output_json_file = output_dir / f"{output_stem }.json"
output_stat_file = output_dir / f"{output_stem}.stats.json"
```

src/scripts/profiling/run_profiling.py
```diff
@@ -22,14 +26,56 @@ def current_time(formatstr: str = "%Y-%m-%d_%H%M") -> str:
...
+    # Population dataframe statistics
+    pops = s.population
+    stats_to_record["pop_df_rows"] = pops.props.shape[0]
+    stats_to_record["pop_df_cols"] = pops.props.shape[1]
+    stats_to_record["pop_df_mem_mb"] = (
+        pops.props.memory_usage(index=True, deep=True).sum() / 1e6
+    )
+    stats_to_record["pop_df_times_extended"] = int(
+        np.ceil((pops.props.shape[0] - pops.initial_size) / pops.new_rows.shape[0])
+    )
```
matt-graham, line 58: Minor nitpick but I would say using a dict literal here is a bit cleaner as avoids repetition of `stats_to_record` variable name and [should be a bit more performant](https://gist.github.com/hofrob/ad143aaa84c096f42489c2520a3875f9) (not that performance here is particularly important)

```suggestion
    # Population dataframe statistics
    pops = s.population
    stats_to_record = {
        "pop_df_rows": pops.props.shape[0],
        "pop_df_cols":  pops.props.shape[1],
        "pop_df_mem_mb": (
            pops.props.memory_usage(index=True, deep=True).sum() / 1e6
        ),
        "pop_df_times_extended": int(
            np.ceil((pops.props.shape[0] - pops.initial_size) / pops.new_rows.shape[0])
        )
    }
```

src/scripts/profiling/run_profiling.py
```diff
@@ -1,53 +1,232 @@
...
+    disk_stats = {
+        "disk_reads": disk_usage["read_count"],
+        "disk_writes": disk_usage["write_count"],
+        "disk_read_MB": disk_usage["read_bytes"] / 1e6,
+        "disk_write_MB": disk_usage["write_bytes"] / 1e6,
+        "disk_read_s": disk_usage["read_time"] / 1e3,
+        "disk_write_s": disk_usage["write_time"] / 1e3,
+    }
+
+    return disk_stats
```
matt-graham, line 122:
```suggestion
    if disk_usage is None:
        return {}

    return {
        "disk_reads": disk_usage["read_count"],
        "disk_writes": disk_usage["write_count"],
        "disk_read_MB": disk_usage["read_bytes"] / 1e6,
        "disk_write_MB": disk_usage["write_bytes"] / 1e6,
        "disk_read_s": disk_usage["read_time"] / 1e3,
        "disk_write_s": disk_usage["write_time"] / 1e3,
    }
```

src/scripts/profiling/run_profiling.py
```diff
@@ -1,53 +1,232 @@
...
+
+def record_profiling_session_statistics(
+    session: Session,
+) -> Dict[str, Union[int, float]]:
+    """
+    Extract important profiling statistics from the session that was captured.
+    Statistics are returned as a dictionary.
+
+                "start_time": self.start_time,
+            "duration": self.duration,
```
matt-graham, line 133:
```suggestion
            "start_time": self.start_time,
            "duration": self.duration,
```

src/scripts/profiling/run_profiling.py
```diff
@@ -1,53 +1,232 @@
...
+    if session is None:
+        return dict()
+
+    profiler_stats = {
+        "start_time": session.start_time,
+        "duration": session.duration,
+        "cpu_time": session.cpu_time,
+    }
+
+    return profiler_stats
```
matt-graham, line 158:
```suggestion
    if session is None:
        return {}

    return {
        "start_time": session.start_time,
        "duration": session.duration,
        "cpu_time": session.cpu_time,
    }
```

src/scripts/profiling/run_profiling.py
```diff
@@ -1,53 +1,232 @@
...
+    output_file: str,
+    html_output_file: str = None,
+    profiling_session: Session = None,
+    completed_sim: Simulation = None,
+    disk_usage: Dict[str, Union[int, float]] = None,
+    **additional_stats: str,
+) -> None:
+    """
+    Organise all statistics to be collected from the profiling run into a single dict,
+    which can then be dumped to a JSON file.
```
matt-graham, line 171:
```suggestion
    which is then be dumped to a JSON file.
```

`can then be` originally made me think the dumping to JSON file happened elsewhere.

src/scripts/profiling/run_profiling.py
```diff
@@ -1,53 +1,232 @@
...
     write_html: bool = False,
-    write_json: bool = False,
     interval: float = 1e-1,
+    **additional_stats: str,
 ) -> None:
+    """
+    Uses pyinstrument to profile the scale_run simulation,
+    writing the output in the requested formats.
+    """
     # Suppress "ignore" warnings
```
matt-graham, line 227:
```suggestion
```
Comment is a bit misleading I think - the `"ignore"` argument to `warnings.filterwarnings` is the `action` to take not a specification of the warnings to filter for (so the call just specifies to ignore all warnings).

matt-graham (approved):
Looks great - thanks @willGraham01. I've made a few minor suggestions but nothing that strictly needs changing.

matt-graham:
I've updated branch to change default values of profiling run parameters to be for 5 years, 50k initial population in mode 2, and now have the arguments to `scale_run` saved to profiling results directory as a JSON file along with the actual simulation outputs (by default only log output, with logging level set to `WARNING` to keep minimal). I've also done a bit of refactoring to make the `run_profiling` script a bit more flexible for using at the command line by exposing some of the key simulation parameters (e.g. simulation period, initial population) and added back an option (disabled by default) to save raw `.pyisession` output for when running locally. From my perspective assuming tests pass this is now good to go in. @tamuri tagging you in case you want to review given I've a series of changes on top of Will's.

Issue #763, opened 2022-11-22
Update Python and packages


tamuri:
*WARNING: Don't do this until it's confirmed everything works!*

We need to:

1. Update to a recent version of Python (3.11? 3.10?)
2. Update the package dependencies.

We've done this before in #144.

tamuri:
Notes for upgrade to Python 3.11. We'll need to:

*   update self-hosted runners' provisioning playbooks. 
*   container used by Azure Batch
*   TLOmodel README installation instructions
*   tox.ini

tamuri:
This particular upgrade is complete.

PR #1112, opened 2023-09-18
Update dependencies to Pandas v2.0

matt-graham:
Stop-gap solution as part of #763 which just updates existing `pip` based requirement files rather than also switching to fully `conda` based approach as in #949.

~~Given Pandas v2.1 has now been released assuming it is probably worth updating to use this assuming there aren't any major additional breaking changes over switch to v2.0 ([which it doesn't look like there is from release notes](https://pandas.pydata.org/docs/whatsnew/v2.1.0.html#backwards-incompatible-api-changes)).~~

Edit: Pandas v2.1 only support Python 3.9+ so rolling back pin to Panda v2.0 to maintain Python 3.8 compatibility for now.

From running tests locally it looks like at least some tests fail due to changes which are not Panda v2.0 compatible since I last worked on this, so will need to find and fix all of these first so marking this as draft for now.


matt-graham:
All the failures seem to be due to #1113 which is fixed by #1114 so 🤞 tests here will pass once that is merged in.

matt-graham:
Tests now all passing with #1142 merged 🎉

tamuri (approved):
Excellent, sterling effort!

PR #1116, opened 2023-09-19
Direct actions to job outputs

willGraham01:
`needs.<dependent-job-id>.outputs.<output-name>` is apparently the syntax, rather than just `needs.<dependent-job-id>.<output-name>`.


matt-graham (approved):
Good catch and sorry for suggesting the wrong syntax to begin with!

PR #1118, opened 2023-09-19
Change `||` to `&&` in `run-profiling`

willGraham01:
Another logic twist that slipped under the radar. Although the "failed workflow" does at least indicate that the hack we're using is working properly!

This error is setting off a bunch of failing workflows though :sweat_smile:


matt-graham (approved):
Eeek sorry for missing this when reviewing (and possibly introducing the error in my suggestion!). Like you say though good to know the approach is (sort of) working.

PR #1121, opened 2023-09-20
Fix getting PR branch tip SHA for comment triggered workflows

### LINKED ISSUE #1119: Comment triggered workflow runs on pull requests using wrong SHA
The comment triggered workflows are picking up the wrong commit SHA to checkout when triggered by a pull request comment where they should use the pull request branch head commit, but instead they are always using the latest commit on the default branch - for example [this workflow run](https://github.com/UCL/TLOmodel/actions/runs/6232713595/job/16916431850) triggered by [this comment](https://github.com/UCL/TLOmodel/pull/1114#issuecomment-1725005414) checked out 12f6cc5ec389257e60d4a2fde9afa77664baab0d (which was the latest commit on `master` at the time of the comment) rather than 69c2c5ab8fcfd8cde6c2129bd65288efa2016f61. 

Looking through [the repository](https://github.com/matt-graham/pr-comment-workflows) I was using for testing the workflow, it looks like this also occurred for the PR comment which triggered [this results comment](https://github.com/matt-graham/pr-comment-workflows/pull/1#issuecomment-1330533804) but not for [the preceding results comment](https://github.com/matt-graham/pr-comment-workflows/pull/1#issuecomment-1275966546) which checked out the correct PR head commit. Looking at the changes to the workflow in between these comments, it seems the problem is probably due to https://github.com/matt-graham/pr-comment-workflows/commit/4bcb875a30edf45c519ec188123471283e39895b which extended the workflow logic to allow also triggering from issue comments. In particular I think the initial `if` clause in
https://github.com/UCL/TLOmodel/blob/b9e157f9562c2fdcd0115bb01c076d8a759d432c/.github/workflows/run-on-comment.yml#L106-L108
is always evaluating to true which suggests `context.payload.pull_request` is not a valid approach for checking if the event was triggered by a PR comment rather than issue comment.

matt-graham:
Fixes #1119 

Also updates version of `workflow-application-token-action` action used to avoid warning about using deprecated `set-output` API.

### DIFF
--- a/.github/workflows/run-on-comment.yml
+++ b/.github/workflows/run-on-comment.yml
@@ -103,7 +103,7 @@ jobs:
       with:
         result-encoding: string
         script: |
-          if (!context.payload.pull_request) {
+          if (!context.payload.issue.pull_request) {
             return context.sha;
           };
           const { data: pr } = await github.rest.pulls.get({
@@ -125,7 +125,7 @@ jobs:
     - name: Get comment-bot token
       if: always() && steps.has_permissions.outputs.result == 'true'
       id: get_comment_bot_token
-      uses: peter-murray/workflow-application-token-action@e8782d687a306fb13d733244d0f2a50e272d3752
+      uses: peter-murray/workflow-application-token-action@6a9741bf210c9a93370560585c908161170ecada
       with:
         application_id: ${{ secrets.application-id }}
         application_private_key: ${{ secrets.application-private-key }}


matt-graham:
Have tested changes on repository at https://github.com/matt-graham/pr-comment-workflows and correct commit seems to now be getting checked out on both issue and pull request comments.

giordano (approved):

Issue #1125, opened 2023-09-21
Warnings from `Hiv` module in the routine runs


tbhallett:
```
This event should not be running, Hiv_DecisionToContinueTreatment is for those already on tx
```

We are seeing this warning come up quite frequently when running the [standard runs on the development server](https://github.com/UCL/TLOmodel/blob/master/src/scripts/calibration_analyses/scenarios/long_run_all_diseases.py).

It seems like something is happening that we do not anticipate.

tamuri:
To check logs on dev server

PR #1129, opened 2023-09-22
Prevent people having `Hiv_DecisionToContinueTreatment` events at inappropriate times

tdm32:
This PR addresses issue #1125 

The issue: people were having Hiv_DecisionToContinueTreatment at inappropriate times, i.e. when they were not on ART.

The issues arises in certain instances whereby people are routed into HIV testing (and subsequent treatment) through multiple routes and if drugs are not available, they are rescheduled for another treatment appt.

Example:

- person A had HIV test and is referred to treatment
- drugs are not available so person A is scheduled a further treatment appt in one month
- separately, person A has another test scheduled through hsi_generic_first_appts
- this test refers them again to treatment, which is successfully given
- Hiv_DecisionToContinueTreatment is scheduled in 3 months
- One month later, the first rescheduled treatment appt runs
- This event cannot determine if person A is due for their 3 monthly prescription, but checks only whether they are a new initiator or continuing - both qualify for ARV dispensation
- This treatment is now also given and Hiv_DecisionToContinueTreatment is scheduled in 3 months
- In this way, there are two separate time-series of treatment/follow-up appts occurring for person A

To resolve this, I have added a property hv_date_last_ART which will track the date of the last ART dispensation. The parameter dispensation_period_months guides how frequently the person should return for follow-up ART appts. 

In HSI_Hiv_StartOrContinueTreatment, an additional condition now checks whether the person had their last successful ART appt at least 3 months ago. If person is on ART but their hv_date_last_ART is less than 3 months ago, the appt will not run. 

Parameter dispensation_period_months can be modified to allow to longer prescription lengths in future. 

@marghe-molaro @tbhallett please let me know if these changes fix the warnings arising in the dev runs.


tbhallett:
Thanks @tdm32
I've commissioned a long run to confirm the warnings are not issued in this case: `long_run_all_diseases-2023-09-22T141433Z`

src/tlo/methods/hiv.py
```diff
@@ -746,7 +751,8 @@ def split_into_vl_and_notvl(all_idx, prob):
         # all those on ART need to have event scheduled for continuation/cessation of treatment
         # this window is 1-90 days (3-monthly prescribing)
         for person in art_idx:
-            days = self.rng.randint(low=1, high=90, dtype=np.int64)
+
+            days = self.rng.randint(low=1, high=self.parameters['dispensation_period_months']*30.5, dtype=np.int64)
```
tbhallett, line 755: @tdm32  -- do we want this? 
I would have thought a refill interval of 1 day doesn't make sense.
Drawing from a uniform with this specification, would give a mean of 45 days.

I think we should use this instead, let me know what you think

```
            days = self.rng.randint(low=1, high=self.parameters['dispensation_period_months']*30.5, dtype=np.int64)
            self.sim.schedule_event(
                Hiv_DecisionToContinueTreatment(person_id=person, module=self),
                self.sim.date + DateOffset(months=self.parameters['dispensation_period_months'], days=self.rng.randint(-14,14),
            )
```

tdm32: My thinking was that at baseline, we don't know how long a person has been on ART. They may have had their last dispensation anywhere between 1-90 days ago so I use a uniform distribution to equally distribute Hiv_DecisionToContinueTreatment across the next three months.

I think it probably won't make a big difference - as long as we don't clump them within a few days of each other then I think that's fine.

tbhallett: Ah yes, you're quite right. I was looking at just the code snippet and didn't notice that it was in the initiation section.

tbhallett (approved):

tbhallett:
Noting here that we do see a change in calibration with these changes:

OLD:
image


NEW:
image



@tdm32  -- is this expected? Do the more detailed calibrations plots you have still seem good with these changes?

tbhallett:
.... But, also, a big reduction in Appoitments type 'EstAdult'


image

tdm32:
> Noting here that we do see a change in calibration with these changes
> 
> (perhaps marginally better, actually (?))
> 
> OLD: image
> 
> NEW: image
> 
> @tdm32 -- is this expected? Do the more detailed calibrations plots you have still seem good with these changes?

@tbhallett thanks for these - I would like to do a run with this latest version of the code and I'll produce the full suite of HIV-related figures to check all is fine. I'll try to get this done before the Tues meeting if that's ok?

src/tlo/methods/hiv.py
```diff
@@ -2434,6 +2439,12 @@ def apply(self, person_id, squeeze_factor):
         # Confirm that the person is diagnosed (this should not run if they are not)
         assert person["hv_diagnosed"]
 
+        # check whether person had Rx at least 3 months ago and is now due repeat prescription
+        # alternate routes into testing/tx may mean person already has recent ARV dispensation
+        if person['hv_date_last_ART'] >= (
+                self.sim.date - pd.DateOffset(months=self.module.parameters['dispensation_period_months'])):
+            return
```
tbhallett, line 2446: @tdm32-- I've also just spotted that we don't have an EXPECTED_APPT_FOOTPRINT on this HSI (!?!?!?!).
Is that what we want?
We should probably be explicit about it, if so.

tdm32: I think it should be set in line 2665:
    @property
    def EXPECTED_APPT_FOOTPRINT(self):

tbhallett: Phew!! My little check didn't think about `properties`. :-)

tbhallett: But, shouldn't the `return` to appointment appointment, still be `return <<blank_footprint>>`

tdm32: yes you're right, we should have a return blank footprint statement there if the condition returns true and the appt is aborted. I'll do it now

tdm32:
@tbhallett Updated outputs from this latest version - full model run 2010-2020, pop=5000

There's a bit of noise due to small population size but I think the outputs look fine. I don't think a re-calibration is necessary. Happy to hear your thoughts.

image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

tbhallett:
TBD re calibratoion to 'EstAdults' appointment  type

PR #1130, opened 2023-09-22
Move CI jobs to separate workflows + add paths filters

### LINKED ISSUE #1122: Add filters to `pull_request` trigger for test workflow to avoid unnecessary runs
It would probably be worth using the `path` / `paths-ignore` optional keys on the `pull_request` trigger in

https://github.com/UCL/TLOmodel/blob/b9e157f9562c2fdcd0115bb01c076d8a759d432c/.github/workflows/ci.yml#L1-L9

to only run tests when actually editing files that will lead to some change in test behaviour. In particular might be worth specifically excluding some of the workflow files (e.g. `.github/workflows/run-on-comment.yml`, `.github/workflows/run-profiling.yml`, `.github/workflows/comment-triggered-ci.yml`) and some metadata files like `CITATION.cff`, `MANIFEST.in`, `LICENSE.txt`. If we separated out the documentation job to a different workflow we could also exclude changes to `.rst` files / files in `docs` folder.

matt-graham:
Would fix #1122

Not entirely sure if this is worthwhile, but thought I'd put out there for discussion!

This PR separates the existing `.github/workflows/CI.yml` workflow file into three separate workflows `.github/workflows/tests.yaml`, `.github/workflows/docs.yaml` and `.github/workflows/checks.yaml`, corresponding respectively to the test (find tests + test <file> matrix job), documentation and checks jobs in the existing single workflow. This allows differentiating the `on` triggers for each workflow, with in particular [a `on.pull_request.paths` key](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) added for the tests and documentation workflows to only run the jobs when files are updated in a PR which will lead to changes in the job status. The checks workflow is continued to be run for all pull request triggers without filtering on the basis that one of the checks is checking the manifest against the checked out repository which potentially is affected by changes to any files and also that this job is quick to run so limited gain to be had  from running selectively.

As there is currently a branch protection rule set up on `master` which requests the `Find tests files` step in the current `CI.yml` file to be run before merging is allowed (as well as the checks job) this would need to be updated if this was to be merged, as the tests workflow would by design not run for all PRs.

The PR also switches to using GitHub hosted test runners for all jobs other that the matrix job in which tests are run, on the basis that this is where the self-hosted runners provide most value, and running the other jobs on GitHub runners means they won't be competing with the long running tests. 

I've also changed to only fetch Git-LFS files in the checkout action when running the tests and building documentation, as I believe no LFS tracked files are required for running checks and finding test files though I might be wrong on this?

We could also potentially apply similar `path` filtering rules for the `push` to `master` trigger, but for now I've left this on the assumption that running everything on merging to `master` is safest and happens less frequently than PR pushes so less gain.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,7 @@ on:
     tags: ["*"]
   pull_request:
     paths:
-      - .github/workflows/tests.yaml
+      - .github/workflows/tests.yml
       - requirements/**
       - resources/**
       - src/tlo/**
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,10 +1,18 @@
-name: CI
+name: Tests
 
 on:
   push:
     branches: "master"
     tags: ["*"]
   pull_request:
+    paths:
+      - .github/workflows/tests.yaml
+      - requirements/**
+      - resources/**
+      - src/tlo/**
+      - tests/**
+      - setup.*
+      - tox.ini
   schedule:
     - cron: 0 0 * * *
 
@@ -24,16 +32,14 @@ jobs:
 
   gen-test-matrix:
     name: Find test files
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
-    runs-on: [self-hosted]
+    runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
-    strategy:
-      fail-fast: false
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
-          lfs: true
+          # No tests should be under LFS so no need to fetch LFS files
+          lfs: false
       - id: set-matrix
         name: Set matrix
         run: |
@@ -49,14 +55,13 @@ jobs:
 
   test:
     needs: gen-test-matrix
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
     name: Test ${{ matrix.file }}
     runs-on: [self-hosted]
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.gen-test-matrix.outputs.matrix) }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           lfs: true
       - name: System info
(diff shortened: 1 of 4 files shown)


tamuri:
Looks like a good idea. Going to ask Mose to review this workflow, as he set it up originally.

tamuri:
> I've also changed to only fetch Git-LFS files in the checkout action when running the tests and building documentation, as I believe no LFS tracked files are required for running checks and finding test files though I might be wrong on this?

Looks like there is a test loading parameters from an `.xlsx` file in `tests/resources` (`TestLoadParametersFromDataframe`). We could exclude that directory from LFS.

.github/workflows/tests.yml
```diff
@@ -24,16 +32,14 @@ jobs:
...
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
-    runs-on: [self-hosted]
+    runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
-    strategy:
-      fail-fast: false
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
giordano, line 39: Only tangentially related to this PR, without having to do this manually, you can get [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) to automatically open PRs to update the workflows, see for example https://github.com/JuliaPhysics/Measurements.jl/blob/21cf93c7dc25b7a1b14d5b877bc8f4884599ff5e/.github/dependabot.yml

.github/workflows/tests.yml
```diff
@@ -24,16 +32,14 @@ jobs:
 
   gen-test-matrix:
     name: Find test files
-    if: "!contains(github.event.head_commit.message, '[skip ci]')"
-    runs-on: [self-hosted]
+    runs-on: ubuntu-latest
```
giordano, line 35: Ingenious

giordano (approved):
Looks good to me!

.github/workflows/tests.yml
giordano: > The PR also switches to using GitHub hosted test runners for all jobs other that the matrix job in which tests are run, on the basis that this is where the self-hosted runners provide most value, and running the other jobs on GitHub runners means they won't be competing with the long running tests.

On the other hand these jobs will have to compete with all other repositories in this organisation using the GitHub hosted runners.  I guess we can see how it goes, and revert it back in case there's too much traffic.

tamuri: Is there somewhere to see the queued jobs for the org?

giordano: https://github.com/organizations/UCL/settings/actions/hosted-runners.  Unfortunately it doesn't look like there's an org-level view of past jobs, to get an idea of usual traffic.

matt-graham: Yeah agree that would be good to keep an eye on this as there isn't a strong gain to running these on shared GitHub hosted runners rather than self-hosted.

matt-graham:
> > I've also changed to only fetch Git-LFS files in the checkout action when running the tests and building documentation, as I believe no LFS tracked files are required for running checks and finding test files though I might be wrong on this?
> 
> Looks like there is a test loading parameters from an `.xlsx` file in `tests/resources` (`TestLoadParametersFromDataframe`). We could exclude that directory from LFS.

I think this shouldn't matter as the LFS files are still fetched when the tests are actually run, I've just disabled fetching them when generating the matrix for the test jobs (which just needs the list of test files).

tamuri:
> I think this shouldn't matter as the LFS files are still fetched when the tests are actually run, I've just disabled fetching them when generating the matrix for the test jobs (which just needs the list of test files).

Sorry, yes - gotcha. Another motivation to \[eventually\] move `\resources` out of this repo - we don't need to be downloading those at all.

tamuri:
> ~Sorry, yes - gotcha. Another motivation to \[eventually\] move `\resources` out of this repo - we don't need to be downloading those at all.~

Ignore - that's totally wrong.

matt-graham:
I think the hanging 'Checks _Expected — Waiting for status to be reported_' status check is as the previous check job required in the branch protection rule is now in a different file and so can't be found. If / when we're ready to merge this I think we'll need to change the rule to point to the new workflow job and we probably need to remove the `Find test files` status check (as it won't always run).

tamuri (approved):

Issue #1141, opened 2023-09-28
Need to replace tclose logic in maternal and newborn health modules


joehcollins:
Logic related to tclose dates for key HSIs within the maternal and newborn health modules needs to be adapted to allow for future work related to prioritisation. Currently overriding tclose dates set by the modules leads to errors within these modules relating to the order in which events occur to mothers during pregnancy.

tamuri:
Done

PR #1140, opened 2023-09-28
Roll back PR #1027

tbhallett:

Issue #1102, opened 2023-09-07
Issue with enforcing two-day limit on delay of care in ANC modules


marghe-molaro:
When running the simulation in mode_appt_constraints = 2 and allowing women to persist in seeking care (such that they may receive care over two days later than topen) the following issue presents with some of the women receiving a generic_first_appt_emergency:

  File "/TLOmodel/src/tlo/methods/healthsystem.py", line 2615, in apply
    self.process_events_mode_2(hold_over)
  File "/TLOmodel/src/tlo/methods/healthsystem.py", line 2432, in process_events_mode_2
    actual_appt_footprint = event.run(squeeze_factor=squeeze_factor)
  File "/TLOmodel/src/tlo/methods/healthsystem.py", line 224, in run
    updated_appt_footprint = self.apply(self.target, squeeze_factor)
  File "/TLOmodel/src/tlo/methods/hsi_generic_first_appts.py", line 97, in apply
    do_at_generic_first_appt_emergency(hsi_event=self, squeeze_factor=squeeze_factor)
  File "/TLOmodel/src/tlo/methods/hsi_generic_first_appts.py", line 313, in do_at_generic_first_appt_emergency
    mni[person_id]['sought_care_for_complication'] &
KeyError: 'sought_care_for_complication'

In a year-long test, this error occurred 9 times. It is most probably linked to PR #1027. 

This error does not present itself if not allowing women to continue to seek care after they have been turned away (i.e. if tclose = topen + 1 day).

marghe-molaro:
Over a 5.5 years period (and as queues get longer), this error ramps up to an average of 76 of such errors a year

marghe-molaro:
Hi @joehcollins @tbhallett,

Another issue has arisen which seems to be linked to the new "inpatient_care_doesnt_run" fnc (see error below). This occurred during a tclose = 1day run. 

 File "/TLOmodel/src/tlo/methods/healthsystem.py", line 2334, in process_events_mode_2
    self.module.call_and_record_never_ran_hsi_event(
  File "/TLOmodel/src/tlo/methods/healthsystem.py", line 1780, in call_and_record_never_ran_hsi_event
    hsi_event.never_ran()
  File "/TLOmodel/src/tlo/methods/care_of_women_during_pregnancy.py", line 2380, in never_ran
    self.module.inpatient_care_doesnt_run(self.target)
[...]

PR #1143, opened 2023-09-28
Don't overwrite tclose parameter for ANC-related HSIs

marghe-molaro:
This PR builds on PR #1140 to solve address  #1102 by ensuring that the tcloses of HSIs in modules CareOfWomenDuringPregnancyr, Labour, PostnatalSupervisor, or NewbornOutcomes are not overwritten when the tclose_overwrite variable is set to True. This is necessary because tclose in these modules does not only encode assumptions about the health-seeking behaviour of patients, as discussed in issue #1141. 

We have submitted run  (Job ID: effect_of_each_treatment-2023-09-28T104657Z) to check errors appearing in runs described in #1102 have been resolved.


src/tlo/methods/healthsystem.py
```diff
@@ -1277,10 +1277,16 @@ def schedule_hsi_event(
         """
         # If there is no specified tclose time then set this to a week after topen.
         # This should be a boolean, not int! Still struggling to get a boolean variable from resource file
-        if self.tclose_overwrite == 1:
-            tclose = topen + pd.to_timedelta(self.tclose_days_offset_overwrite, unit='D')
-        elif tclose is None:
-            tclose = topen + DateOffset(days=7)
+
+        # Clinical time-constraints are embedded in tclose for these modules, do not overwrite their tclose
+        if hsi_event.module == ('PregnancySupervisor' or 'Labour' or 'PostnatalSupervisor' or 'NewbornOutcomes'):
```
joehcollins, line 1282: do we not want `hsi_event.module == ('CareOfWomenDuringLabour' or 'Labour' or 'PostnatalSupervisor' or 'NewbornOutcomes')` instead?

joehcollins: *CareOfWomenDuringPregnancy

marghe-molaro: Yes sorry, have fixed now.

joehcollins (approved):
Looks all good now! Wondering if going forward we need to write a formal test file that simulates different policies with the full model are then checks the log for any error messages brought up across modules?

marghe-molaro:
> Looks all good now! Wondering if going forward we need to write a formal test file that simulates different policies with the full model are then checks the log for any error messages brought up across modules?

Hi @joehcollins, this is an interesting idea; I think this will be (mostly) addressed when we start running routine mode_appt_constraints = 2 long runs from master on the dev server alongside the standard mode_appt_constraints = 1 ones, as these runs should (hopefully) pick up and flag errors along the way. I fear that because errors arising from prioritisation are often only visible when simulating large populations for a while, doing this in tests would become too cumbersome for developers otherwise. But it might be good to advise developers to include a test in mode_appt_constraints = 2 for their modules, as these should likely pick up any obvious problems. 

Wrt which priority policy to consider when performing these dev runs I'm not too sure at this point; I would have naively gone for Naive (of course) but maybe we could consider a couple of the policies. I'll have a think! Thanks again

joehcollins:
Ahhh you're very right r.e. population size- didn't think of that. And yes certainly will capture things when these long runs are happening on the dev server. The errors i'm thinking of are those which wont crash the code and are instead captured in the log files. But maybe checking for this can be baked into the simulations on the dev server (unsure how that all works!!).

Thanks again for all the work on this - sorry again that my modules are such a pain!

src/tlo/methods/healthsystem.py
```diff
@@ -1282,10 +1282,16 @@ def schedule_hsi_event(
         """
         # If there is no specified tclose time then set this to a week after topen.
         # This should be a boolean, not int! Still struggling to get a boolean variable from resource file
-        if self.tclose_overwrite == 1:
-            tclose = topen + pd.to_timedelta(self.tclose_days_offset_overwrite, unit='D')
-        elif tclose is None:
-            tclose = topen + DateOffset(days=7)
+
+        # Clinical time-constraints are embedded in tclose for these modules, do not overwrite their tclose
+        if hsi_event.module == ('CareOfWomenDuringPregnancy' or 'Labour' or 'PostnatalSupervisor' or 'NewbornOutcomes'):
```
tbhallett, line 1287: In this current version, using `or` like that won't give the (I presume) intended behaviour

```suggestion
        if hsi_event.module in ('CareOfWomenDuringPregnancy', 'Labour', 'PostnatalSupervisor' ,'NewbornOutcomes'):
```

Also --- should it be `hsi_event.module.name`???  We're making a comparison to a string for the module's name rather than the module itself.

src/tlo/methods/healthsystem.py
```diff
@@ -1282,10 +1282,16 @@ def schedule_hsi_event(
...
+            if tclose is None:
+                tclose = topen + DateOffset(days=7)
```
tbhallett, line 1289: here, and in line 1294, we have the defaulting to  7 days for `tclose`. I guess it happens somewhere else too. I think we should factorize this so that we have one declaration of the default value for an unspecified `tclose`.

tbhallett:
@marghe-molaro  -- there's a lot of deletion in `src/scripts/healthsystem/impact_of_policy/scenario_impact_of_policy.py`. 
Do you want that? or Should these changes not come into this PR. (I can roll them back if you want)

marghe-molaro:
@tbhallett I think the changes to the scenario file are fine, since they remove all those "idealised" scenarios that we didn't want to consider anyway.

tbhallett (approved):

PR #1149, opened 2023-10-03
Some minor improvement to the Ansible playbook for GHA runners

giordano:
Main changes:

* Parametrise the python version to be installed
* Install also the `python3.x-venv` package

As discussed today offline with Asif and Matt.  Opening as draft in case I have other minor changes to add in the next few days.


matt-graham (approved):
Thanks - all looks good to me.

PR #1150, opened 2023-10-06
Fix to `summarize`

tbhallett:
I noticed that this function previously relied on the order of the index from a `pd.groupby` using index groupers being the same as in the original data frame's index. This used to be fine, but with the change to the new version of pandas causes weird results. I could have made the fix just by adding `sort=False` as an argument to `groupby` .... but I thought it would be safer not rely on ordering and instead let the matching-up be done by `pd.concat`, using the actual indices.

This a **high priority fix** as, without it, results from `summarise` may become jumbled up between the draws, with no warning/errors raised.

### DIFF
--- a/src/tlo/analysis/utils.py
+++ b/src/tlo/analysis/utils.py
@@ -327,6 +327,7 @@ def summarize(results: pd.DataFrame, only_mean: bool = False, collapse_columns: 
         # Remove other metrics and simplify if 'only_mean' across runs for each draw is required:
         om: pd.DataFrame = summary.loc[:, (slice(None), "mean")]
         om.columns = [c[0] for c in om.columns.to_flat_index()]
+        om.columns.name = 'draw'
         return om
 
     elif collapse_columns and (len(summary.columns.levels[0]) == 1):
--- a/src/tlo/analysis/utils.py
+++ b/src/tlo/analysis/utils.py
@@ -319,8 +319,9 @@ def summarize(results: pd.DataFrame, only_mean: bool = False, collapse_columns: 
         },
         axis=1
     )
-    summary.columns = summary.columns.swaplevel(1, 0).sort_index(axis=1)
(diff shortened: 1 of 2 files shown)


src/tlo/analysis/utils.py
```diff
@@ -311,19 +311,15 @@ def summarize(results: pd.DataFrame, only_mean: bool = False, collapse_columns:
...
-                results.columns.unique(level='draw'),
-                ["mean", "lower", "upper"]
-            ],
-            names=['draw', 'stat']),
-        index=results.index
+    summary = pd.concat({
+        'mean': results.groupby(axis=1, by='draw', sort=False).mean(),
+        'lower': results.groupby(axis=1, by='draw', sort=False).quantile(0.025),
+        'upper': results.groupby(axis=1, by='draw', sort=False).quantile(0.975),
+    },
```
matt-graham, line 318:
```suggestion
    summary = pd.concat(
        {
            'mean': results.groupby(axis=1, by='draw', sort=False).mean(),
            'lower': results.groupby(axis=1, by='draw', sort=False).quantile(0.025),
            'upper': results.groupby(axis=1, by='draw', sort=False).quantile(0.975),
        },
```
Matching indentation of opening and closing delimiters makes it a bit easier to see the grouping of the items here.

matt-graham (changes requested):
Good catch - this looks good to me, but would be useful to add a test for the `summarize` function in `tests/test_analysis.py` so that we catch any other breaking changes like this. Possibly just summarizing a hardcoded very small dataframe of the expected format and checking the output aligns with what was expected.

matt-graham (approved):
Thanks @tbhallett for updating and adding test - this looks good to merge to me now once Actions jobs finish running. Confirmed that new test passes locally (and fails prior to other changes in this PR).

PR #1153, opened 2023-10-09
Fix in handling of `Causes`

tbhallett:
It was noted that `merge_dicts_of_causes` over-writing the module's delcaration of the causes. (This gave rise to an error when changing defintion of CAUSES_OF_DEATHS and CAUSES_OF_DISABILITY that was hard to catch and it only occured on the "second" run of the model, after which the declaration had been written-to.)

The main fix is that the `merge_dicts_of_causes` now uses `copy` to prevent mutating the `dicts` that are passed in.

A secondary ("belt and braces") fix is in the `collect_causes_from_disease_modules` helper function, which also uses `copy` to prevent there ever being a link created from the module declaration of CAUSES_OF_DEATHS and CAUSES_OF_DISABILITY to anything used in the simulation.

### DIFF
--- a/src/tlo/methods/causes.py
+++ b/src/tlo/methods/causes.py
@@ -3,6 +3,7 @@ Classes and functions that support declarations of causes of death and disabilit
 """
 
 from collections import defaultdict
+from copy import copy
 from typing import Union
 
 import pandas as pd
@@ -67,8 +68,9 @@ def collect_causes_from_disease_modules(all_modules, collect, acceptable_causes:
                         f"Conflict in declared cause {tlo_cause} by {m.name}. " \
                         f"A different specification has already been registered."
 
-                # If ok, update these causes to the master dict of all causes of death
-                collected_causes.update({tlo_cause: cause})
+                # If ok, update these causes to the master dict of all causes of death  (copies to prevent any
+                # accidental editing of the modules' declarations).
+                collected_causes.update({copy(tlo_cause): copy(cause)})
 
     # Check that each gbd_cause is not defined in respect of more than one label
     gbd_causes = dict()  # dict(<gbd_cause: label>)
--- a/src/tlo/methods/healthburden.py
+++ b/src/tlo/methods/healthburden.py
@@ -2,6 +2,7 @@
 This Module runs the counting of Life-years Lost, Life-years Lived with Disability,
 and Disability-Adjusted Life-years (DALYS).
 """
+from copy import copy
 from pathlib import Path
 from typing import Dict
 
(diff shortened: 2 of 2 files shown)


src/tlo/methods/causes.py
```diff
@@ -67,8 +68,9 @@ def check_cause(_cause: Cause, _acceptable_causes: set):
                         f"Conflict in declared cause {tlo_cause} by {m.name}. " \
                         f"A different specification has already been registered."
 
-                # If ok, update these causes to the master dict of all causes of death
-                collected_causes.update({tlo_cause: cause})
+                # If ok, update these causes to the master dict of all causes of death  (copies to prevent any
+                # accidental editing of the modules' declarations).
+                collected_causes.update({copy(tlo_cause): copy(cause)})
```
matt-graham, line 73: What data types are `tlo_cause` and `cause` here? Most built-in mutable container types (`list`, `dict`, `set`) have a `copy` method that would avoid the need to import `copy` function. If `tlo_cause` is an immutable type (for example `str` or `tuple`), which is likely as generally `dict` keys need to be immutable to be hashable, then no need to create a copy as underlying container cannot be updated after creation. 

Also tangential to changes here, for adding a single entry to a dictionary something like `collected_causes[tlo_cause] = cause.copy()` would I think be both clearer and more performant than using `update` method (which is more meant for merging in a dictionary with multiple items).

src/tlo/methods/healthburden.py
```diff
@@ -179,7 +180,8 @@ def merge_dicts_of_causes(d1: Dict, d2: Dict) -> Dict:
                 for _tlo_cause_name, _cause in d.items():
                     if _cause.label not in labels_seen:
                         # If label is not already included, add this cause to the merged dict
-                        merged_causes[_tlo_cause_name] = _cause
+                        merged_causes[_tlo_cause_name] = copy(_cause)  # Use copy to avoid the merged dict being linked
```
matt-graham, line 183: Similar to above comment, depending on type of `_cause` it might have a `copy` method which would be slightly more performant than calling `copy` function (which will just call this method itself).

matt-graham (approved):
Nice detective work tracking this down 🕵️! I've added a couple of comments / suggested changes, but these are more nice to haves, not essential.

tbhallett:
> Nice detective work tracking this down 🕵️! I've added a couple of comments / suggested changes, but these are more nice to haves, not essential.

Thanks Matt. The "cause" is a custom type I've written, so I guess I would have to write its copy method. Is there a standard format for doing that?

src/tlo/methods/causes.py
tbhallett: I can make this:
```python
collected_causes[tlo_cause] = copy(cause)
```

`tlo_cause` is a string, but `cause` is a custom type I've written

matt-graham:
> Thanks Matt. The "cause" is a custom type I've written, so I guess I would have to write its copy method. Is there a standard format for doing that?

Ah my bad - if it's a custom type, using `copy` from `copy` module as you are doing already is probably the easiest solution. You can define special `__copy__` / `__deepcopy__` methods to hook in to how `copy` deals with copying custom objects, but the default behaviour should work fine here.

src/tlo/methods/causes.py
matt-graham: Yep if `tlo_cause` is a string then no need to copy (as it's immutable so cannot be modified in place) and using `copy` function on `cause` makes sense if a custom type.

tbhallett:
Thanks for all this @matt-graham.

I've made the change. Shall I merge it in now?

matt-graham:
Good to merge from my perspective but I'd probably wait for tests just in case my suggested change break things 😅.

Issue #629, opened 2022-06-15
Inconsistent Behaviour in use of `custom_log`


tbhallett:
When using the argument `custom_log` when initialising a simulation, behaviour is not as expected when there is more than logger defined within a module. The behaviour is as expected when using the method `sim.configure_logging()` after the models have been registered.

As a consequence, all runs that ask for `tlo.methods.demography` (which is most) also go `tlo.methods.demography.detail` (which they don't want and is very heavy).

The issue is demonstrated in the tests added in https://github.com/UCL/TLOmodel/pull/628

* This test passes
```python
def test_can_turn_off_the_detailed_logger_when_using_custom_log_after_registering(seed, tmpdir):
    """Check that the simulation can be set-up to get only the usual demography logger and not the detailed logger,
    when the custom_log information is given after the models are registered."""

    log_config = {
        'filename': 'temp',
        'directory': tmpdir,
        'custom_levels': {
            "*": logging.WARNING,
            'tlo.methods.demography.detail': logging.WARNING,  # <-- Explicitly turning off the detailed logger
            'tlo.methods.demography': logging.INFO,  # <-- Turning on the normal logger
        }
    }

    rfp = Path(os.path.dirname(__file__)) / '../resources'

    sim = Simulation(start_date=Date(2010, 1, 1), seed=seed)
    sim.register(demography.Demography(resourcefilepath=rfp))
    sim.configure_logging(**log_config)
    sim.make_initial_population(n=100)
    sim.simulate(end_date=sim.start_date)

    # Cause one death to occur
    sim.modules['Demography'].do_death(
        individual_id=0,
        originating_module=sim.modules['Demography'],
        cause='Other'
    )

    log = parse_log_file(sim.log_filepath)

    # Check the usual `tlo.methods.demography' log is created and that check persons have died (which would be when the
    # detailed logger would be used).
    assert 'tlo.methods.demography' in log.keys()
    assert 1 == len(log['tlo.methods.demography']['death'])

    # Check that the detailed logger is not created.
    assert 'tlo.methods.demography.detail' not in log.keys()
```

* BUT This test fails
```python
def test_can_turn_off_the_detailed_logger_when_using_custom_log_when_registering(seed, tmpdir):
    """Check that the simulation can be set-up to get only the usual demography logger and not the detailed logger,
    when providing the config_log information when the simulation is initialised."""

    log_config = {
        'filename': 'temp',
        'directory': tmpdir,
        'custom_levels': {
[...]

PR #1157, opened 2023-10-10
`custom_log` not behaving as expected

tbhallett:
This is a demonstration of an issue that think isn't fixed but has been closed: https://github.com/UCL/TLOmodel/issues/629
It may be that I am not understanding the expected behaviour.


tamuri:
You're right the logic is not right (or, at least, not intuitive). When applying the log configuration, we match the logger name to its name exactly. If there is no match, it chops off the last part of the logger and checks again. It keeps doing this until it reaches the `tlo.methods` logger which is the `*` level warning, in your example. So, `tlo.methods.demography.detail` doesn't match, but matches the supplied `tlo.methods.demography` and takes the same level. The idea was that `"tlo.methods.demography": logging.WARNING` would set _all_ the demography loggers to warning.

It sounds like it's best to explicitly set logging levels as configured. We would lose the functionality above, but perhaps that was not very useful anyway because you're unlikely to turn off (i.e. set to warning) all the loggers for one particular module. Rather, it's more useful to turn off all loggers, then explicitly set the ones that are different.

(Thanks for the test - perfect way to demonstrate the problem!)

tbhallett:
> You're right the logic is not right (or, at least, not intuitive). When applying the log configuration, we match the logger name to its name exactly. If there is no match, it chops off the last part of the logger and checks again. It keeps doing this until it reaches the `tlo.methods` logger which is the `*` level warning, in your example. So, `tlo.methods.demography.detail` doesn't match, but matches the supplied `tlo.methods.demography` and takes the same level. The idea was that `"tlo.methods.demography": logging.WARNING` would set _all_ the demography loggers to warning.
> 
> It sounds like it's best to explicitly set logging levels as configured. We would lose the functionality above, but perhaps that was not very useful anyway because you're unlikely to turn off (i.e. set to warning) all the loggers for one particular module. Rather, it's more useful to turn off all loggers, then explicitly set the ones that are different.
> 
> (Thanks for the test - perfect way to demonstrate the problem!)

Thansks; and yes, I'd agree it'd be most intuitive to set each and every logger individually (apart from the wildcard for _every_ logger).

mnjowe:
@tbhallett and @tamuri . I will be pushing some commits here to complete this PR and close issue #629. The following are the files that will be affected by implementing suggestions made in issue #629 

- [ ]  test_analysis

- [ ] analysis_tb

- [ ] schisto_analysis

- [ ] breast_cancer_analyses

- [ ] oesophagealcancer_analyses

- [ ] other_adult_cancers_analyses

- [ ] prostate_cancer_analyses

mnjowe:
@tbhallett . I can see that in this branch you are not explicitly turning off [demography.detail logger](https://github.com/UCL/TLOmodel/blob/f325a3ad5daf9c0c93df0581e6a58f9344a547c7/tests/test_analysis.py#L532) BUT you're in your 2 examples in issue #629. What's your final decision on this. Should we turn off demography.detail explicitly when configuring custom levels or not?

tbhallett:
> @tbhallett . I can see that in this branch you are not explicitly turning off [demography.detail logger](https://github.com/UCL/TLOmodel/blob/f325a3ad5daf9c0c93df0581e6a58f9344a547c7/tests/test_analysis.py#L532) BUT you're in your 2 examples in issue #629. What's your final decision on this. Should we turn off demography.detail explicitly when configuring custom levels or not?

My view is that it shouldn't matter whether or not a logger is implicitly switched (using '*') or explicitly switched (using its exact name). The only thing that should matter is the order --- subsequent instructions take precedence to earlier ones.

mnjowe:
okay good!

mnjowe:
Just observed that test analysis in this PR is failing for not explicitly turning off detailed logger in log configuration. I will [uncomment this line](https://github.com/UCL/TLOmodel/blob/f325a3ad5daf9c0c93df0581e6a58f9344a547c7/tests/test_analysis.py#L532) if its okay by you, restrict setting up logs at init and apply changes to all other affected files.

tamuri (approved):

PR #1159, opened 2023-10-11
Fix calibration scripts for Pandas v2

giordano:
I'm trying to run the calibration with pandas v2, but I'm facing a string of errors, trying to fix them one by one.  Opening as a draft for the time being, as I'm not done yet with all the fixes (I'm also not entirely sure they're all correct).

Do I take correctly these scripts aren't currently tested?


src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
```diff
@@ -87,7 +87,7 @@ def apply(results_folder: Path, output_folder: Path, resourcefilepath: Path = No
     # Load Data: WPP_Annual
     wpp_ann = pd.read_csv(Path(resourcefilepath) / "demography" / "ResourceFile_Pop_Annual_WPP.csv")
     wpp_ann['Age_Grp'] = wpp_ann['Age_Grp'].astype(make_age_grp_types())
-    wpp_ann_total = wpp_ann.groupby(['Year']).sum().sum(axis=1)
+    wpp_ann_total = wpp_ann.groupby(['Year'])['Count'].sum()
```
giordano, line 90: `wpp_ann.groupby(['Year'])['Count'].sum()` isn't exactly the same as the old `wpp_ann.groupby(['Year']).sum().sum(axis=1)`, that'd be `wpp_ann.groupby(['Year'])[['Age', 'Count']].sum().sum(axis=1)`, but it looks suspicious one wants to sum Age _and_ Count columns together?

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
```diff
@@ -102,9 +102,9 @@ def apply(results_folder: Path, output_folder: Path, resourcefilepath: Path = No
             marker='o', markersize=10, linestyle='none', label='Census', zorder=10, color=colors['Census'])
     ax.plot(pop_model.index, pop_model['mean'] / 1e6,
             label='Model (mean)', color=colors['Model'])
-    ax.fill_between((pop_model.index).to_numpy(),
-                    (pop_model['lower'] / 1e6).to_numpy(),
-                    (pop_model['upper'] / 1e6).to_numpy(),
+    ax.fill_between((pop_model.index).to_numpy(dtype=float),
+                    (pop_model['lower'] / 1e6).to_numpy(dtype=float),
+                    (pop_model['upper'] / 1e6).to_numpy(dtype=float),
```
giordano, line 107: This can probably also be fixed by changing how the `pop_model` dataframe is constructed (at the moment the columns are `object`s, and they can't be properly be used inside `ax.fill_between`), but coercing conversion to `float` looked like a quick solution without too much refactoring.

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
matt-graham: Agreed this looks a bit odd - running with Pandas v1.5 we get that `wpp_ann.groupby(['Year']).sum()` outputs

```
        Age       Count
Year
1950  10100   2953871.0
1951  10100   3011789.0
1952  10100   3072386.0
1953  10100   3135689.0
1954  10100   3201722.0
...     ...         ...
2096  10100  65108710.0
2097  10100  65487105.0
2098  10100  65854995.0
2099  10100  66212402.0
2100  10100  66559386.0
```

while `wpp_ann.groupby(['Year']).sum().sum(axis=1)` outputs 

```
Year
1950     2963971.0
1951     3021889.0
1952     3082486.0
1953     3145789.0
1954     3211822.0
           ...
2096    65118810.0
2097    65497205.0
2098    65865095.0
2099    66222502.0
2100    66569486.0
Length: 151, dtype: float64
```

which as @giordano says seems to be summing up `Age` and `Count` columns.  @tbhallett is this intentional?

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
```diff
@@ -294,7 +294,7 @@ def get_mean_pop_by_age_for_sex_and_year(sex, year):
             fig.show()
             plt.close(fig)
 
-        except pd.core.base.DataError:
+        except pd.errors.DataError:
```
giordano, line 297: I also need help to understand the intent of this `try`-`except` block: was it to catch cases where the dataframe was empty, because `year` was after the simulation's end?  Because now I think we're getting an empty dataframe in that case, instead of a `DataError` exception being raised, so that this entire `try`-`except` block might be completely useless, and things should be handled differently

tbhallett:
> I'm trying to run the calibration with pandas v2, but I'm facing a string of errors, trying to fix them one by one. Opening as a draft for the time being, as I'm not done yet with all the fixes (I'm also not entirely sure they're all correct).
> 
> Do I take correctly these scripts aren't currently tested?

Thanks @giordano 

These scripts are used really heavily in analyses and the dev server. They do work for me still (without error) and seem also to have no problems on the dev server runs. I'm just run them all again (with the pandas update) and no problems. Not sure what I'm missing..!

giordano:
What version of Pandas are you using?  They work for me with Pandas v1.2.2, but not v2.0.3, which is now in https://github.com/UCL/TLOmodel/blob/72aaf8707bb4a798ea3728627d3eb2d281e4671c/requirements/dev.txt#L136

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
tbhallett: This is good spot. The old versions look like an error and the new version seems correct to me!

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
tbhallett: That's right --- the simulations that this script is run for may be of different lengths, and so may or may not include some of  the years `[2010, 2015, 2018, 2029, 2049]`, for which this routine is called.

I suppose a better check would be of the form:

`if year in wpp_ann['Year']: `

tbhallett:
> What version of Pandas are you using? They work for me with Pandas v1.2.2, but not v2.0.3, which is now in
> 
> https://github.com/UCL/TLOmodel/blob/72aaf8707bb4a798ea3728627d3eb2d281e4671c/requirements/dev.txt#L136

image

tbhallett:
> What version of Pandas are you using? They work for me with Pandas v1.2.2, but not v2.0.3, which is now in
> 
> https://github.com/UCL/TLOmodel/blob/72aaf8707bb4a798ea3728627d3eb2d281e4671c/requirements/dev.txt#L136

hang-on. Update coming...... I've found my error

tbhallett:
> What version of Pandas are you using? They work for me with Pandas v1.2.2, but not v2.0.3, which is now in
> 
> https://github.com/UCL/TLOmodel/blob/72aaf8707bb4a798ea3728627d3eb2d281e4671c/requirements/dev.txt#L136

I;m on that version, and do indeed reproduce the error (and the dev server will too when I gets to the change in version, I presume)

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
EvaJanouskova: This actually started working to me without changes in the code, once I merged the Fix to summarize [#1150].

tbhallett:
@giordano and @matt-graham -- is this script running well on your end now? Am I safe to merge it in and use it on my branches?

giordano:
No, with this PR I only addressed the first few errors I was facing, but then I got many more at which point I gave up.

I updated this PR to replace the `try`/`except` with the `if` suggested above, and the `ax.fill_between` call which reportedly started working with a different PR.  I think this should be ok to merge as a starting point, but more work needs to be done to actually update all scripts.

tbhallett:
> No, with this PR I only addressed the first few errors I was facing, but then I got many more at which point I gave up.
> 
> I updated this PR to replace the `try`/`except` with the `if` suggested above, and the `ax.fill_between` call which reportedly started working with a different PR. I think this should be ok to merge as a starting point, but more work needs to be done to actually update all scripts.

Ok, thanks.  I'll get it working.

src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
tbhallett: This also worked for me.

tbhallett:
> @tbhallett - thanks this looks good. I've verified that with your changes the `src/scripts/calibration_analyses/analysis_scripts/analysis_all_calibration.py` script runs to completion using the last full set of `long_run_all_disease` scenario results from the dev server (`2023-09-20_090745_b9e157f95_021`) in an environment with pandas v2.
> 
> I've made some suggested minor changes around removing `plt.show()` calls which I think should be unnecessary if plot is being saved and closed immediately, and some other small changes. I think there are also some other `plt.show` instances in the scripts which are touched by changes here so I can't directly make suggestions on. I would say if we're assuming this script is for producing save figure files rather than interactively displaying, it would be best to remove all `plt.show` instances to simplify running on machines without an X server for displaying plots.

Thanks so much for the close review and the good tips, @matt-graham 


I see what you mean about the use of `plt.show()` and `fig.show()`. I'll remove them now.

tbhallett:
@matt-graham  - when you run this, do you also get copious warnings of:
```
/Users/tbh03/GitHub/TLOmodel/src/tlo/population.py:67: PerformanceWarning: DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
  props[prop_name] = prop.create_series(prop_name, size)
```

matt-graham (approved):

PR #1168, opened 2023-10-25
Ensure clean transition between mode 1 and mode 2

marghe-molaro:
Currently, policies can exclude certain HSIs by setting their priority below (i.e. higher than) the "lowest_priority_considered". At time of scheduling the event, if the priority is higher than the lowest_priority_considered the HSI is not added to the queue, but rather  schedule_to_call_never_ran_on_date is scheduled on date tclose. 

Because the "lowest_priority_considered" check happens only at time of scheduling and not at time of queue querying, when transitioning from mode 1 to mode 2 HSIs which were scheduled in mode 1 but may no longer be allowed to run under the policy in mode 2 are still running. We therefore add an additional priority check at time of querying. 

A lingering issue is that the priority assumed when scheduling HSIs during the "mode 1" phase is policy-dependent, and hence may be above the "lowest_priority_considered" assumed by the policy we want to actually consider in mode 2. Therefore, if we wanted to ensure a "clean" transition we should always run with the same policy, even in mode 1 [but ensuring that the lowest_priority_considered check at time of scheduling only occurs in mode 2], and never allow for a policy transition, but only a mode transition. This PR ensures this by additionally removing the policy switch completely, and making the "lowest_priority_considered" check at time of scheduling only occur in mode_appt_constraints=2.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -2487,13 +2487,13 @@ class HealthSystemScheduler(RegularEvent, PopulationScopeEventMixin):
             else:
                 break
 
-        # Traverse the queue again to check all appts which have expired are removed from the queue,
-        # and call did_not_run() for all those that were postponed.
         # In previous iteration, we stopped querying the queue once capabilities
-        # were exhausted, so here ensure if any events expired were left unchecked they are properly
-        # removed from the queue, and did_not_run() is invoked for all postponed events.
-        # (This should still be more efficient than querying the queue as done in mode_appt_constraints
-        #  = 0 and 1 while ensuring mid-day effects are avoided.)
+        # were exhausted, so here we traverse the queue again to ensure that if any events expired were
+        # left unchecked they are properly removed from the queue, and did_not_run() is invoked for all
+        # postponed events. (This should still be more efficient than querying the queue as done in
+        # mode_appt_constraints = 0 and 1 while ensuring mid-day effects are avoided.)
+        # We also schedule a call_never_run for any HSI below the lowest_priority_considered,
+        # in case any of them where left in the queue due to a transition from mode 0/1 to mode 2
         while len(self.module.HSI_EVENT_QUEUE) > 0:
 
             next_event_tuple = hp.heappop(self.module.HSI_EVENT_QUEUE)
@@ -2501,7 +2501,13 @@ class HealthSystemScheduler(RegularEvent, PopulationScopeEventMixin):
 
             event = next_event_tuple.hsi_event
 
-            if self.sim.date > next_event_tuple.tclose:
+            # If the event is lower than lowest_priority_considered, schedule a call_never_ran on tclose
+            # regardless of whether appt is due today or any other time
+            if next_event_tuple.priority > self.module.lowest_priority_considered:
+                self.module.schedule_to_call_never_ran_on_date(hsi_event=event,
+                                                               tdate=next_event_tuple.tclose)
+
+            elif self.sim.date > next_event_tuple.tclose:
                 # The event has expired (after tclose) having never been run. Call the 'never_ran' function
                 self.module.call_and_record_never_ran_hsi_event(
                       hsi_event=event,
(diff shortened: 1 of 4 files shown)


marghe-molaro:
@tbhallett test for mode 1 under different policies now included.

tests/test_healthsystem.py
```diff
@@ -193,6 +193,58 @@ def test_run_no_interventions_allowed(tmpdir, seed):
...
 
 
+@pytest.mark.slow
+def test_policy_has_no_effect_on_mode1(tmpdir, seed):
+    # Events ran in mode 1 should be identical regardless of policy assumed.
+    # In policy "No Services", have set all HSIs to priority below lowest_priority_considered,
+    # in mode 1 they should all be scheduled and delivered regardless
+
+    output = []
+    for i, policy in enumerate(["Naive", "No Services"]):
```
tbhallett, line 203: why not iterate through all the policies?

tests/test_healthsystem.py
```diff
@@ -193,6 +193,58 @@ def test_run_no_interventions_allowed(tmpdir, seed):
...
+                "filename": "log",
+                "directory": tmpdir,
+                "custom_levels": {
+                    "tlo.methods.healthsystem": logging.DEBUG,
+                }
+            }
+        )
+
+        # Register the core modules
+        sim.register(demography.Demography(resourcefilepath=resourcefilepath),
```
tbhallett, line 218: could use the "full model" to make it the hardest possible test.

tests/test_healthsystem.py
```diff
@@ -193,6 +193,58 @@ def test_run_no_interventions_allowed(tmpdir, seed):
...
+
+        print(type(parse_log_file(sim.log_filepath, level=logging.DEBUG)))
+
+        # read the results
+        output.append(parse_log_file(sim.log_filepath, level=logging.DEBUG))
+
+    # Check that the outputs are the same and that all HSIs ran even with policy specifying "No Services"
+    assert len(output[0]['tlo.methods.healthsystem']['HSI_Event']) == \
+           len(output[1]['tlo.methods.healthsystem']['HSI_Event'])
+    assert output[1]['tlo.methods.healthsystem']['HSI_Event']['did_run'].all()
```
tbhallett, line 245: a stronger test would be of the form:

`pd.testing.assert_frame_equal(output[0], output[1])`

tests/test_healthsystem.py
```diff
@@ -193,6 +193,58 @@ def test_run_no_interventions_allowed(tmpdir, seed):
     assert not any(sim.population.props['mi_status'] == 'P')  # No cures
 
 
+@pytest.mark.slow
+def test_policy_has_no_effect_on_mode1(tmpdir, seed):
+    # Events ran in mode 1 should be identical regardless of policy assumed.
+    # In policy "No Services", have set all HSIs to priority below lowest_priority_considered,
+    # in mode 1 they should all be scheduled and delivered regardless
```
tbhallett, line 200: By convention this should be doctoring (i.e., 
```python
"""This is a docstring, enclose by three sets of inverted commas, and immediately below definition line.""")
```

src/tlo/methods/healthsystem.py
```diff
@@ -2370,127 +2369,132 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
-                            if sum(event.BEDDAYS_FOOTPRINT.values()):
-                                event._received_info_about_bed_days = \
-                                    self.module.bed_days.issue_bed_days_according_to_availability(
-                                        facility_id=self.module.bed_days.get_facility_id_for_beds(
-                                                                           persons_id=event.target),
-                                        footprint=event.BEDDAYS_FOOTPRINT
-                                    )
+                            # Have enough capabilities left to run event
+                            else:
+                                # Notes-to-self: Shouldn't this be done after checking the footprint?
```
tbhallett, line 2431: The answer to this note is that the checking of bed-days should happen *just before* the HSI is runs. It doesn't gate the HSI running/not running, but should give information that is as up-to-date as possible.

src/tlo/methods/healthsystem.py
```diff
@@ -1313,8 +1310,8 @@ def schedule_hsi_event(
         # Check that priority is in valid range
         assert priority >= 0
 
-        # If priority of HSI_Event lower than the lowest one considered, ignore event in scheduling
-        if priority > self.lowest_priority_considered:
+        # If priority of HSI_Event lower than the lowest one considered, ignore event in scheduling under mode 2
+        if (self.mode_appt_constraints == 2) and (priority > self.lowest_priority_considered):
```
tbhallett, line 1314: why link this to mode?

src/tlo/methods/healthsystem.py
```diff
@@ -2260,7 +2257,9 @@ def _get_events_due_today(self,) -> Tuple[List, List]:
                 # The event is not yet due (before topen)
                 hp.heappush(_list_of_events_not_due_today, next_event_tuple)
 
-                if next_event_tuple.priority == self.module.lowest_priority_considered:
+                # Only stop querying when reaching lowest_priority_considered in mode 2
+                if (self.module.mode_appt_constraints == 2) and  \
+                   (next_event_tuple.priority == self.module.lowest_priority_considered):
```
tbhallett, line 2262: why only stop queuing when we're in mode 2?

tbhallett (changes requested):

src/tlo/methods/healthsystem.py
marghe-molaro: This is an optimisation check which ensures we stop querying the queue as soon as we've collected all events due today for priorities which are being considered (i.e. equal or lower than the lowest_priority_considered). This however should only be done in mode 2, in which lowest_priority_considered is meaningful.

src/tlo/methods/healthsystem.py
marghe-molaro: To ensure that even if adopting a policy which assigns certain HSIs a priority > lowest_priority_considered, in the mode 1 phase these are still queued.

tests/test_healthsystem.py
marghe-molaro: I was looking for something similar to this, but we are not comparing dataframes (outputs are tlo.analysis.utils.LogsDict instanges).

tests/test_healthsystem.py
marghe-molaro: I think iterating through all policies is unnecessary, and the test is already quite slow, especially if considering a full model. The comparison between two priority policies can already test all the differences we are after (i.e. different priority of services both above and below the lowest_priority_considered).

tests/test_healthsystem.py
tbhallett: Then unpack the dictionary do the check on each frame?

marghe-molaro: Sorry I confused myself. All fine now.

marghe-molaro:
All comments to test have been addressed, just checking a few additional things now. Please do not merge yet.

marghe-molaro:
The PR now ensures that:

In mode 0/1:
- lowest_priority_considered *is not* considered when adding HSIs to the queue;
- lowest_priority_considered *is not* considered when querying the queue for HSI events due today;
In mode 0/1, all HSIs are therefore always scheduled and delivered (provided the officer required is available in mode 1).

In mode 2:
- lowest_priority_considered *is* considered when adding HSIs to the queue;
- lowest_priority_considered *is* considered when querying the queue for / running HSIs due today (if capabilities are still available), i.e. we stop querying the queue once we reach future events for priority = lowest_priority_considered;
- When we go through the queue a final time to ensure did_not_run is called on all events that were postponed, we also "flush" any HSIs present which were potentially scheduled with a priority lower than the lowest_priority_considered during a mode 1 period. This will ensure that when we transition from mode 1 to 2 all those HSIs are correctly removed from the queue. 

PR is now ready for review and merge.

tests/test_healthsystem.py
```diff
@@ -193,6 +193,48 @@ def test_run_no_interventions_allowed(tmpdir, seed):
...
 
 
+@pytest.mark.slow
+def test_policy_has_no_effect_on_mode1(tmpdir, seed):
+    """Events ran in mode 1 should be identical regardless of policy assumed.
+    In policy "No Services", have set all HSIs to priority below lowest_priority_considered,
+    in mode 1 they should all be scheduled and delivered regardless"""
+
+    output = []
+    for i, policy in enumerate(["Naive", "Test Mode 1"]):
```
tbhallett, line 203: This test seems great now.

Ok, we don't have to have them all, but how about also the "No Policy Policy", and the "Clinical Vulnerable" one. Just so we have all bases covered.

tbhallett, line 203:
```suggestion
    for _, policy in enumerate(["Naive", "Test Mode 1", "", "ClinicallyVulnerable"]):
```

tbhallett (approved):

marghe-molaro:
I did a small test run (effect_of_policy-2023-10-31T055900Z) on this branch merged with PR #1169 to check that things look in order; pop size 15k, sim = 2 years long with first year (2010) in mode 1 and second year (2011) in mode 2, testing 3 different policies for tclose =1 day and tclose = 10 days.

Unfortunately it seems that for the first year (mode 1) the DALY output is not identical across scenarios as it should be. 

Screenshot 2023-10-31 at 11 32 13

tbhallett:
> I did a small test run (effect_of_policy-2023-10-31T055900Z) on this branch merged with PR #1169 to check that things look in order; pop size 15k, sim = 2 years long with first year (2010) in mode 1 and second year (2011) in mode 2, testing 3 different policies for tclose =1 day and tclose = 10 days.
> 
> Unfortunately it seems that for the first year (mode 1) the DALY output is not identical across scenarios as it should be.
> 
> Screenshot 2023-10-31 at 11 32 13

Good to check that. Will hold on merging for now then.

marghe-molaro:
Hi @tbhallett ,  some thoughts about non-identical outcomes in mode 1 for different policies assumed:
- The fact that test_policy_has_no_effect_on_mode1 is satisfied and that DALY outcomes are identical under the same priority policy but different tclose assumptions (see above) would suggest that in mode 1 all scheduled HSIs are being ran regardless of policy assumed *and* capacity, as expected; 
- Given that mode 1 is performing as would be expected in terms of number of HSIs delivered, I wonder whether the assumption that the DALY outcome in mode 1 should be completely independent of the order in which they are delivered (specified by the policy) is itself correct. If, for example, the same individual has multiple appts scheduled for the same day, could the order in which those appts are delivered impact diagnosis and treatment, and therefore ultimately the DALY outcome for the individual? E.g. if a person was screened and treated for HIV before/after receiving treatment for a disease X  for which HIV is a risk factor, could the DALY outcome due to disease X be affected? The test run I did only considered 15k individuals, which due to low pop may be biased towards a high incidence of comorbidities (?) so may be particularly vulnerable to this effect.

marghe-molaro:
@tbhallett False alarm, the different order of the queue simply means the random numbers used by the same HSI events will be different, so there is no reason the DALY outcomes should be exactly the same. With a pop size of 15k a difference of 10% seems entirely reasonable.

(In fact, even the subsequent HSIs might be different, so the test designed for this PR might well fail under different seeds). 

This PR is now ready with a small correction to the test made in the last commit.

tbhallett:
@marghe-molaro -- I resolved some conflicts on this with the base branch (through other changes in `master` since you branded off.). Please could you double-check that conflict resolution?

marghe-molaro:
@tbhallett looks good to me. I added a small clarifying comment to make sure we don't get confused down the line about the need to remove HSIs with priority > lowest_priority_considered from the queue in mode 2.

Issue #86, opened 2020-01-17
Saving to file simulations in a suspended state and resuming 


tbhallett:
We have  a common use case as follows:
- We want to run a simulation up to a certain point (e.g. to before some policy changes)
- Then we want to run the simulation from that point under multiple sets of assumptions ('forward projections')
- The state from which the simulation starts should be the same in each 'forward projection' 
- This could be accomplished through control of the random seed so that the first part of simulation. -- but it's wasteful to repeat verbatim the first part of the simulation.

It would seem like a solution to this would be to able to save the simulation at a certain point to a file. Then load up the file and resume the simulation under the same or different parametric conditions. 

I thought this might be relatively straight forward using pickle (i.e. pickle the sim: which contains the sim.population.props and the event_queue, and all the modules and their internal contents). Then, unpickle the sim, manipulate any parameters in the modules, and restart the sim using sim.simulate(end_date = end_of_part_two_date). (see script below)

However, I tried this and the unpicking failed with a RecursionError. Stack overflow suggested this is a common error for pickling complex classes and suggested increasing the limit on recursions -- but this led to the console crashing for me. 

Do you have any thoughts on this?

**Short-term**:
- Having a basic form of this functionality (even if hacky) would really help Iwona's MRes project. Any ideas would be very welcome!?

**Medium-term**:
- Will this be a part (or could it be?) of the run management system?

```
from pathlib import Path

from tlo import Date, Simulation
from tlo.methods import contraception, demography

outputpath = Path("./outputs")
resourcefilepath = Path("./resources")

start_date = Date(2010, 1, 1)
end_date_part_one = Date(2011, 1, 2)
popsize = 1000

sim = Simulation(start_date=start_date)
sim.register(demography.Demography(resourcefilepath=resourcefilepath))
sim.register(contraception.Contraception(resourcefilepath=resourcefilepath))
sim.seed_rngs(1)
sim.make_initial_population(n=popsize)
sim.simulate(end_date=end_date_part_one)



import pickle

with open(outputpath / 'pickled_basic_object', 'wb') as f:
    pickle.dump({'1': 1, '2': 2}, f)

with open(outputpath / 'pickled_sim', 'wb') as f:
    pickle.dump(sim, f)

with open(outputpath / 'pickled_event_queue', 'wb') as f:
    pickle.dump(sim.event_queue, f)

with open(outputpath / 'pickled_basic_object', 'rb') as f:
    x = pickle.load(f)

[...]

PR #1169, opened 2023-10-26
Each run in scenario has same seed, across all draws

tamuri:
Part of #86.

Each scenario has a seed. Each draw has runs. Each numbered run has a seed which is the same across all draws.

This can be tested using `tlo scenario-run` on a small pop size & run for 1 month then check the log file for the seed for "Simulation RNG".

### DIFF
--- a/src/tlo/scenario.py
+++ b/src/tlo/scenario.py
@@ -275,7 +275,6 @@ class DrawGenerator:
     def get_draw(self, draw_number):
         return {
             "draw_number": draw_number,
-            "draw_seed": self.scenario.rng.randint(MAX_INT),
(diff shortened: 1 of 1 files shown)


src/tlo/scenario.py
```diff
@@ -327,7 +326,7 @@ def get_sample(self, draw, sample_number):
         # Instead of using the random number generator to create a seed for the simulation, we use an integer hash
         # function to create an integer based on the sum of the draw_number and sample_number. This means the
```
matt-graham, line 327: I this description still correct (and/or was it correct for previous implementation)? It looks like the only `sample_number` feeds in to generated seed not `draw_number`.

matt-graham (approved):
Looks good to me, other than minor question about comment text.

PR #1178, opened 2023-11-07
Increase HTML docs content column width on wide screens

### LINKED ISSUE #1175: displayed a smaller number of characters than we use per line
A smaller number of characters than what we use per line is displayed in the documentation. Could be the entire lines displayed there?

Here is an example where I noticed this:
https://www.tlomodel.org/_modules/tlo/methods/contraception.html#StartInterventions

matt-graham:
Fixes #1175 

Increases content column width in HTML documentation from maximum of 800px to 960px, when on screens 1280px or wider, which should avoid overflow of fixed width text content in source code blocks with the current default line length of 120 characters.

Building locally the CSS override seems to be getting picked up for the relevant `div` without needing any `!important` directives.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,6 +68,12 @@ html_static_path = ['_static']
 html_extra_path = ['./.nojekyll', './.gitattributes']
 html_favicon = 'favicon.png'
 
+html_context = {
+    'css_files': [
+        '_static/theme_overrides.css',
+    ],
+}
+
 rawfiles = ['.github']
 
 napoleon_use_ivar = True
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -68,6 +68,12 @@
 html_extra_path = ['./.nojekyll', './.gitattributes']
 html_favicon = 'favicon.png'
 
+html_context = {
```
tamuri, line 71: The latest versions of Sphinx & rtd-docs don't support adding css files like this. See https://github.com/readthedocs/sphinx_rtd_theme/issues/1415

Need to use `html_css_files`

matt-graham: Thanks - updated to use `html_css_files

tamuri (approved):

PR #1188, opened 2023-11-13
Refactor population ratio computation in contraception module to avoid `FutureWarning`

### LINKED ISSUE #1187: `FutureWarning` raised in contraception module due to single element `Series` conversion to `float`
When running simulations I get three `FutureWarning` instances from contraception module

```
TLOmodel/src/tlo/methods/contraception.py:371: FutureWarning: Calling float on a single element Series is deprecated and will raise a TypeError in the future. Use float(ser.iloc[0]) instead
  probs_30plus['female_sterilization'] = \
TLOmodel/src/tlo/methods/contraception.py:448: FutureWarning: Calling float on a single element Series is deprecated and will raise a TypeError in the future. Use float(ser.iloc[0]) instead
  float(
TLOmodel/src/tlo/methods/contraception.py:371: FutureWarning: Calling float on a single element Series is deprecated and will raise a TypeError in the future. Use float(ser.iloc[0]) instead
  probs_30plus['female_sterilization'] = \
```

which references

https://github.com/UCL/TLOmodel/blob/02ed59fea1a453a00e9537d54c7c877b2e268af7/src/tlo/methods/contraception.py#L371-L373

and 

https://github.com/UCL/TLOmodel/blob/02ed59fea1a453a00e9537d54c7c877b2e268af7/src/tlo/methods/contraception.py#L447-L450

While the latter has an explicit `float` call the former doesn't so I assume means there is some call to `float` on pandas side?

The `self.n_female1549_in_2010` and `self.n_female3049_in_2010` attributes are defined in 

https://github.com/UCL/TLOmodel/blob/02ed59fea1a453a00e9537d54c7c877b2e268af7/src/tlo/methods/contraception.py#L223-L224

and in addition to the above two usages which raised warnings are also used in

https://github.com/UCL/TLOmodel/blob/02ed59fea1a453a00e9537d54c7c877b2e268af7/src/tlo/methods/contraception.py#L650-L652

In all three cases it is only the ratio of `self.n_female1549_in_2010` and `self.n_female3049_in_2010` that is needed, so probably would make more sense to calculate this once (in a way which avoids the `FutureWarning` and use this ratio everywhere.

matt-graham:
Fixes #1187 

Just noticed on going to create this PR, that @tbhallett has already submitted a fix for this in #1173.

Both the approach there and here should work. The main difference are this PR avoids the `groupby` operations when computing the number of females aged 15&ndash;49 and 40&ndash;49 in the `read_parameters` method, thus directly computing floating point values rather than single element series, and then computes and stores only the ratio in `read_parameters` rather than both values separately. @tbhallett's approach instead references the `Count` element in the single element series thus avoiding the warning.

There is one further instance of a ratio `self.n_female3049_in_2010 / self.n_female1549_in_2010` being computed in the module in 
https://github.com/UCL/TLOmodel/blob/02ed59fea1a453a00e9537d54c7c877b2e268af7/src/tlo/methods/contraception.py#L650-L652
that is unchanged in #1173 which we probably also want to update if merging #1173 instead of this (this didn't raise a warning when I did a short simulation but I suspect that may just be due to the enclosing method not being called).

### DIFF
--- a/src/tlo/methods/contraception.py
+++ b/src/tlo/methods/contraception.py
@@ -216,12 +216,16 @@ class Contraception(Module):
             pd.read_csv(Path(self.resourcefilepath) / 'demography' / 'ResourceFile_ASFR_WPP.csv')
 
         # Import 2010 pop and count numbs of women 15-49 & 30-49
-        pop_2010 = self.sim.modules["Demography"].parameters["pop_2010"].copy()
-
-        female1549_in_2010 = \
-            pop_2010.loc[(pop_2010.Sex == 'F') & pop_2010.Age.between(15, 49), ['Age', 'Count']].groupby('Age').sum()
-        self.n_female1549_in_2010 = female1549_in_2010.sum()
-        self.n_female3049_in_2010 = female1549_in_2010.loc[female1549_in_2010.index >= 30].sum()
+        pop_2010 = self.sim.modules["Demography"].parameters["pop_2010"]
+        n_females_aged_15_to_49_in_2010 = pop_2010[
+            (pop_2010.Sex == "F") & (pop_2010.Age >= 15) & (pop_2010.Age <= 49)
+        ].Count.sum()
+        n_females_aged_30_to_49_in_2010 = pop_2010[
+            (pop_2010.Sex == "F") & (pop_2010.Age >= 30) & (pop_2010.Age <= 49)
+        ].Count.sum()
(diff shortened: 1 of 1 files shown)


tbhallett (approved):
Thanks for this @matt-graham 
This seems a much more comprehensive approach than my solution (where I was trying to be as "surgical" as possible). So, fine for this to go in, and for the earlier PR to be closed.

Issue #1163, opened 2023-10-13
Out-of-memory jobs killed by Azure


marghe-molaro:
The following job was killed by Azure due to RAM issue despite a relatively small population size (150k)
long_run_all_diseases-2023-10-09T081418Z

Many runs failed for this job in mode 2, pop size 100k
effect_of_policy-2023-10-03T165843Z

marghe-molaro:
Hi @tamuri, just checking whether we have any new insights as to what may have been causing the excess in requested RAM for these jobs? If not I think I will try to resubmit switching off the conctraception logger (which is not required) and the tlo.methods.healthsystem.summary (which as suggested in PR 1157 may not be behaving as expected) to see if that's enough to take these jobs past the finish line.

tamuri:
I haven't investigated but I don't think it's the logs because it runs out of memory during the simulation run, not the post-processing of the log file (where are large log file has caused problems in the past). I'm guessing it's the population size. I'll take a look tomorrow.

tamuri:
Quick test. This is little under 9m of simulation time. There are events causing big spikes in memory usage The Azure machine have 7.5GB memory, so this would have already been killed.

![Rplot](https://github.com/UCL/TLOmodel/assets/360002/d240284d-9da2-4e94-8ce9-8a55a819fcc0)

Need to do some more digging, but top offenders are:

```
   total_gb increase_gb                                                                   event
1  4.784164   4.0286446  <tlo.methods.healthsystem.HealthSystemScheduler object at 0x133317ad0>
2  5.274666   3.7423286  <tlo.methods.healthsystem.HealthSystemScheduler object at 0x133317ad0>
3  4.246059   3.7283936     <tlo.methods.contraception.ContraceptionPoll object at 0x13354ae10>
4  4.491314   3.5014229           <tlo.methods.contraception.DirectBirth object at 0x137beeb10>
5  3.891785   2.8981476  <tlo.methods.healthsystem.HealthSystemScheduler object at 0x133317ad0>
6  3.465004   2.6420860               <tlo.methods.alri.AlriPollingEvent object at 0x132d35b10>
7  2.942112   2.2692070               <tlo.methods.alri.AlriIncidentCase object at 0x1354afd90>
8  2.839962   1.9623909           <tlo.methods.contraception.DirectBirth object at 0x137bd9fd0>
9  2.612389   1.8963928  <tlo.methods.healthsystem.HealthSystemScheduler object at 0x133317ad0>
[...]

PR #1199, opened 2023-11-15
Refactor lookup table computation in malaria poll to avoid memory spike

matt-graham:
In https://github.com/UCL/TLOmodel/issues/1163#issuecomment-1811438520 I noted that from memory profiling of a run of the `src/scripts/profiling/scale_run.py` script, there appeared to be sharp peaks in memory usage corresponding to `Malaria.malaria_poll2` method called at

https://github.com/UCL/TLOmodel/blob/4bffc1a3c8d5ae7336ea4534486830255a0173e1/src/tlo/methods/malaria.py#L652

which is in a monthly scheduled event `MalariaPollingEventDistrict`, with around ~300MB increase in memory usage during these peaks, as visible in the following plot of overall memory usage:

image

Looking at the flamegraph representation of the peak memory usage in a narrow window around one of these spikes we get the following breakdown for the `Malaria.malaria_poll2` method call

image

with the width of the bars proportional to the total memory allocated by the function (and each bar a function / method in the call stack with lower bars further 'down' / nested in the call stack).

This suggested it was evaluations of the `_draw_incidence_for` nested function which was the key culprit, and in particular the line

https://github.com/UCL/TLOmodel/blob/4bffc1a3c8d5ae7336ea4534486830255a0173e1/src/tlo/methods/malaria.py#L306

This line selects a subset of the population dataframe rows, sets the index of this resulting new dataframe to use the values of the `district_num_of_residence` and `ma_age_edited` columns, and then gets just the index of this resulting dataframe. As the dataframe returned by the initial indexing by `_where` to get a row subset will still contain all the columns in the original population dataframe, I suspect this is what is leading to the large memory allocations. As we subsequently only need the `district_num_of_residence` and `ma_age_edited` columns, we can use the `loc` indexer to simultaenously select a subset of rows using `_where` and the two relevant columns, then reindex this much smaller (in terms of number of columns) dataframe inplace avoiding a further dataframe copy and memory allocation, before extracting the index. With these changes as implemented in this PR we get the following overall memory usage plot

image

which does seems to have had the desired effect of substantially decreasing the amplitude of the spikes.

### DIFF
--- a/src/tlo/methods/malaria.py
+++ b/src/tlo/methods/malaria.py
@@ -303,7 +303,9 @@ class Malaria(Module):
         def _draw_incidence_for(_col, _where):
             """a helper function to perform random draw for selected individuals on column of probabilities"""
             # create an index from the individuals to lookup entries in the current incidence table
-            district_age_lookup = df[_where].set_index(['district_num_of_residence', 'ma_age_edited']).index
+            district_age_subset = df.loc[_where, ['district_num_of_residence', 'ma_age_edited']]
+            district_age_subset.set_index(district_age_subset.columns.to_list(), inplace=True)
+            district_age_lookup = district_age_subset.index
             # get the monthly incidence probabilities for these individuals
             monthly_prob = curr_inc.loc[district_age_lookup, _col]
             # update the index so it's the same as the original population dataframe for these individuals


matt-graham:
Just to note - I don't think this is a significant factor in the out-of-memory issues described in #1163, but I thought it was worth submitting a PR to fix given I'd noticed it as part of the investigations there!

tdm32 (approved):
This looks fine to me @matt-graham. Thank you

tamuri (approved):

PR #1201, opened 2023-11-15
[CI] Add GitHub Actions workflow to run calibration analysis

giordano:
Opening as draft as I'm running some more tests in [my fork](https://github.com/giordano/TLOmodel), but I had some successful runs in another (private) repositories.

### TODO

* [x] Add `test` label to all other runners
  * Question: do we need to reprovision all virtual machines with #1149, or we keep the existing runners and manually add the labels?
* [x] Should we mount the temp directory with the output of the calibrations to a permanent storage?
* [x] Increase runs number to 14
* [x] Revert the extra commit for testing before merging

### DIFF
--- a/.github/workflows/comment-triggered-ci.yml
+++ b/.github/workflows/comment-triggered-ci.yml
@@ -10,7 +10,7 @@ jobs:
   scaled_sim:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: self-hosted
+      runs-on: [self-hosted, test]
       keyword: scaled-sim
       commands: tox -v -e profile
       description: Profiled scale run of model
@@ -23,7 +23,7 @@ jobs:
   multiple_seed_tests:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: self-hosted
+      runs-on: [self-hosted, test]
       keyword: multiple-seed-tests
       commands: |
         tox -v -e py311,report -- pytest --seed 2671936806 3512589365 --cov --cov-report=term-missing -vv tests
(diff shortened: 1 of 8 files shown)


.github/workflows/comment-triggered-ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   scaled_sim:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: self-hosted
+      runs-on: [self-hosted, test]
```
giordano, line 13: We'll need to add this label to all the other self-hosted runners, so that we can better control which workflows run on which machines.

src/scripts/calibration_analyses/scenarios/long_run_all_diseases.py
```diff
@@ -21,8 +21,8 @@ def __init__(self):
         super().__init__()
         self.seed = 0
         self.start_date = Date(2010, 1, 1)
-        self.end_date = Date(2030, 1, 1)
-        self.pop_size = 20_000
+        self.end_date = Date(2020, 1, 1)
+        self.pop_size = 1000
```
giordano, line 25: This is just for testing, will be reverted before merging

giordano:
Successful run at https://github.com/giordano/TLOmodel/actions/runs/6877383425 :partying_face: It's a small scale simulation with shorter duration and smaller population size, just to show the pipeline works.

giordano:
I added the possibility to trigger the calibration with workflow dispatch, specifying a custom commit hash (note: it has to be full 40-character hash, not an abbreviated one).  It's a bit complicated to test because as far as I understand the workflow file has to be present on the default branch of the repository in order for the workflow dispatch to show up.  I did some tests in another repository and I _think_ it's working now, but we can probably test better and fix any problem after this PR is merged.

giordano:
This should be ready now :rocket:

tamuri (approved):

PR #1216, opened 2023-11-27
JSON encode sequences used as `runs-on` input to `run-on-comment` workflow

### LINKED ISSUE #1215: Profiling workflow failing due to passing sequence to `run-on` input to `run-on-comment` workflow
The profiling workflows are currently failing with a message

```
Invalid workflow file: .github/workflows/run-profiling.yaml#L54
The workflow is not valid. .github/workflows/run-profiling.yaml (Line: 54, Col: 16): A sequence was not expected
```

due to

https://github.com/UCL/TLOmodel/blob/9bf84d9d679b68253eb766cc85fb6ed5c3f9a344/.github/workflows/run-profiling.yaml#L53-L54

which passes a sequence of runner labels to the `runs-on` input to the reusable workflow in `.github/workflows/run-on-comment.yaml`.

The input is defined to be of string type

https://github.com/UCL/TLOmodel/blob/9bf84d9d679b68253eb766cc85fb6ed5c3f9a344/.github/workflows/run-on-comment.yml#L6-L10

and it looks like from the error that reusable workflow inputs cannot be set to sequences.

We could possibly use [the `fromJSON` function](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson) in an expression to convert from a JSON encoded string representation of a list / sequence, though this would require remembering to enclose the list in quotation marks to ensure it is passed as a string I assume.

matt-graham:
Fixes #1215 

Makes all sequence values passed to `runs-on` input to` run-on-comment` reusable workflow JSON encoded strings rather than raw sequences and uses `fromJSON` function to decode string when passing to triggered workflow job's `run-on` property.

### DIFF
--- a/.github/workflows/comment-triggered-ci.yml
+++ b/.github/workflows/comment-triggered-ci.yml
@@ -10,7 +10,7 @@ jobs:
   scaled_sim:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: [self-hosted, test]
+      runs-on: "[self-hosted, test]"
       keyword: scaled-sim
       commands: tox -v -e profile
       description: Profiled scale run of model
@@ -23,7 +23,7 @@ jobs:
   multiple_seed_tests:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: [self-hosted, test]
+      runs-on: "[self-hosted, test]"
       keyword: multiple-seed-tests
       commands: |
         tox -v -e py311,report -- pytest --seed 2671936806 3512589365 --cov --cov-report=term-missing -vv tests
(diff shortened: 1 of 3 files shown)


tamuri (approved):

.github/workflows/comment-triggered-ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   scaled_sim:
     uses: ./.github/workflows/run-on-comment.yml
     with:
-      runs-on: [self-hosted, test]
+      runs-on: "[self-hosted, test]"
```
giordano, line 13: Out of curiosity, why is this a string of a list?  According to [the documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on) the value should be a single string, or a variable containing a list, or an array of strings/variable containing strings,or key-value pairs.

PR #1219, opened 2023-11-28
life expectancy estimates

tdm32:
This PR contains a script src/tlo/analysis/life_expectancy.py which estimates male and female life expectancy from birth using the outputs contained in a results_folder. Life expectancy is estimated for every run/draw and can be summarised to give a median with 95% uncertainty intervals across each draw using arg median=True. 

The accompanying document in folder /write-ups details the methods used to generate the estimates. 

The function used to produce and summarise the life expectancy estimates could potentially be housed in file utils.py with the helper functions hosted elsewhere.


src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    - pd.DataFrame: The DataFrame with the 'age-group' column added.
+    """
+    # Define age groups in 5-year intervals
+    age_groups = ['0'] + ['1-4'] + [f'{start}-{start + 4}' for start in range(5, 90, 5)] + ['90+']
+
+    # Create a new column 'age-group' based on the age-to-age-group mapping
+    dataframe['age_group'] = pd.cut(dataframe['age'], bins=[0] + [1] + list(range(5, 95, 5)) + [float('inf')],
+                                    labels=age_groups, right=False)
+
+    return dataframe
```
tbhallett, line 46: I think more natural if this function returns a `pd.Series` contains the mapped age-groups, rather than directly editing the data frame in place.

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+def _create_multi_index_columns(results_folder: Path):
+    info = get_scenario_info(results_folder)
+
+    return pd.MultiIndex.from_product([range(info['number_of_draws']),
+                                       range(info['runs_per_draw'])],
+                                      names=['draw', 'run'])
+
+
+# %% extract key population data for life expectancy calculations
```
tbhallett, line 57: no need for the script-style comments when you have the nice docstrings

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    multi_index_columns = _create_multi_index_columns(results_folder)
+    res = pd.DataFrame(columns=multi_index_columns)
+
+    for draw in range(info['number_of_draws']):
+        for run in range(info['runs_per_draw']):
+            df: pd.DataFrame = load_pickled_dataframes(results_folder, draw, run, module)[module][key]
+            output_from_eval: pd.Series = df.loc[pd.to_datetime(df.date).between(*TARGET_PERIOD)].stack()
+            res.loc[:, (draw, run)] = output_from_eval[1:]
+
+    return res
```
tbhallett, line 97: I'm wondering why we don't use `extract_results` for all this, in the way done on `_num_deaths_by_age_group`?

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+                    output.columns.unique(level='draw'),
+                    ["median", "lower", "upper"]
+                ],
+                names=['draw', 'stat']),
+            index=output.index
+        )
+
+        summary.loc[:, (slice(None), "median")] = output.groupby(axis=1, by='draw').median().values
+        summary.loc[:, (slice(None), "lower")] = output.groupby(axis=1, by='draw').quantile(0.025).values
+        summary.loc[:, (slice(None), "upper")] = output.groupby(axis=1, by='draw').quantile(0.975).values
```
tbhallett, line 303: wondering if w should use `summarize` for this, as it seems to repeat that logic (?)

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    person_years = _get_population_size_by_age(results_folder)
+
+    for draw in range(info['number_of_draws']):
+        for run in range(info['runs_per_draw']):
+            # select column
+            le = _estimate_life_expectancy(
+                number_of_deaths_in_interval=deaths.loc[:, (draw, run)],
+                person_years_at_risk=person_years.loc[:, (draw, run)])
+
+            output = pd.concat([output, le], axis=1)
```
tbhallett, line 281: small point, but pandas is much faster if you collect these columns into a list, and then do `pd.concat` once on that  list. (I expect speed isn't an issue for this one, but just saying in case you see a situation that is more time-critical).

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # Create a MultiIndex for rows using age group and 'Male' and 'Female'
+    multi_index_rows_male = pd.MultiIndex.from_product([t1.index, ['M']], names=['age_group', 'sex'])
+    multi_index_rows_female = pd.MultiIndex.from_product([t1.index, ['F']], names=['age_group', 'sex'])
+    t3.index = multi_index_rows_male.append(multi_index_rows_female)
+
+    return t3
+
+
+# %% GENERATE LIFE EXPECTANCY ESTIMATES
```
tbhallett, line 164:
```suggestion

```

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+        results_folder,
+        module="tlo.methods.demography",
+        key="death",
+        custom_generate_series=extract_deaths_by_age_group,
+        do_scaling=False
+    )
+
+
+# get population size in target period
+# get one row of log for each draw/run and compile into df
```
tbhallett, line 80: move into doctoring

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
+"""
+This file will read in the output files generated by model runs
+then employs a series of helper functions to
+generate life tables to estimate life expectancy for each run/draw
+
+function: produce_life_expectancy_estimates() will produce summary statistics
+or will output life expectancy at birth for M and F separately for every
+run/draw if argument(...median=False)
```
tbhallett, line 8: is there a reason to favour 'median' for the summary of this measure across runs (c.f. `mean`)?

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    t1 = pd.concat([num_m0, num_males])
+    # set correct index
+    t1.index.values[0] = '0'
+
+    # repeat for females
+    num_females_adj = num_females.iloc[0:1, :] - num_f0.iloc[0:1, :].values
+    num_females.iloc[0] = num_females_adj.values
+    num_females.index.values[0] = '1-4'
+    t2 = pd.concat([num_f0, num_females])
+    t2.index.values[0] = '0'
```
tbhallett, line 137: as all the logic is repeat for M and F, it would make sense to have it written once

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # combine pop aged 90+
+    # Sum the last three rows column-wise
+    last_three_rows_sum = t1.iloc[-3:].sum()
```
tbhallett, line 141: it may be "unsafe" to rely on it being last three rows. the index would have had the ages in (if if wasn't dropped) and then you could do a `Slice` to get everyone above 90.

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    """
+    for a single run, estimate life expectancy for males and females
+    return: pd.Series
+    """
+
+    estimated_life_expectancy = pd.Series(dtype=float)
+
+    # first age-group is 0, then 1-4, 5-9, 10-14 etc. 22 categories in total
+    level_0_values = person_years_at_risk.index.get_level_values('age_group').unique()
+    interval_width = pd.Series([1] + [4] + [5] * 18, index=level_0_values)
```
tbhallett, line 176: I notice that level_0_values is not sorted before/after using `unqiue` so we don't know the order of the values. The logic on line 176 relies on the age-groups being as defined somewhere else. So, I think its much safer to work do a sort of the values and work out the width directly what you get: e.g. using `diff`.

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    # separate male and female data
+    for sex in ['M', 'F']:
+        person_years_by_sex = person_years_at_risk.xs(key=sex, level='sex')
+        number_of_deaths_by_sex = number_of_deaths_in_interval.xs(key=sex, level='sex')
+
+        death_rate_in_interval = number_of_deaths_by_sex / person_years_by_sex
+        # if no deaths or person-years, convert nan to 0
+        death_rate_in_interval = death_rate_in_interval.fillna(0)
+        # if no deaths in age 90+, set death rate equal to value in age 85-89
+        if death_rate_in_interval[number_age_groups - 1] == 0:
```
tbhallett, line 190: index to '90+' rather than with reference to it being the last

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+            death_rate_in_interval[number_age_groups - 1] = death_rate_in_interval[number_age_groups - 2]
+
+        # Calculate the probability of dying in the interval
+        condition = number_of_deaths_by_sex > (
+            person_years_by_sex / interval_width / fraction_of_last_age_survived)
```
tbhallett, line 195: brackets would help make sense of this (for me!)

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+        death_rate_in_interval = number_of_deaths_by_sex / person_years_by_sex
+        # if no deaths or person-years, convert nan to 0
+        death_rate_in_interval = death_rate_in_interval.fillna(0)
+        # if no deaths in age 90+, set death rate equal to value in age 85-89
+        if death_rate_in_interval[number_age_groups - 1] == 0:
+            death_rate_in_interval[number_age_groups - 1] = death_rate_in_interval[number_age_groups - 2]
+
+        # Calculate the probability of dying in the interval
+        condition = number_of_deaths_by_sex > (
```
tbhallett, line 194: what does condition refer to? (suggest to add comments)

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+        number_alive_at_start_of_interval = pd.Series(index=range(number_age_groups), dtype=float)
+        number_alive_at_start_of_interval[0] = 100_000  # hypothetical cohort
+        for i in range(1, number_age_groups):
+            number_alive_at_start_of_interval[i] = (1 - probability_of_dying_in_interval[i - 1]) * \
+                                                   number_alive_at_start_of_interval[i - 1]
+
+        # number_dying_in_interval
+        number_dying_in_interval = pd.Series(index=range(number_age_groups), dtype=float)
+        for i in range(0, number_age_groups - 1):
+            number_dying_in_interval[i] = number_alive_at_start_of_interval[i] - number_alive_at_start_of_interval[
```
tbhallett, line 215: this should match `number_of_deaths_by_sex`, right?
good to check that?! and is it necessary to compute it this way?

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+    multi_index_rows_male = pd.MultiIndex.from_product([t1.index, ['M']], names=['age_group', 'sex'])
+    multi_index_rows_female = pd.MultiIndex.from_product([t1.index, ['F']], names=['age_group', 'sex'])
+    t3.index = multi_index_rows_male.append(multi_index_rows_female)
+
+    return t3
+
+
+# %% GENERATE LIFE EXPECTANCY ESTIMATES
+
+def _estimate_life_expectancy(person_years_at_risk, number_of_deaths_in_interval):
```
tbhallett, line 166: naming of this argument is a bit confusing, as it's actually population size (although could be PY from the demography module)

tbhallett (changes requested):
Thanks very much for this @tdm32 
This will be super helpful. Please find some checks (mostly in the interest of 'safety').

I did wonder about the estimate of PY. We're using population size to estimate PY, as that is what demographers have to do (because they see population size). And they have to go through some steps to estimate PY. But, we can calculate it accurately (and do). So, I'm wondering if it's better to use that. It would also be fewer steps.

We have the PY by single-year ages and deaths by single-years ages. We can then aggregate both on any set of age-groups and time periods defined by the user, and then we have the mortality schedule. (e.g. here for ALRI: 

I may well be missing something though, as I believe I you said that you had done in this way and then changed your mind.

tdm32:
> Hi @tdm32,
> 
> I am trying to use the life expectancy script on my outputs and I'm getting the following error:
> 
> File "src/tlo/analysis/life_expectancy.py", line 176, in _estimate_life_expectancy interval_width = pd.Series([1] + [4] + [5] * 18, index=level_0_values) File "/Users/mm2908/miniconda3/envs/tlo/lib/python3.8/site-packages/pandas/core/series.py", line 500, in **init** com.require_length_match(data, index) File "/Users/mm2908/miniconda3/envs/tlo/lib/python3.8/site-packages/pandas/core/common.py", line 576, in require_length_match raise ValueError( ValueError: Length of values (20) does not match length of index (23)
> 
> level_0_values for me returns the following: My Index([ '0', 10, '1-4', '5-9', '10-14', '15-19', '20-24', '25-29', '30-34', '35-39', '40-44', '45-49', '50-54', '55-59', '60-64', '65-69', '70-74', '75-79', '80-84', '85-89', 'date', '0-4', '90+'], dtype='object', name='age_group')
> 
> Based on your description it seems that three values shouldn't be included, namely "10" (which could be the number of runs per draw), "date", and "0-4" (since 0 and 1-4 are two distinct categories). I'm not sure how they're making their way into the dataframe, just wondering if you had any advice (no rush)!

Hi @marghe-molaro , that is really strange. I'll try and look at it tomorrow.

tdm32:
> Thanks very much for this @tdm32 This will be super helpful. Please find some checks (mostly in the interest of 'safety').
> 
> I did wonder about the estimate of PY. We're using population size to estimate PY, as that is what demographers have to do (because they see population size). And they have to go through some steps to estimate PY. But, we can calculate it accurately (and do). So, I'm wondering if it's better to use that. It would also be fewer steps.
> 
> We have the PY by single-year ages and deaths by single-years ages. We can then aggregate both on any set of age-groups and time periods defined by the user, and then we have the mortality schedule. (e.g. here for ALRI:
> 
> I may well be missing something though, as I believe I you said that you had done in this way and then changed your mind.

@tbhallett I don't think the link is showing here - is it a script for extracting deaths due to ALRI that you were referencing?

src/tlo/analysis/life_expectancy.py
tdm32: I've removed this

tdm32:
> Hi @tdm32,
> 
> I am trying to use the life expectancy script on my outputs and I'm getting the following error:
> 
> File "src/tlo/analysis/life_expectancy.py", line 176, in _estimate_life_expectancy interval_width = pd.Series([1] + [4] + [5] * 18, index=level_0_values) File "/Users/mm2908/miniconda3/envs/tlo/lib/python3.8/site-packages/pandas/core/series.py", line 500, in **init** com.require_length_match(data, index) File "/Users/mm2908/miniconda3/envs/tlo/lib/python3.8/site-packages/pandas/core/common.py", line 576, in require_length_match raise ValueError( ValueError: Length of values (20) does not match length of index (23)
> 
> level_0_values for me returns the following: My Index([ '0', 10, '1-4', '5-9', '10-14', '15-19', '20-24', '25-29', '30-34', '35-39', '40-44', '45-49', '50-54', '55-59', '60-64', '65-69', '70-74', '75-79', '80-84', '85-89', 'date', '0-4', '90+'], dtype='object', name='age_group')
> 
> Based on your description it seems that three values shouldn't be included, namely "10" (which could be the number of runs per draw), "date", and "0-4" (since 0 and 1-4 are two distinct categories). I'm not sure how they're making their way into the dataframe, just wondering if you had any advice (no rush)!

Hi @marghe-molaro , I've pushed one fix to an append call but I can't reproduce your error. Tim's suggestion above will switch the age-groups to a dict but I'll have to update the other functions to use this. 

The age-groups are created by:     age_groups = ['0'] + ['1-4'] + [f'{start}-{start + 4}' for start in range(5, 90, 5)] + ['90+']
is it possible to check where the additional age-group labels are created?

marghe-molaro:
Hi @tdm32, 

I noticed that when using the _extract_pop_during_target_period, the dataframe returned looks a bit asymmetric (see image below), in that the second year's results include the "date" index, whereas the first one doesn't; there also seems to be an additional index (9 and 10?) which could be making their way in. I just wanted to double check if that's how the dataframe looks for you too?

Screenshot 2023-12-07 at 12 35 19

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    Args:
+    - dataframe (pd.DataFrame): The DataFrame containing the age data.
+    - age_column (str): The name of the column containing ages.
+
+    Returns:
+    - pd.DataFrame: The DataFrame with the 'age-group' column added.
+    """
+    # Define age groups in 5-year intervals
+    age_groups = ['0'] + ['1-4'] + [f'{start}-{start + 4}' for start in range(5, 90, 5)] + ['90+']
```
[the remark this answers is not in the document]
tdm32: I couldn't get this working using the age_groups.values() which are strings and can't be used as bins so I've left it as it is for now. Happy to come back to it later.

src/tlo/analysis/life_expectancy.py
tdm32: This is changed now and the new dataframe column is set within the function where it is needed

src/tlo/analysis/life_expectancy.py
tdm32: if we move to extracting person-years directly we won't need this part of the code

src/tlo/analysis/life_expectancy.py
tdm32: Yes good point - I've replaced this now

src/tlo/analysis/life_expectancy.py
tdm32: Thanks Tim - I've made this change

src/tlo/analysis/life_expectancy.py
tdm32: removed

src/tlo/analysis/life_expectancy.py
tdm32: This part of the code has been removed and replaced by person-years instead of population size

src/tlo/analysis/life_expectancy.py
tdm32: yes this is definitely a bit wordy - removed now and replaced by person-years estimation

src/tlo/analysis/life_expectancy.py
tdm32: yes good point - removed now

src/tlo/analysis/life_expectancy.py
tdm32: I've made this change now

src/tlo/analysis/life_expectancy.py
tdm32: changed now

src/tlo/analysis/life_expectancy.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # number_dying_in_interval
+        number_dying_in_interval = pd.Series(index=range(number_age_groups), dtype=float)
+        for i in range(0, number_age_groups - 1):
+            number_dying_in_interval[i] = number_alive_at_start_of_interval[i] - number_alive_at_start_of_interval[
+                i + 1]
+
+        number_dying_in_interval[number_age_groups - 1] = number_alive_at_start_of_interval[number_age_groups - 1]
+
+        # person-years lived in interval
```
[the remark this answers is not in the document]
tdm32: Yes I was following the standard life-table approach. I can use the person-years within each interval instead of population size at the start of the interval to generate the death rates within intervals, but using the hypothetical cohort for the actual life expectancy calculation makes it more stable as running this on small populations will often lead to errors. If the number of person-years in any interval is 0 then it won't run. This is especially an issue for older populations in scenarios where we have high mortality. 

Also, I think generally a population size of 5000 is required for reliable estimates of life expectancy. As many people (including me) might want to run this function on small test populations, I went for the "safer" version using the hypothetical cohort. If we feel strongly that this shouldn't be allowed to run on small populations, then I'm happy to adjust the code.

src/tlo/analysis/life_expectancy.py
tdm32: I've added this explanation now: 
        # Calculate the probability of dying in the interval
        # condition checks whether the observed number deaths is significantly higher than would be expected
        # based on population years at risk and survival fraction
        # if true, suggests very high mortality rates and returns value 1

src/tlo/analysis/life_expectancy.py
tdm32: I think hopefully the comment added in the code should clarify this part

src/tlo/analysis/life_expectancy.py
tdm32: It would be scaled to reflect the hypothetical population size so in absolute numbers may differ, the mortality rates should be identical

tbhallett:
Thanks @tdm32 

@marghe-molaro  - as discussed, let us know if this function works smoothly for you in your application.

marghe-molaro:
Hi @tdm32 @tbhallett, 

I can now run the life expectancy script on my outputs without any problem, and from a first glance the results seem to be reasonable (e.g. shorter life expectancy from no healthcare system compared to mode 2).

tbhallett:
> Hi @tdm32 @tbhallett, 
> 
> 
> 
> I can now run the life expectancy script on my outputs without any problem, and from a first glance the results seem to be reasonable (e.g. shorter life expectancy from no healthcare system compared to mode 2).

Great thanks. And the value as expected? Ie matching life expectancy for Malawi for our status quo?

marghe-molaro:
> > Hi @tdm32 @tbhallett,
> > I can now run the life expectancy script on my outputs without any problem, and from a first glance the results seem to be reasonable (e.g. shorter life expectancy from no healthcare system compared to mode 2).
> 
> Great thanks. And the value as expected? Ie matching life expectancy for Malawi for our status quo?

I believe so - below are the estimates in the 2010-2020 range.  Draw 0 is the one with no healthcare system, all other draws should be statistically identical as they assumed mode 1 in that period. Means are around 65-66 years of life expectancy which I believe is consistent with expectation?

Screenshot 2024-02-07 at 12 36 05

tdm32:
> @tdm32 -- the document `EstimatingLifeExpectancy.docx`: am I right in thinking that this is now slightly out-dated, with the calculation moved to using person-years, or not?

@tbhallett yes it will be - I'll have to update it. Can we push without the doc and I can do a PR for that when it's ready?

tbhallett:
> > @tdm32 -- the document `EstimatingLifeExpectancy.docx`: am I right in thinking that this is now slightly out-dated, with the calculation moved to using person-years, or not?
> 
> @tbhallett yes it will be - I'll have to update it. Can we push without the doc and I can do a PR for that when it's ready?

yes, that's fine. I'll take the doc off for now. (Maybe with the updated method using PY a doc is less needed?)

tbhallett (approved):

tbhallett:
@marghe-molaro --- I've done some refactoring in here that should have no material effect on the computation.

Please could you re-run [your calcs ](https://github.com/UCL/TLOmodel/pull/1219#issuecomment-1931960751)and see if you get same answers?

marghe-molaro:
@tbhallett 

They look mostly identical, I just noticed draw 7, mean value has a small difference

life expectancy estimates 2024-02-07
life expectancy estimates 2024-02-08

PR #1222, opened 2023-12-08
Include resource files in documentation

matt-graham:
In the quarterly meeting in York in October, it was mentioned it would be useful to have the resource files shown on the documentation site at https://www.tlomodel.org

This PR adds a script `docs/tlo_resources.py` which creates a `docs/resources` subdirectory with reStructuredText files for each of the resource files. For CSV and XLSX files, the data is shown as a table / tables on the page (unless the file is larger than a threshold, 1MiB by default) along with a link to download the original file from GitHub. For files with other extensions / too large files / non UTF-8 encoded files (there is one CSV file with a different encoding) a placeholder page is instead created with just a download link.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,6 +44,7 @@ Contents
    readme
    azure_batch
    reference/index
+   resources/index
    background
    publications
    contributors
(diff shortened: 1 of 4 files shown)


tbhallett:
Brilliant. Thanks very much for this Matt

tbhallett (approved):

tbhallett (approved):

PR #1231, opened 2023-12-12
Separate out malaria test in to two test functions to avoid reinitialising simulation

### LINKED ISSUE #1230: `test_dx_algorithm_for_malaria_outcomes` calls `simulate` on same `Simulation` instance twice
The test function

https://github.com/UCL/TLOmodel/blob/7984913f075d8f96a670bc70b54cb21319f35e58/tests/test_malaria.py#L193-L287

in `tests/test_malaria.py` calls `Simulation.simulate` twice on the same `Simulation` instance. This is causing a failing test in #1227 which adds an exception which is raised if the simulation is initialised more than once. It looks like this is due to the changes I made in #449 from the Git blame 🙈 as I factored out a previous function for initialising simulation in to a fixture which is then shared across the test invocation.

matt-graham:
Fixes #1230 

Separates out the previous `test_dx_algorithm_for_malaria_outcomes` into two separate test functions `test_dx_algorithm_for_malaria_outcomes_clinical` and `test_dx_algorithm_for_malaria_outcomes_severe`, with the previous nested function `make_blank_simulation` now moved outside of the function definition and renamed to `_setup_simulation_for_dx_algorithm_test` so it can be used in both tests. Separating out the test functions ensures the `sim` simulation fixture only has `simulate` method called once per test, and as far as I could see the tests for the clinical and severe cases were not dependent on each other.

### DIFF
--- a/tests/test_malaria.py
+++ b/tests/test_malaria.py
@@ -190,43 +190,44 @@ def test_schedule_rdt_for_all(sim):
     assert sum(df["ma_clinical_counter"]) > 0
 
 
-def test_dx_algorithm_for_malaria_outcomes(sim):
-    """Create a person and check if the functions in dx_algorithm_child return the correct diagnosis"""
+def _setup_simulation_for_dx_algorithm_test(sim):
+    popsize = 200  # smallest population size that works
 
-    def make_blank_simulation():
-        popsize = 200  # smallest population size that works
+    sim.make_initial_population(n=popsize)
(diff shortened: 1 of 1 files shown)


tests/test_malaria.py
tdm32, line 1: I think we can remove self.ALERT_OTHER_DISEASES = [] as this was left over from earlier HSI versions

tdm32 (approved):
Thanks Matt - all looks good

Issue #1123, opened 2023-09-20
Conditionally run `set-variables` job in `run-profiling.yaml` workflow file


matt-graham:
At the moment the `set-variables` job in `.github/workflows/run-profiling.yaml` runs unconditionally:

https://github.com/UCL/TLOmodel/blob/b9e157f9562c2fdcd0115bb01c076d8a759d432c/.github/workflows/run-profiling.yaml#L31-L46

which means even if the subsequent steps are skipped we get a workflow success status rather than skipped status, which for example stops us using a query like https://github.com/UCL/TLOmodel/actions/workflows/run-profiling.yaml?query=is%3Asuccess to find successful runs of the workflow (or at least ones in which any profiling was performed!). 

I think if we add an `if` condition something like

```yaml
if: (github.event_name != 'issue_comment') || ((github.event_name == 'issue_comment') && (github.event.comment.body == '/run profiling')
```

we should prevent the job being run when the succeeding jobs will be skipped, though then we do end up needing to keep the condition here in sync with the logic in the next jobs.

willGraham01:
I feel like the `if` condition is the lesser of two evils - the alternative (which I think we tried when first setting up the profiling) is to put all of the logic in the `run-profiling` workflow into the `command` argument to the comment-triggered stuff, which wasn't the most readable.

PR #1236, opened 2023-12-15
Minor profiling output improvements

matt-graham:
Some minor improvements to the profiling output

- Set `pyinstrument` renderers to explicitly show `tlo` frames and hide (internals of) `pandas` frames. Currently frames corresponding to all functions outside top-level script are hidden by default and when hidden frames are expanded it's hard to differentiate between `tlo` and lower level functions, particularly `pandas`.
- Add option for generating flat HTML output, showing total time spent in each function, ordered from longest to shortest, without showing call stack structure. This can be useful for getting a sense of which functions are dominating overall (while it still being useful to see the call breakdown in the standard HTML output). Currently this shows total time spent in function excluding any child frames - I've opened a PR on `pyinstrument` repository (joerick/pyinstrument/pull/285) to add an option to instead create a flat profile output using total time including child function calls (as I think this is more useful in our case). If that goes in will update this to use that option so leaving this as draft for now.


src/scripts/profiling/run_profiling.py
```diff
@@ -268,13 +279,31 @@ def run_profiling(
...
 
     if write_pyisession:
         output_ipysession_file = output_dir / f"{output_name}.pyisession"
         print(f"Writing {output_ipysession_file}", end="...", flush=True)
         scale_run_session.save(output_ipysession_file)
         print("done")
+        
+    if write_flat_html:
+        if not ANSI2HTML_AVAILABLE:
+            raise ValueError("ansi2html required for flat HTML output.")
```
willGraham01, line 292:
```suggestion
            print("ansi2html required for flat HTML output - SKIPPING.")
```

For robust-ness (and not to waste the ~8 hours or so a profiling session takes) I've tended to not throw errors when something is unavailable, so I'd prefer to just print a message to STDOUT here and then skip the rest of the block.

That way we don't interrupt the workflow, and the usual profiling output will be pushed to the profiling repo in the workflow step that follows. If we throw an error here, the profiling workflow will stop before uploading the results that have already been rendered to the profiling site.

willGraham01 (commented):
Just wanted to flag up that if we throw an error at the end, we won't get any profiling outputs since python error causes the GH workflow to stop before it gets to the "upload profiling results" step.

src/scripts/profiling/run_profiling.py
matt-graham: That's a very good point - probably it would make sense to either just print a message as you say or fail early by checking `ansi2html` is available at the beginning of the function and raising an error there. I have a slight preference for latter as I think failing loudly is generally preferably to a message which might get missed. Will update accordingly!

matt-graham:
I've now moved error on requesting flat HTML output when `ansi2html` not available to beginning of function to fail early and we've also now merged in Will's changes from #1301 which will fix #1123 and fix #1300.

Re-requested a review @willGraham01 in case you have any further things you think need changing otherwise we can merge this in if tests pass.

willGraham01 (approved):
> merged in Will's changes from https://github.com/UCL/TLOmodel/pull/1301 which will fix https://github.com/UCL/TLOmodel/issues/1123 and fix https://github.com/UCL/TLOmodel/issues/1300

:crossed_fingers: :sweat_smile: 

All good by me - will pre-emptively approve.

matt-graham:
I've tagged on a commit to disable the progress bar output in the GitHub Actions profiling runs as requested by @tamuri in https://github.com/UCL/TLOmodel/issues/1277#issuecomment-1956215953 as this produced a lot of noise in the recorded log output that makes it slow to load. I added a `--disable-output-to-stdout` flag to both to `scale_run.py` and `run_profiling.py` scripts which also allows disabling log output to `stdout` (as this also generates a lot of output for long profiled runs and is recorded to the log file separately anyway).

Issue #1244, opened 2023-12-20
Outstanding tasks on refresh of the calibration analyses [aka dev] server


tamuri:
- [ ] write file `info.txt` in commit output directory (e.g. `/mnt/tlodev2stg/tlo-dev-fs-2/task-runner/output/2023-11-24_152805_d06d3d4`) with two lines:
	- [ ] line 1: a git log message with format: `git log --pretty=format:"%H%x09%ad%x09%s" --date=iso-strict ${commit_sha1}`
	- [ ] line 2: URL to action run page on GitHub e.g. `https://github.com/UCL/TLOmodel/actions/runs/7256652901`
- [ ] gzip `*.log` files from all runs at end of post-processing step
	- [ ] check with @tbhallett if we can get rid of (i) module-specific log files (`tlo.*.log`) (ii) pickled module dataframes (`tlo.*.pickle`)
- [ ] setup webserver on new vm (tlo-dev-vm-1) & make results publically accessible
	- [ ] use Caddy webserver - write Caddyfile
	- [ ] Ansible playbook to provision server with Caddy
	- [ ] open inbound ports 80, 443 on NIC (via Azure)
	- [ ] setup DNS record (e.g. analyses.tlomodel.org) - speak to TimH re. GoDaddy record

tamuri:
The info file is being written in calibration runs whether or not it's a new commit that's running. If it's a commit that's already run, the workfow writes over the original file, meaning the URL to workflow run points to the run that is skipped. (Hope that made sense!)

PR #1245, opened 2023-12-21
[calibration] Store extra information in `info.txt` file

giordano:
Address first point of #1244

### DIFF
--- a/.github/workflows/calibration.yaml
+++ b/.github/workflows/calibration.yaml
@@ -119,6 +119,15 @@ jobs:
           echo "tasks=${RUNS}"
           echo "tasks=${RUNS}" >> "${GITHUB_OUTPUT}"
 
+      - name: Write info.txt file
+        run: |
+          mkdir -p "${output_dir}"
+          info_txt="${output_dir}/info.txt"
+          git -C "${WORKTREE_PATH}" log --pretty=format:"%H%x09%ad%x09%s" --date=iso-strict -n 1 > "${info_txt}"
+          echo -e "\n${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" >> "${info_txt}"
+          echo "Content of ${info_txt}:"
+          cat "${info_txt}"
+
   # Run the tasks.
   tasks:
     needs: setup


.github/workflows/calibration.yaml
```diff
@@ -163,6 +163,13 @@ jobs:
           mkdir -p "${task_output_dir}"
 
           python3 "${{ needs.setup.outputs.worktree_path }}/src/scripts/calibration_analyses/analysis_scripts/process.py" "${task_output_dir}" "${RUN_NAME}" "${{ needs.setup.outputs.worktree_path }}/resources"
+
+          # Print info.txt file with some useful information
+          info_txt="${task_output_dir}/info.txt"
+          git -C working-directory log --pretty=format:"%H%x09%ad%x09%s" --date=iso-strict -n 1 > "${info_txt}"
+          echo "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" >> "${info_txt}"
```
giordano, line 170: Note: logs are automatically deleted after ~3 months or so, so this line can be useful for reference for some time, but not forever.

.github/workflows/calibration.yaml
```diff
@@ -163,6 +163,13 @@ jobs:
           mkdir -p "${task_output_dir}"
 
           python3 "${{ needs.setup.outputs.worktree_path }}/src/scripts/calibration_analyses/analysis_scripts/process.py" "${task_output_dir}" "${RUN_NAME}" "${{ needs.setup.outputs.worktree_path }}/resources"
+
+          # Print info.txt file with some useful information
+          info_txt="${task_output_dir}/info.txt"
```
tamuri, line 168: I think this would put the file in the `022_long_run_all_diseases_process` directory. Can we put it in the parent (I guess `${{ needs.setup.outputs.output_dir }`)?

.github/workflows/calibration.yaml
tamuri: I think that's okay - it's for an easy way to get to see stdout when something goes wrong.

.github/workflows/calibration.yaml
giordano: You're right :facepalm: 
```suggestion
          info_txt="${{ needs.setup.outputs.output_dir }}/info.txt"
```

giordano:
I was wondering, if the idea was to use this file in case something goes wrong, maybe we should create it early on, in the setup step, instead of at the end of postprocessing, where it's too late if anything failed before?

tamuri:
Good point - you're right. With the info.txt file we could show something on the generated webpage, enough to indicate "we tried to run analyses on this commit, and it failed".

giordano:
Ok, I moved the writing of the file to its own step in the setup job, and addressed all other silly mistakes I made yesterday.  Hopefully now it should be all good :smiling_face_with_tear:

tamuri (approved):

PR #1246, opened 2023-12-22
[calibration] Compress log files at the end of postprocessing job

giordano:
Second point of #1244


.github/workflows/calibration.yaml
```diff
@@ -175,6 +175,10 @@ jobs:
           python3 "${{ needs.setup.outputs.worktree_path }}/src/scripts/task_runner/generate_html.py" "${OUTPUT_ROOT}" > "${OUTPUT_ROOT}/index.html"
         working-directory: "${{ needs.setup.outputs.worktree_path }}"
 
+      - name: Compress log files
+        run: |
+          find "${{ needs.setup.outputs.output_dir }}" -name '*.log' -exec gzip --best --force --verbose '{}' \;
```
giordano, line 180: I was unsure about the `--force` option (hopefully we don't need to force anything!), but it's in unlikely case the job gets stuck forever asking for input.

tamuri (approved):

PR #1252, opened 2024-01-04
Update publicationa as of January 2024

tbhallett:
Adding some recent publications, and pre-prints where we have them.

### DIFF
--- a/docs/publications.rst
+++ b/docs/publications.rst
@@ -5,10 +5,15 @@ Publications
 
 These are the publications that have been generated either in the course of the model's development or its application.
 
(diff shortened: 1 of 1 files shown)


marghe-molaro:
Hi @tbhallett, what about Bingling's Covid impact paper? Is that not available on pre-print?

tbhallett:
> Hi @tbhallett, what about Bingling's Covid impact paper? Is that not available on pre-print?

I don't see it, and @BinglingICL's comment on TimC's paper says that it's "in production" with PLOS One still..

marghe-molaro (approved):

Issue #1277, opened 2024-02-16
Scheduled profiling runs from GitHub Actions `run-profiling.yaml` workflow intermittently failing


matt-graham:
The scheduled runs of the `run-profiling.yaml` workflow are [intermittently failing](https://github.com/UCL/TLOmodel/actions/workflows/run-profiling.yaml?query=event%3Aschedule). 

In most cases this appears to be due to an out of memory issue (process exits with error code 247 which some searching suggests is related to container running out of memory) for example

https://github.com/UCL/TLOmodel/actions/runs/7304903455/job/19907772489#step:4:16223

This seems to happen after profiling run has completed during the writing of HTML output in the lines

https://github.com/UCL/TLOmodel/blob/0914625ca6a985f44108ce553be943ad742a6d9e/src/scripts/profiling/run_profiling.py#L257-L260

as we get the `"Writing {output_html_file}"` message printed but not the corresponding `"done"`. This is probably therefore the `pyinstrument` HTML renderer taking up too much memory for the large profiles generated. Simplest option to fix would be to just reduce the simulation length (currently 5 years), population size (currently 50000) or frame recording interval (currently 0.1s). Alternatively @tamuri do any of the other self-hosted runners have a larger memory (profiling is currently being run on runners with `test` tag, not sure if those with `tasks` tag have more resources and/or if it would make sense to use these instead?). 

In the most [recent run a different error occured](https://github.com/UCL/TLOmodel/actions/runs/7851093893/job/21427635584)

```
 [00:26:04:INFO] Starting profiling runs
Traceback (most recent call last):
  File "/runner/_work/TLOmodel/TLOmodel/src/scripts/profiling/run_profiling.py", line 389, in <module>
    run_profiling(**vars(args))
  File "/runner/_work/TLOmodel/TLOmodel/src/scripts/profiling/run_profiling.py", line 232, in run_profiling
    completed_simulation = scale_run(
                           ^^^^^^^^^^
  File "/runner/_work/TLOmodel/TLOmodel/src/scripts/profiling/scale_run.py", line 83, in scale_run
    sim.register(
  File "/runner/_work/TLOmodel/TLOmodel/.tox/profile/lib/python3.11/site-packages/tlo/simulation.py", line 168, in register
    module.read_parameters('')
  File "/runner/_work/TLOmodel/TLOmodel/.tox/profile/lib/python3.11/site-packages/tlo/methods/healthsystem.py", line 776, in read_parameters
[...]

PR #1278, opened 2024-02-16
Enable LFS in profiling workflow checkout + reduce frame recording interval

matt-graham:
Possible fixes for #1277 though not linking to that issue as will need to test workflow does reliably run after these changes have been made to be sure.

Adds `lfs: true` option in `actions/checkout` Action in `run-profiling` workflow to ensure LFS files are checked out correctly (necessary for profiling as this includes resource files we need for running simulation).

Also reduces the default interval at which frames are recorded during profiling from 0.1s to 0.2s with the hope this will reduce memory requirements when rendering profiling record to HTML which currently seems to be exceeding memory available on runners.

### DIFF
--- a/.github/workflows/run-profiling.yaml
+++ b/.github/workflows/run-profiling.yaml
@@ -91,6 +91,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ github.sha }}
+          lfs: true
 
       ## The profile environment produces outputs in the /results directory
       - name: Run profiling in dev environment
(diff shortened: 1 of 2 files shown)


willGraham01 (approved):
I can approve but not merge :sweat_smile: 

One thing we might want to consider is just going to 1s frame-capture intervals (maybe not now but in future). If the simulations are running for nearly 6 hours, would we really loose that much information by going from 0.1s to 1s? And this reduces memory usage 10-fold too (theoretically)

PR #1280, opened 2024-02-22
Include task-shifting

marghe-molaro:
Initial task-shifting implementation, currently assuming that the health-system adopts a "global policy" on task shifting. We will potentially transition to treatment-specific after health-econ discussion on 8th March.

PR includes test (test_task_shifting_in_mode_2)

Other notes:
- We assume ranking in preferred option for task-shifting (e.g. first choice for task-shifting pharmacists are nurses, followed by clinicians);
- We assume that task-shifting occurs immediately when appt is extracted from the queue (i.e. we don't wait until the end of the day to check if we could have performed extra appts by allowing task-shifting). This seems consistent with the fact that appts are ranked by priority: earlier appts in the queue are higher priority, and therefore the health system does everything possible, including task-shifting, to complete them first.
- We assume that tasks performed by different medical officers don't have to be performed simultaneously, i.e. if appt requires both nurse and pharmacist, nurse will be able to perform own task *and* that of pharmacist if the latter is no longer available.
- Currently we have no way of logging occurrences of task-shifting (although of course in mode 2 it will impact the number of appts performed) because we don't have appt_footprints that reflect the task shifting. We can discuss whether we want to address this or not.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -568,7 +568,7 @@ class HealthSystem(Module):
             Types.STRING, "Name of global task-shifting mode adopted"),
         
         'global_task_shifting': Parameter(
-            Types.DICT, "Global task-shifting policy adopted by the health care system. It includes a list of the officers"
+            Types.DICT, " Global task-shifting policy adopted by the health care system. It includes a list of the officers"
                         " eligible for task shifting. For each of those officers, it specifies which officers"
                         " can take over the officer's tasks, and a factor explaining how the originally scheduled time "
                         " should be scaled if performed by a different officer."),
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -1,10 +1,11 @@
 import datetime
 import heapq as hp
 import itertools
+import re
 import warnings
 from collections import Counter, defaultdict
 from collections.abc import Iterable
-from itertools import repeat
+from itertools import combinations, product, repeat
 from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Tuple, Union
 
@@ -635,8 +636,6 @@ class HealthSystem(Module):
             and priority
         :param ignore_priority: If ``True`` do not use the priority information in HSI
             event to schedule
-        :param include_task_shifting: If ``True`` when in mode 2 consider task-shifting of officers if one originally
-            required is not available
         :param policy_name: Name of priority policy that will be adopted if any
         :param capabilities_coefficient: Multiplier for the capabilities of health
             officers, if ``None`` set to ratio of initial population to estimated 2010
@@ -891,7 +890,7 @@ class HealthSystem(Module):
         self.setup_priority_policy()
         
         # Set up framework for considering a global task-shifting policy
-        self.setup_global_task_shifting(self.parameters['global_task_shifting_mode'])
+        self.setup_global_task_shifting()
 
 
     def initialise_population(self, population):
@@ -994,7 +993,7 @@ class HealthSystem(Module):
         if 'Tb' in self.sim.modules:
             self.list_fasttrack.append(('tb_diagnosed', 'FT_if_tbdiagnosed'))
             
-    def setup_global_task_shifting(self,mode):
+    def setup_global_task_shifting(self):
     
         # Select the global task shifting mode to be considered
         self.global_task_shifting_mode = self.get_global_task_shifting_mode_initial()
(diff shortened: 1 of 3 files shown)


tdm32:
@marghe-molaro we maybe do want some measure of the amount of task-shifting which occurs during a simulation so we could quantify for example, what would we gain if nurses were allowed to spend 25% of their time on pharmacy tasks? Or conversely, if global task-shifting is permitted, how much nurses time would be required to fill the gaps through pharmacists being under-resourced? We could do this through counterfactuals of course, but having direct measures is always useful as counterfactuals quickly diverge making run-by-run comparisons tricky.

tdm32:
@marghe-molaro also noting here the important role that DCSAs may play when we consider task-shifting - they are already doing MDA and child health days for example.

src/tlo/methods/healthsystem.py
```diff
@@ -2445,17 +2462,58 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
+                                    
+                                    # Extract task-shifting options for this officer
+                                    task_shift_options_for_officer = self.sim.modules['HealthSystem'].global_task_shifting.get(officer_no_facility)
+
+                                    # Check if possible alternatives to officer are available
+                                    # If they are, must register that we'll be using alternative.
+                                    if task_shift_options_for_officer:
+                                        
+                                        # Unpack values
+                                        new_officer_no_facility, time_scaling = task_shift_options_for_officer
```
tbhallett, line 2497: why unpack only to zip in the following line?  I think we can skip this line.

src/tlo/methods/healthsystem.py
```diff
@@ -2445,17 +2462,58 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
+                                            shift_target_officer = officer.replace(officer_no_facility, new_officer_no_facility)
+                                            
+                                            # Check if this task-shifting officer is available, and save task_shift
+                                            if shift_target_officer in set_capabilities_still_available:
+                                                # Record task-shifting
+                                                task_shifting_adopted[officer] = (shift_target_officer, time_scaling)
+                                                out_of_resources = False
+                                                
+                                                # Once we've found available officer to replace, no need to go through other
+                                                # options
```
tbhallett, line 2510:
```suggestion
                                                # options for other officers to task-shift to.
```

src/tlo/methods/healthsystem.py
```diff
@@ -2445,17 +2462,58 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
                         squeeze_factor = 0.
 
-                        # Check if any of the officers required have run out.
+                        # Check if any of the officers required have run out. If including task-shifting,
+                        # this will involve checking if alternative capabilities are available for officers
+                        # that are no longer available.
                         out_of_resources = False
+                        
+                        # This dictionary stores all task-shifting officers considered for this appointment
+                        task_shifting_adopted = {}
```
tbhallett, line 2471: this dict isn't logged or used. 

To allow logging, we _could_ save a flag onto the HSI Event itself (e.g. `.global_task_shifting_used: bool`) and then use this property when do the logging (summary and detail).

src/tlo/methods/healthsystem.py
```diff
@@ -2445,17 +2462,58 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
+                                        new_officer_no_facility, time_scaling = task_shift_options_for_officer
+                                        
+                                        for new_officer_no_facility, time_scaling in zip(new_officer_no_facility, time_scaling):
+                                            # Get the task-shifting officer
+                                            shift_target_officer = officer.replace(officer_no_facility, new_officer_no_facility)
+                                            
+                                            # Check if this task-shifting officer is available, and save task_shift
+                                            if shift_target_officer in set_capabilities_still_available:
+                                                # Record task-shifting
+                                                task_shifting_adopted[officer] = (shift_target_officer, time_scaling)
```
tbhallett, line 2506: As noted above, this dict of record isn't used anywhere.

src/tlo/methods/healthsystem.py
```diff
@@ -2499,6 +2557,7 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
                                 f"Cannot run HSI {event.TREATMENT_ID} without facility_info being defined."
 
                             # Expected appt footprint before running event
+                            # NOTE-TO-ADD: This appt footprint needs to reflect that a different officer was used
```
tbhallett, line 2560: this could be accomplished by saving the flag onto the HSI Event as per above?

src/tlo/methods/healthsystem.py
```diff
@@ -2521,7 +2580,40 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
+                                    else:
+                                        if k in updated_call_inc_task_shift.keys():
+                                            # Ensure that if this officer already present in dictionary
+                                            # this task is added, not overwritten
+                                            updated_call_inc_task_shift[k] += updated_call[k]
+                                        else:
+                                            updated_call_inc_task_shift[k] = updated_call[k]
+
+                                updated_call = updated_call_inc_task_shift
+
```
tbhallett, line 2616: As we already have this type of logic (when the ACTUAL_APPT_FOOTPRINT is different to the EXPECTED_APPT_FOOTPRINT), I wonder if this is duplicative.

Could we instead let the HSI event return an ACTUAL_APPT_FOOTPRINT that reflects the task-shifting, so that we re-use that logic?

src/tlo/methods/healthsystem.py
```diff
@@ -2543,13 +2635,16 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
                             self.module.running_total_footprint += updated_call
 
                             # Write to the log
+                            # WARNING: the logged appt footprint does not contain information
+                            # on whether task-shifting was performed or not.
```
tbhallett, line 2639: indeed, this would be another reason to let the HSI return actual footprint that its provided with

tests/test_healthsystem.py
```diff
@@ -1952,6 +1956,128 @@ def apply(self, person_id, squeeze_factor):
...
+
+    # read the results
+    output = parse_log_file(sim.log_filepath, level=logging.DEBUG)
+    hs_output = output['tlo.methods.healthsystem']['HSI_Event']
+
+    # Check that all events could run, even if Pharmacy capabilities were set to zero,
+    # and that when task-shifting first option (nurses) where always preferentially chosen over second
+    # one (clinicians)
+    assert hs_output['did_run'].sum() == Ntarget
+
```
tbhallett, line 2079: To strengthen the test and see the impact of task-shifting versus no task-shifting, I think it would be good to repeat this for with/without global task-shifting so we can see the difference.

tbhallett (changes requested):
Thanks so much for this, @marghe-molaro 

This is all very nice and seems to work well.

My suggestions boil-down to three things:

1) In the test, I think we should repeat that experiment with/without global task shifting and see the difference.

2) Interface-wise, I think let this only and fully controlled by a csv ResourceFile describing the global task shifting we want (If it's blank it means we don't want any). And let's not have the key-word argument thing, as it's not necessary.

3) Logic and logging-wise: I thought that we could perhaps store onto the HSI Event itself the 'task-shifted' appointment and then manufacture that this is returned from the HSI_Event (as the 'ACTUAL_APPT_FOOTPRINT`). That way the updating of capabilities and the logging of what actual is used all happens automatically. In addition we could have a flag that this is a "task shifted" appointment, which would also flow into the logs.

tests/test_healthsystem.py
```diff
@@ -1952,6 +1956,128 @@ def apply(self, person_id, squeeze_factor):
...
+    # Get pointer to the HealthSystemScheduler event
+    healthsystemscheduler = sim.modules['HealthSystem'].healthsystemscheduler
+
+    # Force entire population in one district (keys_district[0]) and get facility ID
+    person_for_district = {d: i for i, d in enumerate(sim.population.props['district_of_residence'].cat.categories)}
+    keys_district = list(person_for_district.keys())
+
+    for i in range(0, int(tot_population)):
+        sim.population.props.at[i, 'district_of_residence'] = keys_district[0]
```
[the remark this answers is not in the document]
marghe-molaro: No they wouldn't be all for the same person, each person gets a separate appointment.

src/tlo/methods/healthsystem.py
marghe-molaro: See reply to comment above

src/tlo/methods/healthsystem.py
```diff
@@ -2827,6 +2922,7 @@ class HealthSystemChangeParameters(Event, PopulationScopeEventMixin):
     """Event that causes certain internal parameters of the HealthSystem to be changed; specifically:
         * `mode_appt_constraints`
         * `ignore_priority`
+        * `include_task_shifting`
```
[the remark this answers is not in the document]
marghe-molaro: yes you are right, the reason this wasn't optimised at this stage is that we're still not sure whether we wanted to allow for a global *and* HSI-specific task-shifting approach (in which case e.g. include_task_shifting could be categorical - None, Global-level, HSI-level and we would have an additional resource file in case of global taskshifting, one for HSI-based on etc)

sakshimohan (approved):
Thanks for adding me to this @marghe-molaro. I'm unable to offer comments on the code, but the approach sounds right to me.

src/tlo/methods/healthsystem.py
marghe-molaro: It is adopted, and it must be used to know which officers were assigned what tasks (line 2506)

marghe-molaro:
Many thanks @tdm32 and @tbhallett for your comments!

Tara's points:
- I take your point about the need to log the amount of task-shifting - how we do it will depend on whether we go down the "global" or "HSI-specific" task-shifting I think. Notice however that we currently don't cap the amount of task-shifting that specific cadres can do (25% in your example); if we wanted to include that it would have to be implemented quite differently I think. 
- The current task-shifting set up was only an example, I take your point about including DCSAs.

Tim's points:
1. No problem about adding the alternative scenario in the test;
2.-3. I think these points will depend on how the specifics of the scenarios we want to implement, in particular whether we opt to go for a global approach to the task-shifting or make it HSI-specific. I think for this reason it might be best to wait until the discussion at the Health Econ meeting next friday to decide how we answer these.

tests/test_healthsystem.py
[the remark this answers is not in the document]
tbhallett: Sorry. You're quite right.

src/tlo/methods/healthsystem.py
tbhallett: I see it being written to but not read from (?)

src/tlo/methods/healthsystem.py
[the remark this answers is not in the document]
tbhallett: I think it's always the case that we should modify the behaviour of the module through Parametees rather than kwargs though, as that it has the Scenario class expects to pass through changes.

src/tlo/methods/healthsystem.py
marghe-molaro: Lines 2593 to 2596

marghe-molaro:
Hi @tbhallett and @tdm32, 

(Note: Tests are failing on github but running fine locally.)

Following the discussion on Wed, I have now made the following modifications to this PR:

_appt_times (which links appt_footprints to the officers&times required to perform the appointment) is now inflated in pre-processing to include new appt_footprints, which account for all possible task-shifting that may take place given the policy specified by global_task_shifting.
The name of the new appt_footprints is the original appt_footprint followed by "_withTS_xxxx-yyyy_and_xxxx-yyyy..." where xxxx are the first four letters of the officer which will be task-shifted, and yyyy are the first four letters of the officer taking over the tasks. E.g. "Under5OPD_withTS_Phar-Nurs_and_Clin-DCSA" is an Under5OPD  appts where the tasks originally assigned to a Pharmacist are performed by a Nurse, and where tasks assigned to a Clinician are performed by a DCSA instead. Times required are updated to reflect TS taking place.

At the time of running HSIs in mode 2, if task-shifting is required a tag "_withTS_..." with all relevant task-shifting info will be added to the actual_appt_footprint returned by the HSI, such that times requested from each officer can be quickly looked-up from the expanded _appt_times (and hence subtracted from daily capabilities) rather than having to be recalculated for each HSI at run time. In doing so the TS can be logged as having occurred with details of how that took place directly as part of the logged appointment footprint.

It is difficult to say whether this approach will be faster than the one we had implemented before, as this will very much depend on how much TS actually takes place. It shouldn't impact runs where TS is not included. This approach however allows us to easily log TS as having taken place in all relevant detail.

marghe-molaro:
@tbhallett following today's Health Econ discussion I would suggest pausing review of this PR until we have a clearer idea of how we want to enforce task-shifting, in particular if we want this to be more HSI-driven.

tbhallett:
Found this ResourceFile on another approach (#1296) but not here.
I removed it from #1296 during review 
[ResourceFile_GlobalTaskShifting.xlsx](https://github.com/UCL/TLOmodel/files/14654330/ResourceFile_GlobalTaskShifting.xlsx)

PR #1292, opened 2024-03-06
Avoid DataFrame accesses and standardise HSI scheduling interactions during first appointments

### LINKED ISSUE #407: Factorise and simplify the GenericFirstAppt* events
Inspired by #402, @matt-graham  suggests: 

"would it possibly make sense to instead have each Module subclass define a pair of methods `do_at_generic_first_appt_non_emergency` and `do_at_generic_first_appt_emergency` with signatures something like
```python
    def do_at_generic_first_appt_non_emergency(self, generic_hsi_event, person, symptoms, squeeze_factor):
        ...

    def do_at_generic_first_appt_emergency(self, generic_hsi_event, person, symptoms, squeeze_factor):
        ...
```

which implement the logic associated with scheduling the specific HSI events associated with each Module subclass? The `HSI_GenericFirstApptAtFacilityLevel0.apply` method could then be defined as something like:

```python
def apply(self, person_id, squeeze_factor):

    person = df.loc[person_id]

    if not person.is_alive:
        return

    symptoms = hsi_event.sim.modules['SymptomManager'].has_what(person_id=person_id)

    for module in self.sim.modules.values():
        module.do_at_generic_first_appt_non_emergency(self, person, symptoms, squeeze_factor)
```

or equivalently using `module.do_at_generic_first_appt_emergency` for `HSI_GenericEmergencyFirstApptAtFacilityLevel1.apply`.

This would allow keeping the disease specific logic within the corresponding modules and avoid needing to keep adding to the hsi_generic_first_appts module as new disease modules are added.

For example the implementation for the Malaria module could be something like
```python
def do_at_generic_first_appt_non_emergency(self, generic_hsi_event, person, symptoms, squeeze_factor):
    person_id = generic_hsi_event.target
    test_result_to_treatment_event_map = {
        "severe_malaria": 
        HSI_Malaria_complicated_treatment_child if person.age < 15 else
        HSI_Malaria_complicated_treatment_adult,
        "clinical_malaria": 
        HSI_Malaria_non_complicated_treatment_age0_5 if person.age < 5 else
        HSI_Malaria_non_complicated_treatment_age5_15 if person.age < 15 else
        HSI_Malaria_non_complicated_treatment_adult
    }
    if 'fever' in symptoms:
        test_result = self.check_if_fever_is_caused_by_malaria(
            person_id=person_id, hsi_event=generic_hsi_event
        )
        # Treat / refer based on diagnosis
        treatment_event = test_result_to_treatment_event_map.get(test_result)
        if treatment_event is not None:
            self.sim.modules['HealthSystem'].schedule_hsi_event(
                treatment_event(person_id=person_id, module=self),
                priority=1,
                topen=self.sim.date,
                tclose=None
            )

def do_at_generic_first_appt_emergency(self, generic_hsi_event, person, symptoms, squeeze_factor):
    person_id = generic_hsi_event.target
    # Quick diagnosis algorithm - just perfectly recognises the symptoms of severe malaria
    sev_set = {
        "acidosis", "coma_convulsions", "renal_failure", "shock", "jaundice", "anaemia"
    }

    # if person's symptoms are on severe malaria list then consider treatment for malaria
    any_symptoms_indicative_of_severe_malaria = len(sev_set.intersection(symptoms)) > 0
    if any_symptoms_indicative_of_severe_malaria:
        # Check if malaria parasitaemia:
        test_result = self.check_if_fever_is_caused_by_malaria(
            person_id=person_id, hsi_event=generic_hsi_event
        )
        # if any symptoms indicative of malaria and they have parasitaemia (would return a positive rdt)
        if test_result in ("severe_malaria", "clinical_malaria"):
            # Launch the HSI for treatment for Malaria - choosing the right one for adults/children
            treatment_event = (
                HSI_Malaria_complicated_treatment_child if person.age_years < 5 else 
                HSI_Malaria_complicated_treatment_adult
            )
            self.sim.modules['HealthSystem'].schedule_hsi_event(
                hsi_event=treatment_event(self, person_id=person_id),
                priority=0,
                topen=self.sim.date,
            )
```

### LINKED ISSUE #1237: A lot of time spent in `do_at_generic_first_appt_non_emergency` due to repeated dataframe access
In the [latest scheduled profiling run](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_254_7984913f075d8f96a670bc70b54cb21319f35e58.html), 21.4% of the overall run time was spent in calls to `do_at_generic_first_appt_non_emergency`. Looking at the breakdown of time spent inside the function, most of the time is spent in evauating `_LocIndexer.__getitem__` (that is performing `loc` indexing operations on the population dataframe), both directly and as part of `SymptomManager.has_what` calls. 

My understanding of this function is that it acts as an initial trigger for the scheduling of specific HSI events connected with different disease modules, but beyond scheduling further HSI events, should not immediately make any changes to the properties of the individual with the associated `person_id` ? @tbhallett can I check this is correct.

If this is the case, then we can avoid the repeated dataframe access by doing a single initial read of the individuals properties at the beginning of the function and then passing a datastructure containing these values to all called functions. Ideally we would make this datastructure immutable - for example created a `namedtuple` to be explicit that the properties are read only and cannot be updated. Given a lot of the dataframe accesses are part of `SymptomManager.has_what` calls, we would need to refactor this / provide an alternative method to allowing passing the properties directly (as for example a named tuple), and potentially also evaluate the symptoms the individual has once and pass in explicitly to called functions to avoid overhead in reconstructing the symptom list multiple times.

If we do require allowing direct updates to individual properties as part of these calls, and alternative to using an immutable data structure, would be to create a class which will allows changing property values, but records which property are changed and writes these back to the population dataframe at the end of the call.

A possibly nice way to do all of this and also simplify the structure of `do_at_generic_first_appt_non_emergency` (and analagously `do_at_generic_first_appt_emergency`) would be to add methods (defaulting to a no-op) to the base `Module` class which deal with scheduling the necessary follow-on HSI events  from a generic first [non-emergency | emergency] appointment for each disease module, something like

```Python

class Module:
    
    ...

    def do_at_generic_first_appt_non_emergency(self, person_id: int, person_properties: namedtuple, symptoms: list[str]):
        """Schedule any HSI events following on from a non-emergency generic first appointment."""
        ...


    def do_at_generic_first_appt_emergency(self, person_id: int, person_properties: namedtuple, symptoms: list[str]):
        """Schedule any HSI events following on from an emergency generic first appointment."""
        ...
```

with corresponding functions in `hsi_generic_first_appts` module

```Python
def do_at_generic_first_appt_non_emergency(hsi_event, squeeze_factor):
    person_id = hsi_event.target
    person_properties = hsi_event.sim.population.get_individual_properties(person_id)
    symptoms = hsi_event.sim.modules['SymptomManager'].has_what_from_properties(person_properties )
    for module in hsi_event.sim.modules.values():
        module.do_at_generic_first_appt_non_emergency(person_id, person_properties, symptoms)

def do_at_generic_first_appt_emergency(hsi_event, squeeze_factor):
    person_id = hsi_event.target
    person_properties = hsi_event.sim.population.get_individual_properties(person_id)
    symptoms = hsi_event.sim.modules['SymptomManager'].has_what_from_properties(person_properties )
    for module in hsi_event.sim.modules.values():
        module.do_at_generic_first_appt_emergency(person_id, person_properties, symptoms)
```

which avoids the large number of `if` statements and module specific logic in the current implementation. Given it's passed to these function we could also pass through `squeeze_factor` to the modules methods if its thought this will be used at some point in future.

My main concern with this approach would be the difficulty in enforcing the condition that modules `do_at_generic_first_appt_non_emergency` and `do_at_generic_first_appt_emergency` only schedule HSI events and do not edit properties of the individual directly (which would invalidate the approach of reading the properties once at beginning of method). We could just rely on documenting this expectation and adding tests to check it is adhered to. 

Alternatively at the cost of more complex function signatures, we could make `do_at_generic_first_appt_non_emergency` and `do_at_generic_first_appt_emergency` static methods of the modules, without direct access to the module attributes (including the population data frame). In this case we would need to pass through all objects that would be needed for dealing with scheduling following on HSI events as arguments.

willGraham01:
Closes #1237 and closes #407 |

Provides a performance improvement my minimising the number of DataFrame accesses made by the various disease modules when scheduling generic, first-appointment HSIs. Also uses this opportunity to standardise and refactor how this part of the simulation is handled by the large number of objects that are involved.

## Key Changes

- The `do_at_generic_first_appt_non_emergency` and `do_at_generic_first_appt_emergency` are now a single method, `_do_on_generic_first_appt`, that belongs to a new class `HSI_BaseGenericFirstAppt`. This reduces the amount of code duplication in the `hsi_generic_first_appts` module, and avoids un-necessary argument passing between functions (as the objects that are passed to the existing functions _are_ all available as class attributes anyway). The new `HSI_BaseGenericFirstAppt` is then inherited by the existing`HSI_GenericNonEmergencyFirstAppt` and `HSI_GenericEmergencyFirstAppt` classes to maintain compatibility with the rest of the codebase.
- All `Modules` (and classes that inherit from this class) have two stub methods - `do_at_generic_first_appt_{non_}emergency`. These methods determine how first-appointment HSIs are scheduled on a per-module basis. This implementation is now found within the modules themselves, rather than in the `_do_on_generic_first_appt` function where it was previously. As a result, the `_do_on_generic_first_appt` method is now reads in a simpler manner and does not need to consist of a long list of `if` statements performing the per-module behaviours.
- Scheduling of HSI events from first appointments is now the responsibility of the module-level `do_on_generic_first_appt` method(s).
- These methods however **do not** have the power to update the population dataframe, but instead return any updates that they want to implement. These updates are conducted after every module has had an opportunity to schedule the HSI events that they want based on the patient's details upon arrival at a first appointment, to avoid repeated dataframe reads causing slowdowns.


src/tlo/core.py
```diff
@@ -365,3 +369,89 @@ def on_birth(self, mother_id, child_id):
...
+            [
+                Union[None, np.integer, int, List, Set, Dict],
+                Union[None, np.integer, int, List, Set, Dict],
+            ],
+            Union[bool, Dict],
+        ] = None,
+        facility_level: str = None,
+        treatment_id: str = None,
+        random_state: "RandomState" = None,
+    ) -> Tuple[List[Tuple["HSI_Event", Dict[str, Any]]], Dict[str, Any]]:
```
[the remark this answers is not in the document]
matt-graham: This would also simplify defining type hints on overridden instances of this method on `Module` subclasses.

src/tlo/methods/hsi_generic_first_appts.py
```diff
@@ -6,369 +6,179 @@
...
+                consumables_checker=self.get_consumables,
+                facility_level=self.ACCEPTED_FACILITY_LEVEL,
+                treatment_id=self.TREATMENT_ID,
+                random_state=self.module.rng,
+            )
+            # Schedule any requested updates
+            for info in event_info:
+                event = info[0]
+                options = info[1]
+                schedule_hsi(event, **options)
```
matt-graham, line 92: Is there an advantage / is it necessary to do the actual scheduling of the HSI events in this function rather than within the called `do_at_generic_first_appt` / `do_at_generic_first_emergency_appt` methods? It seems like having those methods return a list of HSI events and then schedule here ends up making a more complex return signature than needed, while as if we just had them directly call `schedule_hsi` (either by passing this in as an argument or calling out `HealthSystem.schedule_hsi_event` directly via `Module.sim.modules["HealthSystem"]`) we would avoid some of the boilerplate / complex return semantics we get from doing the scheduling indirectly.

src/tlo/methods/hsi_generic_first_appts.py
```diff
@@ -6,369 +6,179 @@
...
+        super().__init__(module, person_id=person_id)
+        # No footprint, as this HSI (mostly just) determines which
+        # further HSI will be needed for this person. In some cases,
+        # small bits of care are provided (e.g. a diagnosis, or the
+        # provision of inhaler).
+        self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint(
+            {}
+        )
+
+    def _diagnosis_fn(self, tests, use_dict: bool = False, report_tried: bool = False) -> Any:
```
matt-graham, line 39: Could the return type hint be more specific here as docstring suggests it will be a dict of some form?

src/tlo/methods/hsi_generic_first_appts.py
```diff
@@ -6,369 +6,179 @@
...
+        for module in modules.values():
+            event_info, df_updates = getattr(module, self.MODULE_METHOD_ON_APPLY)(
+                patient_id=self.target,
+                patient_details=patient_details,
+                symptoms=symptoms,
+                diagnosis_fn=self._diagnosis_fn,
+                consumables_checker=self.get_consumables,
+                facility_level=self.ACCEPTED_FACILITY_LEVEL,
+                treatment_id=self.TREATMENT_ID,
+                random_state=self.module.rng,
```
matt-graham, line 86: Looks like the random number generator is only used by one module downstream (`Labour`). It might be worth checking either if (i) the random sampling used there is actually required (seems to be randomly assigning facility level) and (ii) if it might not make more sense to use the module specific `RandomState` instance under the `rng` attribute as that is generally the pattern used in other methods of  `Module` instances.

matt-graham (commented):
Submitting some initial thoughts and suggestions as realise you've been left waiting on feedback for a while @willGraham01 and thought it better to give some partial comments now rather than wait on me finishing reviewing all the changes. Will continue looking at this tomorrow afternoon.

src/tlo/analysis/utils.py
```diff
@@ -610,6 +610,7 @@ def filter_treatments(_treatments: Iterable[str], depth: int = 1) -> List[str]:
             [
                 "".join(f"{x}_" for i, x in enumerate(t.split('_')) if i < depth).rstrip('_') + '_*'
                 for t in set(_treatments)
+                if t # In the event an abstract base class is detected, that does not set TREATMENT_ID by default
```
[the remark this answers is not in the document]
willGraham01: As far as I can tell, this portion of the code uses `iterpackage` to read the package's source code and detect (HSI-event specific) classes that are defined in the various disease modules.

It then looks through the HSI-event classes it decides it has found to determine all the possible `treatment_id`s that are provided. However, some classes like `HSI_BaseGenericFirstAppt` do not set a `treatment_id` by default (since it is overwritten by derived classes). In the list (technically `set`) of `treatment_ids` that is being assembled, this causes an empty string to be added to the list. The test then fails since the list of `treatment_id`s now contains this empty string, whilst the reference set of `treatment_id`s does not. Adding this `if t` check ensures that these empty strings are filtered out.

src/tlo/methods/hsi_generic_first_appts.py
willGraham01: Yeah it is only used by the `labour` module, however switching to the `labour` module's `rng` attribute then causes the deterministic simulation test to fail (clearly the two random states generate a different random choice for the facility level at some point).

src/tlo/methods/hsi_generic_first_appts.py
willGraham01: Only because the original issue suggested doing everything (DF updates + event scheduling) outside the module-level methods. Though scheduling the HSI events within the methods themselves presumably won't hurt - will do some reworking

src/tlo/analysis/utils.py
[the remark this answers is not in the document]
matt-graham: Ah that makes sense - thanks for the explanation. I should have expanded the code context more around the changed snippet!

src/tlo/methods/hsi_generic_first_appts.py
matt-graham: Hmm - I'm surprised that test fails as the module specific random number generator instances are also also deterministically seeded. I wonder if there another occurence of the dreaded non-deterministic ordering of iterating over sets coming in to play 🤔. I will look in to this more!

src/tlo/methods/hsi_generic_first_appts.py
matt-graham: Ah sorry I guess you mean the last comment
>Alternatively at the cost of more complex function signatures, we could make do_at_generic_first_appt_non_emergency and do_at_generic_first_appt_emergency static methods of the modules, without direct access to the module attributes (including the population data frame). In this case we would need to pass through all objects that would be needed for dealing with scheduling following on HSI events as arguments.
Yeah I think thought at that point this could be an alternative to make things stricter in terms of what the `Module` implementations of the method could do, but specifically I was thinking here to force that they are static methods so they don't have access to the instance at all. Thanks for updating!

matt-graham (changes requested):
Adding some further comments and suggestions. Some overall points

- `NamedTuple` is not a valid type hint I think for the `patient_details` arguments as the types returned by `namedtuple` function are subclasses of `tuple` but not of `NamedTuple`. You could possibly either use the stringified type hint `"PatientDetails"` though this is a bit odd as the type is only defined when `_do_on_generic_first_appt` function is called, use `tuple` as typehint, which is equally a bit unclear or omit a type hint altogether.
- I commented once below about changing `df_updates` to `patent_details_updates` for consistency with the naming of the `patient_details` argument and to make what is being updated clearer. I think this is only a minor improvement and probably a fair bit of faff to change consistently everywhere so possibly not worth effort.
- As the snippet `patient_details = namedtuple("PatientDetails", df.columns)(*df.loc[person_id])` is repeated in both `hsi_generic_first_appts` module and several test modules probably worth factoring it out in to a function or method. From the comment in the [`namedtuple` documentation](https://docs.python.org/3/library/collections.html#collections.namedtuple) that `To support pickling, the named tuple class should be assigned to a variable that matches _typename_.` it might also be worth separating out the creation of the type and creation of an instance of the type as we want to be able to support pickling to allow checkpointing (this might not apply with `dill` but probably still worthwhile from a readability perspective).
- As in most cases the overriden `do_at_generic_first_appt` methods (and corresponding emergency variants) return an empty dictionary of updates to the patient details, we could avoid some boilerplate by having a return of `None` (or equivalently no `return` statement at all) be regarded as equivalent to returning an empty dict. This would require a bit of refactoring in `hsi_generic_first_appts` module to check for `None` returns and act accordingly, but would remove lots of `return {}` instances.

src/tlo/methods/rti.py
```diff
@@ -2500,17 +2487,104 @@ def rti_assign_injuries(self, number):
...
+        # Things to do upon a person presenting at a Non-Emergency Generic
+        # HSI if they have an injury.
+        persons_injuries = [
+            getattr(patient_details, injury) for injury in RTI.INJURY_COLUMNS
+        ]
+        if (
+            pd.isnull(patient_details.cause_of_death)
+            and not patient_details.rt_diagnosed
+        ):
+            if set(RTI.INJURIES_REQ_IMAGING).intersection(set(persons_injuries)):
```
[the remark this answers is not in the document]
matt-graham: Oh interesting! No need to change then, I vaguely thought there was some sort of recommendation in PEP8 to avoid implicit falseyness tests like this but I checked and there isn't so I think it's just me imagining it and better to stick with the linter recommendations

src/tlo/methods/cardio_metabolic_disorders.py
```diff
@@ -208,6 +208,13 @@ class CardioMetabolicDisorders(Module):
                   'nc_risk_score': Property(Types.INT, 'score to represent number of risk conditions the person has')
                   }
 
+    @staticmethod
+    def is_next_test_due(current_date, date_of_last_test):
+        return (
+            pd.isnull(date_of_last_test)
+            or (current_date - date_of_last_test).days > DAYS_IN_YEAR / 2
+        )
+
```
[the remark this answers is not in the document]
willGraham01: It was. Though the method itself is only actually called once (across the entire class & codebase), so we could even just resolve this with a variable assignment rather than a nested function too.

src/tlo/methods/stunting.py
```diff
@@ -275,31 +275,44 @@ def do_recovery(self, idx: Union[list, pd.Index]):
...
+            # probability of stunting diagnosis
+            if p_stunting_diagnosed > self.rng.random_sample():
+                event = HSI_Stunting_ComplementaryFeeding(
+                    module=self, person_id=patient_id
+                )
+                self.healthsystem.schedule_hsi_event(
+                    event,
+                    priority=2,  # <-- lower priority that for wasting and most other HSI
+                    topen=self.sim.date,
+                )
```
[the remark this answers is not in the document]
willGraham01: Yep good catch, the Bernoulli RV in the outer `if` statement shouldn't be there, it should only appear in the block below. The original implementation also only generates this RV if `is_stunted` is `True`, which is why there's the seemingly random nested `if` statement (rather than the combined one).

src/tlo/methods/malaria.py
```diff
@@ -662,100 +664,138 @@ def report_daly_values(self):
...
-        """Run by an HSI when an adult presents with fever"""
+    def check_if_fever_is_caused_by_malaria(
+        self,
+        true_malaria_infection_type: str,
+        diagnosis_function: DiagnosisFunction,
+        patient_id: int = None,
+        fever_is_a_symptom: bool = True,
+        patient_age: Union[int, float] = None,
+        facility_level: str = None,
+        treatment_id: str = None,
```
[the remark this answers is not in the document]
willGraham01: The tests in `test_malaria.py` imply that these arguments are required with the corresponding default values (they fail otherwise).

Have added the `Optional` wrapper accordingly.

willGraham01:
Thanks Matt, have responded to the smaller comments where they're present.

As for the general stuff:

> NamedTuple is not a valid type hint I think for the patient_details arguments as the types returned by namedtuple ...
> As the snippet `patient_details = namedtuple("PatientDetails", df.columns)(*df.loc[person_id])` ...

Refactoring the snippet was on my radar of future low-hanging fruit, so I've done that now.
The `Population` class now has a method `row_in_readonly_form` which does what this code snippet does, and it returns a `PatientDetails` type (which is an alias of `Tuple` in accordance with the return type of `namedtuple`).

I put it in this class since we need access to the DataFrame during the method anyway, and the PatientDetails object can only be made from the population DataFrame (since it needs the column names). `sim.population` is always exposed in the cases where we need to run this method.

> I commented once below about changing df_updates to patent_details_updates for consistency with the naming of the patient_details argument ...

VSCode `F2`-refactor has taken care of this :sweat_smile: 

> As in most cases the overriden do_at_generic_first_appt methods (and corresponding emergency variants) return an empty dictionary ...

The function docstring, module methods, and `hsi_generic_first_appt` class method has been updated accordingly. There will be some cases where returning an empty dictionary is more convenient than returning nothing (or omitting the return), which I tried to capture with the cases in the docstring's explanation of the return value. You might want to double-check that it's clear.

src/tlo/core.py
```diff
@@ -445,19 +445,19 @@ def do_at_generic_first_appt(
         :param treatment_id: The treatment id of the HSI event triggering the generic appointment.
         :param random_state: Random number generator to be used when making random choices during event creation.
         """
-        return {}
+        return None
```
matt-graham, line 448:
```suggestion
        return
```

As we're recommending to not use `return None` in the docstring. I think you could also just omit the `return` statement entirely here (as the docstring is sufficient for this to qualify as valid function definition with indendented body) which would align with the prefered style in docstring.

src/tlo/core.py
```diff
@@ -467,4 +467,4 @@ def do_at_generic_first_appt_emergency(
         events when a patient presents symptoms indicative of the
         corresponding illness or condition.
         """
-        return {}
+        return None
```
matt-graham, line 470:
```suggestion
        return
```
Same rationale as above

matt-graham (approved):
Thanks @willGraham01 for all your work on this and putting up with my many rounds of comments! This is looking fantastic - irrespective of the performance aspect I think this a really nice refactoring from making the code more readable and maintainable perspective. And 🤞 it will bring some performance improvement! I've suggested a very minor change to default return semantics on methods on base class for consistency with recommendations in docstring, but this is very minor.

willGraham01:
@matt-graham Tests are passing; and despite [the last profiling run on master](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_591_012ef6d7fd1e5635ce1987abd70cda0f8d5e4ae9.html) taking less total time than [the one I set off above](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/issue_comment_601_fa709e20bb84783e89543baabc58c446cd680704.html), we are actually spending less (absolute!) time inside the `do_at_generic_first_appt` with these changes... granted that it's only 78 seconds.

If you're happy with the current state, shall we merge this in?

matt-graham:
Yep lets get this merged!

PR #1296, opened 2024-03-08
Set date at which capabilities are rescaled and add rescaling by district

marghe-molaro:
This PR replaces #1241, now ensuring that user can specify date on which rescaling of capabilities by level & officer type takes place (whereas before it would always occur at start of simulation on 1st of Jan 2010).

In addition, can now schedule a separate event to rescale capabilities based on district. This will be useful for modelling catastrophic events, which may affect specific districts but not others.


tbhallett:
Builds on:
* #1297

tbhallett (approved):

PR #1297, opened 2024-03-09
Increase range of options for yearly rescaling of HR capabilities

marghe-molaro:
Following discussion on 03/08 at the Health Econ meeting, we decided to add some flexibility to the yearly rescaling of HR capabilities. 

The yearly rescaling is controlled by two factors, one fixed and user-specified (dynamic_HR_scaling_factor) and one determined by the relative growth of the population size from the previous year (control whether to include this or now via scale_HR_by_popsize, boolean)

Under this PR, the user can now specify any number of changes to these factors by specifying the years in which such changes should take place in file ResourceFile_dynamic_HR_scaling.xlsx

The user should always specify values for 2010, which will be the initially assumed values at the start of the simulation. However if the user would like to change these values e.g. in 2023, they can simply add a new row with 2023 and the new values for the two variables in the resource file. These values will be updated accordingly once the simulation reaches the 1st of Jan 2023. Any number of these changes can be considered, providing maximum flexibility to the modelling of scenarios.


tbhallett:
Thanks @marghe-molaro 
This is perfect.
As discussed, I've added a new test to check that the specification of the configuration in each year works as expected (`test_dynamic_HR_scaling_multiple_changes`).

The final commit propose a restructure of the event `DynamicRescalingHRCapabilities` for speed (avoiding using `pandas.DataFrame.loc`) and readability.... but the latter is highly subjective. So you can roll it back if you prefer the original way.

tbhallett (approved):

PR #1299, opened 2024-03-19
Force scale_run distinction to check which version is running

willGraham01:


willGraham01:
Confirming that the correct SHA is checked out, but the wrong SHA is being reported in the profiling results. Likely because the SHA passed in the `trigger_on_comment` workflow is that of main, not the PR branch.

PR #1301, opened 2024-03-19
Fix comment-triggered profiling SHA reports

willGraham01:
Builds on #1236 by addressing the following issues:

- Will close https://github.com/UCL/TLOmodel/issues/1123 by adding the `if` condition to prevent the variables being setup, only to then cancel the workflow later.
- Will close #1300 by correcting the logic used when reporting the commit SHA that the profiling workflow ran on.

PR #1306, opened 2024-03-27
Fix profiling workflow failures

willGraham01:
Fixes the bug encountered in https://github.com/UCL/TLOmodel/actions/runs/8397884074, that was introduced in https://github.com/UCL/TLOmodel/pull/1301.

When profiling is triggered by comment on a pull request, the workflow needs to fetch the SHA of the PR branch, rather than the HEAD of `master` (which is the default for `issue_comment`-event triggers). This is in contrast to when it is triggered through scheduling, workflow dispatch, or comments on ISSUES - when it should run on `master`.

Since the `issue_comment` trigger is raised when comments on _issues_ and on _pull requests_ are made, we need to do some manual inspection to distinguish the two cases, hence the need for the additional logic in this pull request.

### DIFF
--- a/.github/workflows/run-profiling.yaml
+++ b/.github/workflows/run-profiling.yaml
@@ -39,20 +39,26 @@ jobs:
       profiling-on-sha: ${{ steps.determine-correct-sha.outputs.result }}
       profiling-event-trigger: ${{ steps.set-github-info.outputs.event }}
     steps:
+      - id: determine-if-pr-comment
+        name: Determine if this is a PR comment
+        run: |
+          if ! [ -z "${{ github.event.issue.pull_request }}" ]; then echo "is-pr=true" >> "${GITHUB_OUTPUT}"; else echo "is-pr=false" >> "${GITHUB_OUTPUT}"; fi
+
       - id: determine-correct-sha
+        name: Determine the SHA to run profiling on
         uses: actions/github-script@v7
         with:
           result-encoding: string
           script: |
-            if (!context.payload.issue.pull_request) {
-              return context.sha;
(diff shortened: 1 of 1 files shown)


.github/workflows/run-profiling.yaml
```diff
@@ -39,20 +39,26 @@ jobs:
       profiling-on-sha: ${{ steps.determine-correct-sha.outputs.result }}
       profiling-event-trigger: ${{ steps.set-github-info.outputs.event }}
     steps:
+      - id: determine-if-pr-comment
+        name: Determine if this is a PR comment
+        run: |
+          if ! [ -z "${{ github.event.issue.pull_request }}" ]; then echo "is-pr=true" >> "${GITHUB_OUTPUT}"; else echo "is-pr=false" >> "${GITHUB_OUTPUT}"; fi
```
matt-graham, line 45: What happens here if `github.event.issue.pull_request` is not defined? Does GitHub interpolate a zero-length string or raise an error? Can see this would give the correct behaviour if the former but if the latter would this not have a similar issue to previous solution as `github.event.issue` and so `github.event.issue.pull_request`  will only be defined when the triggering event was an issue comment I think?

willGraham01: It interpolates an empty string - you can check the (recent) logs in https://github.com/willGraham01/benchmark-action-playtest/actions/workflows/fetch_sha.yml.

That workflow produces no errors in any of the cases (trigger by push, PR comment, issue comment) and returns the expected SHA in each case. Sorry - should've probably linked to this in the main text!

matt-graham: Thanks for verifying - this looks goods to go then.

matt-graham (approved):

PR #1308, opened 2024-04-05
Option to rescale to "effective capabilities" before switching to mode 2

marghe-molaro:
In mode 2 we do not account for potential overtime and rushing of appointments which may be taking place in reality. In mode 1, however, we assume these can take place unconstrained to ensure that all appointments can be delivered on the day they are requested. As a result, the transition from mode 1 to mode 2 usually results in a very steep increase in DALYs in the first few years, which would be unlikely to occur in reality.

Following @tbhallett's suggestion, we now propose the following option to allow the user to simulate HR resource-constraints in mode 2 while still somewhat capturing the effect of overtime and rushing of appointments. 
- Every year, we calculate the average fraction of daily minutes "used" by each medical officer type at each level. This gives an estimate of the combined (if any) overtime+rushing of appointments for those officer types + level in that year;
- If the HealthSystem parameter "scale_to_effective_capabilities" is set to True, at the end of the year *preceding* the mode 1 --> 2 switch we rescale the daily capabilities of officers that experienced overtime+rushing of appts by those fractions. This means that the capabilities assumed in mode 2 will be rescaled to those "effectively" used in the preceding year.
- Notice that if officers did not experience overtime+rushing of appts in previous year, their capabilities are not rescaled (i.e. we do not reduce capabilities of underused officers).


src/tlo/methods/healthsystem.py
```diff
@@ -203,6 +203,10 @@ class HealthSystem(Module):
             Types.STRING, "Name of priority policy adopted"),
         'year_mode_switch': Parameter(
             Types.INT, "Year in which mode switch is enforced"),
+         'scale_to_effective_capabilities': Parameter(
```
matt-graham, line 206:
```suggestion
        'scale_to_effective_capabilities': Parameter(
```
To match other indentation

src/tlo/methods/healthsystem.py
```diff
@@ -1032,6 +1052,20 @@ def get_mode_appt_constraints(self) -> int:
...
+        return self.parameters['year_mode_switch'] \
+            if self.arg_year_mode_switch is None \
+            else self.arg_year_mode_switch
+
+    def get_scale_to_effective_capabilities(self) -> int:
+        """Returns `scale_to_effective_capabilities`. (Should be equal to what is specified by the parameter, but overwrite with
+        what was provided in argument if an argument was specified -- provided for backward compatibility/debugging.)"""
+        return self.parameters['scale_to_effective_capabilities'] \
+            if self.arg_scale_to_effective_capabilities is None \
+            else self.arg_scale_to_effective_capabilities
```
matt-graham, line 1068:
```suggestion
    @property
    def year_mode_switch(self) -> int:
        """Returns `year_mode_switch`. (Should be equal to what is specified by the parameter, but overwrite with
        what was provided in argument if an argument was specified -- provided for backward compatibility/debugging.)"""
        return (
            self.parameters['year_mode_switch']
            if self._arg_year_mode_switch is None
            else self._arg_year_mode_switch
        )

    @property
    def scale_to_effective_capabilities(self) -> int:
        """Returns `scale_to_effective_capabilities`. (Should be equal to what is specified by the parameter, but overwrite with
        what was provided in argument if an argument was specified -- provided for backward compatibility/debugging.)"""
        return (
            self.parameters['scale_to_effective_capabilities']
            if self._arg_scale_to_effective_capabilities is None
            else self._arg_scale_to_effective_capabilities
        )
```

Decorating these methods as properties allows accessing like standard attributes (for example `health_system.scale_to_effective_capabilities`) and is more 'Pythonic' than defining getter-methods. Also generally [explicit line continuation characters are discouraged in Python in favour of using implicit line continuation by wrapping in parenthesis](https://peps.python.org/pep-0008/#maximum-line-length).

src/tlo/methods/healthsystem.py
```diff
@@ -584,6 +598,12 @@ def pre_initialise_population(self):
 
         # Determine mode_appt_constraints
         self.mode_appt_constraints = self.get_mode_appt_constraints()
+        
+        # Determine year mode switch
+        self.year_mode_switch = self.get_year_mode_switch()
+
+        # Determine year mode switch
+        self.scale_to_effective_capabilities = self.get_scale_to_effective_capabilities()
```
matt-graham, line 606:
```suggestion
```
If changing the `get_*` methods to properties as suggested below these definitions can be removed.

src/tlo/methods/healthsystem.py
```diff
@@ -379,6 +387,12 @@ def __init__(
         if mode_appt_constraints is not None:
             assert mode_appt_constraints in {0, 1, 2}
         self.arg_mode_appt_constraints = mode_appt_constraints
+        
+        self.year_mode_switch = None
+        self.arg_year_mode_switch = year_mode_switch
+        
+        self.scale_to_effective_capabilities = None
+        self.arg_scale_to_effective_capabilities = scale_to_effective_capabilities
```
matt-graham, line 395:
```suggestion
        self._arg_year_mode_switch = year_mode_switch
        self._arg_scale_to_effective_capabilities = scale_to_effective_capabilities
```
The `self.arg_*` attributes seem to be only for internal use so should probably be prefixed with an underscore to hint this and there is no need to separately initialise the `self.scale_to_effective_capabilities` and `self.year_mode_switch` attributes if changing to properties as suggested below.

src/tlo/methods/healthsystem.py
```diff
@@ -1687,7 +1721,13 @@ def log_current_capabilities_and_usage(self):
...
         ).replace([np.inf, -np.inf, np.nan], 0.0)
         summary_by_officer.index.names = ['Officer_Type', 'Facility_Level']
-
+        
+        # Store information on fraction of time used, broken down by officer type and level
+        summary_by_officer['OfficerType_FacilityLevel'] = summary_by_officer.index.get_level_values('Officer_Type') + '_' + summary_by_officer.index.get_level_values('Facility_Level')
+        
+        for index, row in summary_by_officer.iterrows():
+            self._summary_counter._frac_time_used_by_officer_type_and_level[row['OfficerType_FacilityLevel']] += float(row['Fraction_Time_Used'])
+
```
matt-graham, line 1730: Accessing an attribute of another object prefixed with an underscore `_` should generally be avoided as the convention is that these are for internal use only within the object (so accessing `self._summary_counter` here is fine as `_summary_counter` is an internal attribute of the module, but `_summary_counter._frac_time_used_by_officer_type_and_level` should only be accessed by the `HealthSystemSummaryCounter` class). 

One option would be to just mark `_frac_time_used_by_officer_type_and_level` as for external use by removing `_` prefix, but I would say it would be better here to extend the existing `HealthSystemSummaryCounter.record_hs_status` method to deal with also updating the fractions of time by officer type and level counter as something like
```Python
    def record_hs_status(
        self,
        fraction_time_used_across_all_facilities: float,
        fraction_time_used_by_officer_type_and_level: Dict[Tuple[str, int], float],  
    ) -> None:
        """Record a current status metric of the HealthSystem."""
        # The fraction of all healthcare worker time that is used:
        self._frac_time_used_overall.append(fraction_time_used_across_all_facilities)
        for officer_type_facility_level, fraction_time in fraction_time_used_by_officer_type_and_level.items():
            self._sum_frac_time_used_by_officer_type_and_level[officer_type_facility_level] += fraction_time
```
and then calling here by replacing the current line 1741 with
```
    self._summary_counter.record_hs_status(
        fraction_time_used_across_all_facilities=fraction_time_used_overall,
        fraction_time_used_by_officer_type_and_level=summary_by_officer["Fraction_Time_Used"].to_dict()
    )
```

I would also say an attribute name `_sum_frac_time_used_by_officer_type_and_level` would be more explicit that this counter is recording sum of fractions of time hence the above also suggests changing this name.

src/tlo/methods/healthsystem.py
```diff
@@ -2473,6 +2515,8 @@ def _reset_internal_stores(self) -> None:
         self._never_ran_appts_by_level = {_level: defaultdict(int) for _level in ('0', '1a', '1b', '2', '3', '4')}
 
         self._frac_time_used_overall = []  # Running record of the usage of the healthcare system
+        self._frac_time_used_by_officer_type_and_level = Counter()
```
matt-graham, line 2518:
```suggestion
        self._sum_frac_time_used_by_officer_type_and_level = Counter()
```
I think prefixing with `sum` here makes it more explicit this is a sum of fractions (see also suggested change above).

src/tlo/methods/healthsystem.py
```diff
@@ -2523,6 +2567,31 @@ def record_hs_status(self, fraction_time_used_across_all_facilities: float) -> N
...
+                matches = re.match(pattern, officer)
+                # Extract ID and officer type from
+                facility_id = int(matches.group(1))
+                officer_type = matches.group(2)
+                level = self.module._facility_by_facility_id[facility_id].level
+                # Only rescale if rescaling factor is greater than 1 (i.e. don't reduce available capabilities
+                # if these were under-used the previous year).
+                if self._frac_time_used_by_officer_type_and_level[officer_type + "_" + level] > 1:
+                    self.module._daily_capabilities[officer] *= \
+                        self._frac_time_used_by_officer_type_and_level[officer_type + "_" + level]
```
matt-graham, line 2594: This feels like something that should live in a method in `HealthSystem` module rather than here as it is largely to do with updating attributes of that module which should be the responsibility of that class, and doing it in this class requires us to store a reference to `HealthSystem` module which couples the implementations making it more difficult to test and maintain. I would say create a method of `HealthSystemSummaryCounter` to allow accessing the internal `_sum_frac_time_used_by_officer_type_and_level` as a fraction for a particular office type and level *,
```Python
    def frac_time_used_by_officer_type_and_level(self, officer_type, level):
        """Average fraction of time used by officer type and level since last reset."""
        # Use len(self._frac_time_used_overall) as proxy for number of days in past year.
        return (
            self._sum_frac_time_by_officer_type_and_level[officer_type, level] 
            / len(self._frac_time_used_overall)
        )
```
define a new internal method of `HealthSystem` module
```Python
    def _rescale_capabilities_to_capture_effective_capability(self):
        # Notice that capabilities will only be expanded through this process 
        # (i.e. won't reduce available capabilities if these were under-used in the last year).
        # Note: Currently relying on module variable rather than parameter for 
        # scale_to_effective_capabilities, in order to facilitate testing. However
        # this may eventually come into conflict with the Switcher functions.
        pattern = r"FacilityID_(\w+)_Officer_(\w+)"
        for officer in self.module._daily_capabilities.keys():
            matches = re.match(pattern, officer)
            # Extract ID and officer type from
            facility_id = int(matches.group(1))
            officer_type = matches.group(2)
            level = self.module._facility_by_facility_id[facility_id].level
            # Only rescale if rescaling factor is greater than 1 (i.e. don't reduce 
            # available capabilities if these were under-used the previous year).
            rescaling_factor = self._summary_counter.frac_time_used_by_officer_type_and_level(
                officer_type, level
            )
            if rescaling_factor > 1:
                self.module._daily_capabilities[officer] *= rescaling_factor 
```
and update the existing `on_end_of_year` method of `HealthSystem` to something like
```Python
    def on_end_of_year(self) -> None:
        """Write to log the current states of the summary counters and reset them."""
        # If we are at the end of the year preceeding the mode switch, and if wanted 
        # to rescale capabilities to capture effective availability as was recorded, on
        # average, in the past year, do so here.
        if (self.sim.date.year == self.year_mode_switch - 1) and self.scale_to_effective_capabilities:
            self._rescale_capabilities_to_capture_effective_capability()
        self._summary_counter.write_to_log_and_reset_counters()
        ...
```

*Note this also assumes that the relevant internal attribute of `HealthSystemSummaryCounter` is renamed to `_sum_frac_time_by_officer_type_and_level` as suggested above and is a `Counter` keyed by a tuple of officer type and level rather than the concatenated string, again as suggested above.

matt-graham (changes requested):

src/tlo/methods/healthsystem.py
matt-graham: Depending on whether you are viewing this in conversation view or file view on GitHub 'below' might actually be above 😅

src/tlo/methods/healthsystem.py
matt-graham: Note this also changes the `_sum_frac_time_used_by_officer_type_and_level ` counter to be keyed by a tuple of officer type and level rather than a concatenated string.

tbhallett (approved):
Thanks for this Margherita. (and also Matt for great and super quick review.

This is really nice little change that you've already proven does exactly what we want. And the test is nice too.

These comments of mine are in two categories; (i) agreeing with you say yourself about removing the key-word argument interface; (ii) agreeing with the refactoring proposed by Matt.

I'm sorry that this will be a bit annoying as we know that it does really work already 'as is', and these changes are only organisational/stylistic etc. However, as Matt has made the changes as suggestions already, hopefully it's not too bad.

I'm clicking Approve now, as I'm perfectly happy with the core logic, test, functionality.

matt-graham (changes requested):
Thanks @marghe-molaro for making those changes, this is looking great. I've made a couple of minor suggested changes with respect to removing introduced reference to `HealthSystem` module in `HealthSystemSummaryCounter` class as I think with the new design this is no longer required.

matt-graham (approved):

tbhallett:
> @tbhallett let me know when the other merges have gone into master and I'll submit the new batch of runs

The changes are going to come in on https://github.com/UCL/TLOmodel/pull/1326

tbhallett:
> > @tbhallett let me know when the other merges have gone into master and I'll submit the new batch of runs
> 
> The changes (allowing ScenaroSwitcher to occur midway through the simulation) are going to come in on #1326

@marghe-molaro  #1326 now in so you can finish-off these scenarios, and then we can (i) merge this branch; (ii) submit your runs.

marghe-molaro:
Hi @tbhallett,  many thanks for the merges. Scenarios are ready at:
src/scripts/healthsystem/impact_of_const_capabilities_expansion/scenario_impact_of_capabilities_expansion_scaling.py

If you could double check they look ok to you I'll submit the jobs.

tbhallett:
> Hi @tbhallett, many thanks for the merges. Scenarios are ready at: src/scripts/healthsystem/impact_of_const_capabilities_expansion/scenario_impact_of_capabilities_expansion_scaling.py
> 
> If you could double check they look ok to you I'll submit the jobs.

That file looks great to me.

tbhallett:
I'll merge this now, as we're happy with the functionality. (Any further tweaks to the scenarios / plotting scripts can obviously follow in separate PR).

PR #1310, opened 2024-04-08
Add scheduled workflow to run tests with unpinned dependencies

### LINKED ISSUE #1161: Add scheduled workflow job running tests on latest stable versions of dependencies
To get prior warning of things breaking on new releases of Python package dependencies and/or Python versions, we should set up a scheduled workflow to run for example weekly, running tests in an environment with latest stable Python release and latest versions of packages installed from (non-locked) dependencies.

matt-graham:
Fixes #1161 

Adds a new workflow plus associated tox environment to run tests with unpinned dependencies on latest stable Python version once a week (Satuday at midnight). This currently is set up to run on GitHub hosted runners to allow testing across multiple operating systems and uses `pytest-xdist` plug-in to parallelise across available cores on the runner, rather than creating a matrix to parallelise across test files as in the main tests workflow. It is not clear if the full tests will complete in the 6 hours maximum GitHub Actions job time on GitHub hosted runners (may do though given Ubuntu runners now have 4 cores by default and I think the MacOS runners have 3 cores).

Questions:
- Do we want to test against a specific Python version (3.11?) rather than or in addition to latest stable Python version?
- Is once a week the appropriate frequency? Possibly once a month would suffice?
- Should this be done on self-hosted runners instead?


.github/workflows/tests-unpinned.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  schedule:
+    - cron: 0 0 * * 6
+
+jobs:
+  test:
+    name: Run tests
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-14, windows-latest]
```
willGraham01, line 12: What's the reason for `macos-14` as opposed to `macos-latest`? If we want forewarning on new versions / OSes, it seems counter-intuitive to pin the macos version here.

tox.ini
```diff
@@ -41,6 +41,13 @@ deps =
     pytest
     pytest-cov
 
+[testenv:py3-latest]
+deps =
+    pytest
+    pytest-xdist
+commands =
+    {posargs:pytest -n auto -vv tests}
```
willGraham01, line 49: If we're not sure the tests will pass in <6 hours, we could `--skip-slow` here?

A lot of the slow tests are only slow in lieu of running (essentially a) full simulation. Any breakages in running simulations - by this logic - I would presume would be flagged up by the other tests that run shorter simulations. Or at least, I'd be prepared to sacrifice the slow tests for complete information about whether the non-slow ones passed :sweat_smile:

Plus we'll always be running the full test suite in any PR against `master` that does bump the dependencies, so if something does slip through the cracks we'll still catch it before releasing it into `master`.

.github/workflows/tests-unpinned.yml
matt-graham: Sorry should have put a comment explaining this and also adding a TODO to update to `macos-latest` in future. Currently `macos-latest` I think maps to an earlier MacOS version that importantly doesn't run on Apple Silicon, although this is due to change in the next few months- see https://github.com/orgs/community/discussions/102846 and https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/ . I think testing against MacOS on M1 would be of the most value as we seem to have been hitting problems with dependencies most often on Apple Silicon machines. Will add an explanatory comment with a link to the discussion and a note to update to `macos-latest` once this is update to point to `macos-14` or later

tox.ini
matt-graham: Yeah agree this is probably worthwhile doing - will update.

.github/workflows/tests-unpinned.yml
matt-graham: Hmm actually from https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories it looks like this has now been updated so looks like can just move directly to `macos-latest`

tamuri:
> Do we want to test against a specific Python version (3.11?) rather than or in addition to latest stable Python version?

This workflow should be our canary in the coal mine - latest stable Python & packages. To that end, we state `pandas~=2.0.0` in the pyproject. Can we ignore that too?

> Is once a week the appropriate frequency? Possibly once a month would suffice?

Once a month is probably okay. Don't want to get too distracted.

> Should this be done on self-hosted runners instead?

We'd only be targetting Linux in that case. I do like the the extensive coverage we get if we use GH runners.

matt-graham:
> This workflow should be our canary in the coal mine - latest stable Python & packages. To that end, we state `pandas~=2.0.0` in the pyproject. Can we ignore that too?

Agree it would be worth testing against unpinned pandas versions as this is probably most likely dependency we'll get breaking changes in. We could remove the version specified for pandas from `dependencies` in `pyproject.toml` altogether or relax `~=2.0.0` to `>=2`, `~=2.0` or similar. While this would then mean the new workflow here tests against latest stable pandas version, it would also mean that it becomes more of a pain to update the requirements files with `pip-compile`, as without the version specifier `pip-compile` would also I think resolve to the latest stable `pandas` version at the time run, and currently the `tlo` package will not work with pandas v2.1 or newer I believe. It looks like this would be relatively easy to workaround by using the `--upgrade--packages` argument to `pip-compile` to specify a constraint on the `pandas` versions though. Altermatively we could install the dependencies manually in the new tox environment, and force it to not ignore / not install package dependencies by updating the `install_command` option with `--no-deps` but this feels a bit hacky and requires us to keep the base requirements in two different places.

> Once a month is probably okay. Don't want to get too distracted.

Will update.

> We'd only be targetting Linux in that case. I do like the the extensive coverage we get if we use GH runners.

👍

matt-graham:
@tamuri I've finally got around to addressing the two points in my comment above (changing frequency to monthly and relaxing version requirement on pandas).

tamuri (approved):

PR #1313, opened 2024-04-08
Allowing passing multiple job IDs to `tlo batch-download` command

### LINKED ISSUE #828: Add command or recipe for downloading results from multiple batch scenario runs at once
As mentioned by @joehcollins in York workshop - it could be useful to either to allow passing a list of job IDs to `tlo batch-download` or having a different subcommand that allows downloading the results from multiple scenario jobs at once. Alternatively we might just want to have a recipe for doing this using a script in the wiki.

matt-graham:
Fixes #828 

Renames `job_id` argument to `tlo batch-download` to `job_ids` and changes to allow accepting an unlimited number of values using [`click` `nargs=-1` option to specify variadic argument](https://click.palletsprojects.com/en/latest/arguments/#variadic-arguments, and changes implementation to loop over the resulting tuple of job IDs, and perform downloading for each in sequence. The current implementation continues to try to download subsequent job ID output files if the local directory for a previous job ID already exists rather than exiting altogether. When passed one job ID this should still match the behaviour of the current implementation.

As [recommended in the `click` documentation](https://click.palletsprojects.com/en/latest/arguments/#variadic-arguments), the implementation allows for no job IDs being passed rather than setting `required=True` to enforce at least one being provided, to provide more graceful behaviour when used with wildcard arguments which might expand to zero matches.

### DIFF
--- a/src/tlo/cli.py
+++ b/src/tlo/cli.py
@@ -469,11 +469,11 @@ def print_basic_job_details(job: dict):
 
 
 @cli.command()
-@click.argument("job_id", type=str)
+@click.argument("job_ids", type=str, nargs=-1)
 @click.option("--username", type=str, hidden=True)
 @click.option("--verbose", default=False, is_flag=True, hidden=True)
 @click.pass_context
-def batch_download(ctx, job_id, username, verbose):
+def batch_download(ctx, job_ids, username, verbose):
     """Download output files for a job."""
     config = load_config(ctx.obj["config_file"])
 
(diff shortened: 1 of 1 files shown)


matt-graham:
@joehcollins - have requested a review from you to double check this does address the issue you mentioned at the York quarterly meeting sometime ago! With the changes here it should be possible to run for example
```
tlo batch-download <jobid1> <jobid2> <jobid3>
```
to download results for three specific job IDs

joehcollins (approved):
Hi @matt-graham - completely forgot about this! But looking at the example and the for loop in the code this does look exactly like what we'd mentioned before. Thanks!!

tamuri (approved):

PR #1315, opened 2024-04-12
Access individual properties lazily with memoization in `PatientDetails` type

matt-graham:
Follow on from #1292 

Changes the `PatientDetails` type used to mediate access to individual property values in the generic first appointments HSI event to be a class with provides a read-only memoized view of the relevant population dataframe row as opposed to a `namedtuple` instance generated from the `Series` object representing a particular row. On first access the individual properties are read from the dataframe using the `at` indexer, with the read values cached in a dictionary to allow subsequent lookup without accessing the dataframe. This avoids reading in all properties from the dataframe on construction of instanced of `PatientDetails`, of which only a subset may be used downstream, however, it comes with the potential of greater overhead from multiple dataframe accesses (which is what I thought was the performance bottleneck in discussion in #1237). Not clear overall which is the big factor - from some short profiling runs locally looks like the approach in this PR might be slightly more performant but unclear if this is just due to variance in profiling runs. 

Also adds in an unrelated change of wrapping the update to individual properties in `_do_on_generic_first_appt` in an `if` block to only perform if `proposed_patient_details_updates` is non-empty (to avoid overhead from entering Pandas `__setitem__` logic with an empty set of updates).

### DIFF
--- a/src/tlo/population.py
+++ b/src/tlo/population.py
@@ -1,8 +1,7 @@
 """The Person and Population classes."""
 
 import math
-from collections import namedtuple
-from typing import Any, Tuple, Type, TypeAlias
+from typing import Any, Dict
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


matt-graham:
From [results of profiling run](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/issue_comment_617_c0cd7333f9ffe2884d3f80e172187a1f8aef9054.html triggered above compared to [results of run on current `master`](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_621_bd0a7632e1b0add9ae0ae8d3afc2a151bdd15f7d.html) it looks like this probably does give a small performance boost so marking this as ready to review.

src/tlo/population.py
```diff
@@ -11,8 +10,31 @@
...
+    """Read-only memoized view of population dataframe row."""
+    
+    def __init__(self, population_dataframe: pd.DataFrame, person_id: int):
+        self._population_dataframe = population_dataframe
+        self._person_id = person_id
+        self._property_cache: Dict[str, Any] = {}
+        
+    def __getitem__(self, key: str) -> Any:
+        cached_value = self._property_cache.get(key)
+        if cached_value is not None:
```
willGraham01, line 24: I'm sure it's fine for now, since as far as I've seen the convention for missing data in the simulation is NaN or Inf; but I can forsee a situation where the DataFrame value genuinely is `None`. In which case with this code we'd be reading the property over and over again from the DataFrame.

But I also don't know what else we can do in this situation anyway.

willGraham01 (approved):

src/tlo/population.py
matt-graham: Yeah that's a good point, definitely worth making sure this is future proofed. We could great a dedicated sentinel type to use here or index directly and wrap in a `try ... except` block.

PR #1320, opened 2024-04-22
Maternal/Newborn health cohort model

joehcollins:
This is the draft PR for development of the MNH cohort model for upcoming analyses/calibration


joehcollins:
@tbhallett @matt-graham @tamuri - I'm trying to set up some functionality for the model to be able to run with a cohort of only pregnant women. I've generated a dataframe using a 100k full model run of the properties of all women at pregnancy onset for the simulation. I am using the properties of the women who became pregnant in 2024 to overlap the properties of those within the simulation at baseline. 

I've managed to do this in a hacky way (see `mnh_cohort_module.py`) which works in practice but there are a couple of issues:

1. This only works when I put the logic in `initialise_population` and ensure the cohort module runs first as there are a number of modules who schedule HSIs etc off the initial `population.props` within `initialise_simulation `- theres probably a way around this with dependencies/or something cleverer (see below)?
2. A more pressing issue - the new class `PatientDetails` in `population.py` is retaining the old version of the dataframe after the dataframe has been updated with the details of the pregnancy cohort. This is causing issues with generic HSIs being scheduled based on details which are different from this class to the real dataframe. If you run `test_mnh_cohort.py `on this branch you'll see what I mean r.e. RTI when debugging.

When you have some time coming up I would love to discuss an alternative approach than the one i'm using now which is maybe neater/cleaner and could avoid some of these issues.

joehcollins:
(@mnjowe @thewati FYI)

mnjowe:
Thanks Joe

joehcollins:
@matt-graham @tamuri - Tim mentioned you were thinking more on how best to approach this? Would be great to catch up to see if I can help at all

joehcollins:
@tbhallett - quick question r.e. our convo earlier. As we said, we want to start the sim at 2024 - issue is that across the modules there quite a few instances of events (mostly related to epi) being scheduled on a set year within the burn in period. Obviously this crashes the code. 

I'm obviously trying to keep changes internal to this 'cohort module' but wanted your thoughts on making in-module changes to avoid these crashes?

joehcollins:
(can give an example currently as i'm having some issues pushing to github - think my internet is acting up)

src/scripts/maternal_perinatal_analyses/cohort_analysis/dummy_cohort_azure_calib.py
```diff
@@ -0,0 +1,129 @@
...
+# Create bar chart with error bars
+fig, ax = plt.subplots()
+ax.bar(labels, means, yerr=errors, capsize=5, alpha=0.7, ecolor='black')
+ax.set_ylabel('MMR')
+ax.set_title('Average MMR under each scenario')
+
+# Adjust label size
+plt.xticks(fontsize=8, rotation=90)
+plt.tight_layout()
+plt.show()
```
joehcollins, line 106: "|"

src/scripts/maternal_perinatal_analyses/cohort_analysis/dummy_cohort_azure_calib.py
```diff
@@ -0,0 +1,129 @@
...
+
+results = {k:get_data_frames(k, results_folder)[0] for k in
+           ['mat_comp_incidence', 'nb_comp_incidence', 'deaths_and_stillbirths','service_coverage',
+            'yearly_mnh_counter_dict']}
+
+results_sum = {k:get_data_frames(k, results_folder)[1] for k in
+               ['mat_comp_incidence', 'nb_comp_incidence', 'deaths_and_stillbirths','service_coverage',
+                'yearly_mnh_counter_dict']}
+
+def get_data(df, draw):
```
joehcollins, line 47: '

'

src/scripts/maternal_perinatal_analyses/cohort_analysis/dummy_cohort_azure_calib.py
```diff
@@ -0,0 +1,195 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+from tlo.analysis.utils import extract_results, get_scenario_outputs, summarize
+
+outputspath = './outputs/sejjj49@ucl.ac.uk/'
+
+scenario = 'block_intervention_test-2024-11-06T145016Z'
+
+results_folder= get_scenario_outputs(scenario, outputspath)[-1]
```
joehcollins, line 14: hfgh

Issue #1144, opened 2023-09-28
Review and possibly merge simulations used in slow running tests


matt-graham:
Running `pytest --durations 0 -n 4` locally the top 20 slowest tests are

```
2733.20s call     tests/test_healthsystem.py::test_which_hsi_can_run[83563095832589325021]
2612.51s call     tests/test_determinism.py::test_scale_run_script_deterministic
2138.37s call     tests/test_fullmodel.py::test_dtypes_and_checksum[83563095832589325021]
2066.90s call     tests/test_contraception.py::test_defaulting_off_method_if_no_consumables_at_population_level[83563095832589325021]
1766.72s call     tests/test_alri.py::test_impact_of_all_hsi[83563095832589325021]
1416.80s call     tests/test_schisto.py::test_run_with_mda[83563095832589325021]
1166.75s call     tests/test_pregnancy_supervisor.py::test_run_with_all_referenced_modules_registered[83563095832589325021]
999.53s call     tests/test_rti.py::test_sim_high_incidence[83563095832589325021]
936.00s call     tests/test_diarrhoea.py::test_zero_deaths_when_perfect_treatment[83563095832589325021]
876.79s call     tests/test_cardiometabolicdisorders.py::test_basic_run[83563095832589325021]
830.83s call     tests/test_epilepsy.py::test_epilepsy_treatment[83563095832589325021]
817.89s call     tests/test_labour.py::test_run_no_constraints[83563095832589325021]
812.20s call     tests/test_malaria.py::test_remove_malaria_test[83563095832589325021]
799.06s call     tests/test_copd.py::test_basic_run[83563095832589325021]
790.04s call     tests/test_newborn_outcomes.py::test_run_and_check_dtypes[83563095832589325021]
782.80s call     tests/test_pregnancy_supervisor.py::test_run_core_modules_normal_allocation_of_pregnancy[83563095832589325021]
765.87s call     tests/test_postnatal.py::test_run_and_check_dtypes[83563095832589325021]
615.82s call     tests/test_consumables.py::test_every_declared_consumable_for_every_possible_hsi_using_actual_data
545.56s call     tests/test_pregnancy_supervisor.py::test_run_core_modules_high_volumes_of_pregnancy[83563095832589325021]
534.58s call     tests/test_pregnancy_supervisor.py::test_run_core_modules_high_volumes_of_pregnancy_hsis_cant_run[83563095832589325021]
```

<details>
[...]

PR #1324, opened 2024-04-24
Trial sharing simulations across tests

willGraham01:
Related to https://github.com/UCL/TLOmodel/issues/1144 | Leaving as a draft for now since this is mainly a proof of concept.

This is a proof of concept for how we might decide to share simulation objects across tests, with a view to speeding up the execution of the test suite.

This PR creates the `_BaseSharedSim` class, whose purpose is to be used as a base class for the module-level tests. We also create the session-scoped `shared_small_sim` fixture, which is a simulation object that is created and run once right at the start of the test session. `_BaseSharedSim` then defines an `autouse` fixture `_attach_to_shared_sim` that allows any tests within the class (and any tests in derived classes) to access the shared simulation object.

As a proof of concept, this idea has been implemented in two modules (`Diarrhoea` and `Stunting`) for the "0-day-simulations" that appear semi-frequently across the test suite: simulations that have the same start date as end date, and a particularly low initial population (100). Two new test classes, `TestStunting_SharedSim` and `TestDiarrhoea_SharedSim` are defined, which inherit from the base class.

## Complications due to simulation sharing

There are a number of complications regarding sharing simulation objects.
The most obvious is that the shared simulation object has to register all the modules (and their dependencies) that then wish to hook into this shared simulation to run tests. This isn't particularly problematic beyond remembering to adjust the dependencies if a new module is created and wishes to hook into the shared object.

### Multiple Persistent Simulation Objects

If we want several _different_ persistent simulation objects, we will need a `_BaseSharedSim` class for each one (this is due to how pytest handles setup and teardown fixtures). The current implementation has been designed with this possible generalisation in mind; everything except the `_attach_to_shared_sim` method just needs to be moved into a further abstract class - say `__AbstractSharedSim` which is then derived from:

```python
@pytest.fixture(scope="session")
def shared_sim_1():
    # foo

@pytest.fixture(scope="session")
def shared_sim_2():
    # bar

class __AbstractSharedSim:
    # Everything in the current _BaseSharedSim except _attach_to_shared_sim

class _BaseShareSim1(__AbstractSharedSim):
    # Inherit from this to have access to the shared_sim_1 object
    @pytest.fixture(scope="class")
    def _attached_to_shared_sim(self, shared_sim_1):
        self.sim = shared_sim_1


class _BaseShareSim2(__AbstractSharedSim):
    # Inherit from this to have access to the shared_sim_2 object
    @pytest.fixture(scope="class")
    def _attached_to_shared_sim(self, shared_sim_2):
        self.sim = shared_sim_2
```

### Why is there a need for the `_BaseSharedSim` class at all given the `small_shared_sim` fixture is session-scoped?

Technically there isn't, but:
- It's convenient for bundling all the functionality concerning sharing simulations into a single place.
- It _will_ reduce code bloat if we do decide to create multiple persistent simulation objects. In particular because we will not have to redefine the "setup/teardown" fixtures that are discussed in reference to "changing simulation properties" below.
- It's easier to inherit a class (and a bunch of features that come with said class) than to manually type out the name of a bunch of global fixtures each time we want to use them.
- `pytest` internals are a mystery and this was the only way I could get this to work.

### Changing Simulation Properties / State

A lot of tests need a simulation object for the sole reason that they want to check if a particular event is scheduled under certain conditions.
Setting up these conditions can involve a number of changes to the simulation object, including:
- Adjusting module parameters.
- Adjusting (even clearing) the HSI queue or simulation event queue.
- Adjusting a row in the population dataframe to artificially introduce symptoms.

As such, it is necessary for us to have "setup/teardown" fixtures that allow us to restore any potential edits to the shared object state. This is cheaper than _recreating_ the simulation object from scratch, but still introduces some overhead. It does have the effect of forcing each test to highlight which aspects of the simulation object it expects to change though, which is quite useful from a debugging perspective.

These safe "setup/teardown" fixtures are defined in the `_BaseSharedSim` class so that they can be inherited by module-level test classes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 """Collection of shared fixtures"""
 from __future__ import annotations
+from copy import copy
 import os
 from pathlib import Path
 from typing import TYPE_CHECKING, List
@@ -9,10 +10,14 @@ import pytest
 from tlo import Date, Module, Simulation
 from tlo.methods import (
     demography,
+    diarrhoea,
     enhanced_lifestyle,
+    healthburden,
+    healthseekingbehaviour,
     healthsystem,
     simplified_births,
     stunting,
+    symptommanager
 )
 
 DEFAULT_SEED = 83563095832589325021
@@ -77,13 +82,20 @@ def small_shared_sim(seed, jan_1st_2010, resource_filepath):
     sim = Simulation(start_date=jan_1st_2010, seed=seed)
     sim.register(
         demography.Demography(resourcefilepath=resource_filepath),
+        diarrhoea.Diarrhoea(resourcefilepath=resource_filepath, do_checks=True),
+        diarrhoea.DiarrhoeaPropertiesOfOtherModules(),
         enhanced_lifestyle.Lifestyle(resourcefilepath=resource_filepath),
+        healthburden.HealthBurden(resourcefilepath=resource_filepath),
+        healthseekingbehaviour.HealthSeekingBehaviour(
+            resourcefilepath=resource_filepath,
+            force_any_symptom_to_lead_to_healthcareseeking=True,
+        ),
         healthsystem.HealthSystem(
             resourcefilepath=resource_filepath, cons_availability="all"
         ),
         simplified_births.SimplifiedBirths(resourcefilepath=resource_filepath),
         stunting.Stunting(resourcefilepath=resource_filepath),
-        stunting.StuntingPropertiesOfOtherModules(),
+        symptommanager.SymptomManager(resourcefilepath=resource_filepath),
     )
     sim.make_initial_population(n=100)
     sim.simulate(end_date=sim.start_date)
@@ -155,6 +167,18 @@ class _BaseSharedSim:
         simulation fixture defined above. This ensures that every test
         is run with the persistent simulation object in its context.
 
+        NOTE: this is not strictly necessary in the current implementation
+        where we only have one simulation object to share; as we could
+        just pass the shared_small_sim explicitly to every test in the
+        (derived) class.
+        However, it does make accessing the simulation much more similar
+        to the main codebase (via self.sim.XXX rather than
+        shared_small_sim.XXX) and means we save on explicitly passing the
+        same fixture to a lot of tests since we do it automatically.
+        If we later define another simulation object that we want to share
+        between another set of tests, we can re-use this base class for
+        that purpose too, further saving on code repetition.
+
         WARNING: Writes to the shared simulation object will thus be
         persistent between tests! If a test needs to modify module
         parameters, use a fresh HSI queue, or similar, use the setup
@@ -168,14 +192,34 @@ class _BaseSharedSim:
         Flags this test as needing to clear the HSI event queue
         in the shared simulation.
 
-        Using this fixture will cause pytest to cache the current
-        HSI queue of the shared simulation, clear the queue, run
-        the test, then restore the old queue during test teardown.
+        Using this fixture will cause pytest to:
+        - Cache the current HSI queue of the shared simulation,
+        - Clear the queue,
+        - Run the test,
+        - Restore the old queue during test teardown.
+        The queue can safely be manually cleared again during the
+        test if this is necessary (EG if testing two calls to the
+        HSI scheduler).
         """
         cached_queue = list(self.shared_sim_healthsystem.HSI_EVENT_QUEUE)
         self.shared_sim_healthsystem.reset_queue()
         yield
         self.shared_sim_healthsystem.HSI_EVENT_QUEUE = list(cached_queue)
+
+    @pytest.fixture(scope="function")
+    def changes_event_queue(self):
+        """
+        Flags the test as needing to change the simulation
+        event queue, normally to check that certain events
+        have been scheduled based on treatment routines.
+
+        Using this fixture will cause pytest to cache the
+        event queue prior to running the test, then restore
+        the event queue to this state at the end of the test.
+        """
+        old_event_queue = copy(self.sim.event_queue)
+        yield
+        self.sim.event_queue = old_event_queue
 
     @pytest.fixture(scope="function")
     def changes_module_properties(self):
@@ -207,3 +251,18 @@ class _BaseSharedSim:
         cached_values = self.shared_sim_df.loc[patient_id].copy()
         yield
         self.shared_sim_df.loc[patient_id] = cached_values
+
+    @pytest.fixture(scope="function")
+    def changes_sim_date(self):
+        """
+        Flags this test as needing to manually change the date of the
+        shared simulation; typically needed when testing cures or deaths
+        that are scheduled then occur.
+
+        Using this fixture will cause pytest to cache the date of the
+        shared simulation prior to running the test, then restore the
+        simulation to that date during teardown.
+        """
+        old_date = Date(self.sim.date)
+        yield
+        self.sim.date = Date(old_date)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,8 @@
 """Collection of shared fixtures"""
 from __future__ import annotations
-from copy import copy
+
 import os
+from copy import copy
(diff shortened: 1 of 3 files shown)


willGraham01:
# Performance Summary

- Pre shared simulation figures are quoted from [5c5bb25a2](https://github.com/UCL/TLOmodel/commit/5c5bb25a247fdf9fe49efc0599c73f20b6e2c580), run locally.
- Post shared simulation figures are quoted from [3501a70](https://github.com/UCL/TLOmodel/pull/1324/commits/3501a707b652668ea09435525236f79475e11e34), run locally.

For the implementation on this branch, durations <0.01s are not listed, but _are_ broken down into `setup`, `call`, and `teardown` sections. This is what the "worst case maximum" column is accounting for in relation to the "post shared sim" column.

So for example, a "post" time of `1.0 (call) + 0.02 (teardown)` could take as long as `1.23s` to execute, since the setup step might have taken 0.01 seconds or less. Similarly, a figure of `0.02 (setup)` could correspond to a total execution time of `0.04s` (0.02 from setup plus just under 0.01 for each of the unlisted call and teardown).

Test names have been truncated so that the table formats properly, see the list below for full names (in row order).

| Name of test | Module | Pre shared sim | Post shared sim | Worst case max |
|-|-|-|-|-|
| `stunted_and_correctly_diagnosed` | Stunting | 1.34 | 1.60 (setup) | 1.62 |
| `stunted_but_no_checking` | Stunting | 1.29 | - | 0.03 |
| `not_stunted` | Stunting | 1.33 | - | 0.03 |
| `diarrhoea_severe_dehydration` | Diarrhoea | 1.93 | 2.18 (setup) + 0.01 (teardown) | 2.20 |
| `diarrhoea_severe_dehydration_dxtest_notfunctional` | Diarrhoea | 2.08 | 0.01 (teardown) | 0.03 |
| `diarrhoea_non_severe_dehydration` | Diarrhoea | 1.90 | 0.01 (teardown) | 0.03 |
| `test_run_each_of_the_HSI` | Diarrhoea | 2.00 | 0.01 (call) + 0.01 (teardown) | 0.03 |
| `test_effect_of_vaccine` | Diarrhoea | 2.86 | - | 0.03 |
| `perfect_treatment_leads_to_zero_risk_of_death` | Diarrhoea | 2.82 | 0.06 (call) | 0.08 |
| `treatment_for_those_that_will_not_die` | Diarrhoea | 3.02 | 0.02 (call) + 0.01 (teardown) | 0.04 |

## Conclusions

- The new implementation is slower for the "first" test in each of the `Test` classes, almost surely due to the overheads in setting up the test class.
- After this initially slower test, the remaining tests that use the same simulation are much faster. Almost surely because the tests which are refactored to use the same simulation do not have to re-make the simulation object each time!

As such, there appears to be a net time save to be obtained from sharing simulations.

### Full test names

1. `test_routine_assessment_for_chronic_undernutrition_if_stunted_and_correctly_diagnosed`
1. `test_routine_assessment_for_chronic_undernutrition_if_stunted_but_no_checking`
1. `test_routine_assessment_for_chronic_undernutrition_if_not_stunted`
1. `test_do_when_presentation_with_diarrhoea_severe_dehydration`
1. `test_do_when_presentation_with_diarrhoea_severe_dehydration_dxtest_notfunctional`
1. `test_do_when_presentation_with_diarrhoea_non_severe_dehydration`
1. `test_run_each_of_the_HSI`
1. `test_effect_of_vaccine`
1. `test_check_perfect_treatment_leads_to_zero_risk_of_death`
1. `test_do_treatment_for_those_that_will_not_die`

tests/conftest.py
```diff
@@ -34,4 +50,220 @@ def pytest_collection_modifyitems(config, items):
...
+        simplified_births.SimplifiedBirths(resourcefilepath=resource_filepath),
+        stunting.Stunting(resourcefilepath=resource_filepath),
+        symptommanager.SymptomManager(resourcefilepath=resource_filepath),
+    )
+    sim.make_initial_population(n=100)
+    sim.simulate(end_date=sim.start_date)
+    return sim
+
+
+class _BaseSharedSim:
```
matt-graham, line 106: The idea of using fixtures to safely mediate updates to the shared simulation fixture is cool, I hadn't come across the use of `yield` in fixtures to allow setup / teardown before. 

What I'm less sure of is if we need the `_BaseSharedSim` class and corresponding subclasses? We could do something similar with just test functions, for example

```Python
@pytest.fixture(scope="session")
def start_date() -> Date:
    return Date(2010, 1, 1)

@pytest.fixture(scope="session")
def resource_filepath() -> Path:
    return (Path(os.path.dirname(__file__)) / "../resources").resolve()

@pytest.fixture(scope="session")
def shared_sim(seed, start_date, resource_filepath):
    ...

@pytest.fixture
def clear_shared_sim_hsi_event_queue(shared_sim):
    cached_queue = shared_sim_healthsystem.HSI_EVENT_QUEUE.copy()
    shared_sim.modules["HealthSystem"].reset_queue()
    yield
    shared_sim.modules["HealthSystem"].HSI_EVENT_QUEUE = cached_queue

@pytest.fixture
def changes_shared_sim_event_queue(shared_sim):
    cached_event_queue = shared_sim.event_queue.copy()
    yield
    shared_sim.event_queue = cached_event_queue

@pytest.fixture
def changes_shared_sim_module_parameters(shared_sim, changed_module_name):
    cached_parameter_values = shared_sim.modules[changed_module_name].parameters.copy()
    yield
    shared_sim.modules[changed_module_name] = cached_parameter_values 

@pytest.fixture
def changes_shared_sim_person_properties(shared_sim, changed_person_id):
    cached_values = shared_sim_df.loc[changed_person_id].copy()
    yield
    shared_sim_df.loc[changed_person_id] = cached_values
```

This assumes a `changed_module_name` / `changed_person_id` fixture would be defined in the scope of the test function using the `changes_shared_sim_module_parameters`  / `changes_shared_sim_person_properties` fixture, which would work when we want to use the same module / person ID values across a test module (or test class). If we wanted to also be able to set these at a per test function level we could use [the ability to use marks to pass data to fixtures ](https://docs.pytest.org/en/6.2.x/fixture.html#using-markers-to-pass-data-to-fixtures) and have defaults for the fixtures like

```Python
@pytest.fixture
def changed_module_name(request):
    return request.node.get_closest_marker("changed_module_name").args[0]

@pytest.fixture
def changed_person_id(request):
    return request.node.get_closest_marker("changed_person_id").args[0]
```

with an example test function usage then something like

```Python
@pytest.mark.changed_module_name("Diarrhoea")
@pytest.mark.changed_person_id(0)
def test_do_when_presentation_with_diarrhoea_severe_dehydration(
    shared_sim,
    changed_person_id,
    changes_shared_sim_module_parameters,
    changes_shared_sim_person_properties,
    clear_shared_sim_hsi_event_queue,
):
    ...
```

matt-graham: Also if we do keep the base class approach, I would say we don't want to use an underscrore prefixed name as it is intended for use outside this module?

matt-graham (commented):
Thanks @willGraham01 for your work on this, there are some interesting ideas here!

I've added some initial thoughts and comments. Overall I think the idea of providing standardized infrastructure for allowing sharing simulations across tests safely when the test functions involves manipulating the simulation object is nice. If we could avoid requiring refactoring test functions into test classes I think that would be good, as we generally have not used test classes, so this would be both quite a lot code refactoring to apply to existing tests and also make it less familiar for existing users when adding new tests. 

Also as our biggest test computational burden are the GitHub Actions runs which run a matrix with one test module per runner, a session scoped shared simulation fixture will in reality have the same computational burden as a module scoped one. This to some degree limits how much computational gain we could get from this sort of approach, but also means we don't necessarily need to find a 'lowest common denominator' shared simulation across multiple modules, but can instead just look at sharing simulations within a single module, which presumably should be easier to manage in terms of specifying the `Module` subclasses to register, population size etc. appropriately. In this case having standard fixtures to allow safely updating properties of a `shared_sim` fixture within a test function without persisting those changes would still be useful.

PR #1326, opened 2024-04-25
Update ScenarioSwitcher to allow changes to occur midway through the simulation.

tbhallett:
(Cherry-picked from `hallett/analysis_for_vertical_and_horizontal_progs`)

This change allows the ScenarioSwitcher to cause a change at any point in the simulation, rather than only at the beginning of the simulation.

### DIFF
--- a/src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/scenario_impact_of_healthsystem.py
+++ b/src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/scenario_impact_of_healthsystem.py
@@ -19,7 +19,7 @@ from typing import Dict
 from tlo import Date, logging
 from tlo.analysis.utils import get_parameters_for_status_quo, mix_scenarios
 from tlo.methods.fullmodel import fullmodel
-from tlo.methods.scenario_switcher import ScenarioSwitcher
+from tlo.methods.scenario_switcher import ImprovedHealthSystemAndCareSeekingScenarioSwitcher
 from tlo.scenario import BaseScenario
 
 
@@ -48,7 +48,8 @@ class ImpactOfHealthSystemAssumptions(BaseScenario):
         }
 
     def modules(self):
-        return fullmodel(resourcefilepath=self.resources) + [ScenarioSwitcher(resourcefilepath=self.resources)]
+        return fullmodel(resourcefilepath=self.resources) + [
+            ImprovedHealthSystemAndCareSeekingScenarioSwitcher(resourcefilepath=self.resources)]
 
     def draw_parameters(self, draw_number, rng):
         if draw_number < len(self._scenarios):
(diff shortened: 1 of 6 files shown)


tests/test_analysis.py
```diff
@@ -420,50 +420,68 @@ def test_mix_scenarios():
     }
 
 
-def test_scenario_switcher(seed):
+def test_improved_healthsystem_and_care_seeking_scenario_switcher(seed):
```
mnjowe, line 423: This is a great test but I feel like its not fully testing our expectation from the scenario switcher. If I have understood PR https://github.com/UCL/TLOmodel/pull/1326 quite well, the intention of this module/class is to cause changes(updating parameters) during simulation(yearly?) and not only at the start or end of the simulation. if thats the case then I could love it the more if this test could demonstrate that by  creating a dummy module with an event that runs `check_parameters` yearly for some few years and `phase_of_simulation` argument not manually set rather taken from scenario switcher. Don't know if this makes sense?

tests/test_analysis.py
```diff
@@ -420,50 +420,68 @@ def test_mix_scenarios():
...
     """Check the `ScenarioSwitcher` module can update parameter values in a manner similar to them being changed
-    directly after registration in the simulation (as would be done by the Scenario class)."""
+    directly after registration in the simulation (as would be done by the Scenario class).
+    And check that these parameters can be changed mid-way through the simulation."""
 
     sim = Simulation(start_date=Date(2010, 1, 1), seed=seed)
     sim.register(*(
-        fullmodel(resourcefilepath=resourcefilepath) + [ScenarioSwitcher(resourcefilepath=resourcefilepath)]
+        fullmodel(resourcefilepath=resourcefilepath) +
+        [ImprovedHealthSystemAndCareSeekingScenarioSwitcher(resourcefilepath=resourcefilepath)]
```
mnjowe, line 431: This could be included in full model?

tests/test_analysis.py
```diff
@@ -420,50 +420,68 @@ def test_mix_scenarios():
...
+        """Check that the paramter values in the simulation currently match expectation of that phase of the
+        simulation (0=first phase before the change, 1=second phase after the change)."""
+
+        updated_values = get_parameters_for_improved_healthsystem_and_healthcare_seeking(
+            resourcefilepath=resourcefilepath,
+            max_healthsystem_function=max_healthsystem_function[phase_of_simulation],
+            max_healthcare_seeking=max_healthcare_seeking[phase_of_simulation]
+        )
+
+        for module, param in updated_values.items():
```
mnjowe, line 460: Can't we use the below lines to test `update_parameters` method defined in scenario switcher? I understand it is also calling same `get_parameters_for_improved_healthsystem_and_healthcare_seeking` and I feel like that will give us a clear picture of parameters if they are really changing during simulation.

tests/test_analysis.py
```diff
@@ -420,50 +420,68 @@ def test_mix_scenarios():
...
-    # Spot check for health care seeking being forced to occur for all symptoms
-    hcs = sim.modules['HealthSeekingBehaviour'].force_any_symptom_to_lead_to_healthcareseeking
-    assert isinstance(hcs, bool) and hcs
+    # Check that all the parameter values in the simulation are updated to be the value expected at the beginning of
+    # the simulation.
+    check_parameters(sim, 0)
 
+    # Run the simulation until after the date of the change and then check the values have been updated.
+    sim.simulate(end_date=Date(year_of_change, 2, 1))
+    check_parameters(sim, 1)
```
mnjowe, line 484: As I have said above I would avoid manually setting `phase_of_simulation` argument. Would be great if we used `phase_of_simulation`  variable defined in scenario switcher

mnjowe (approved):
Hi @tbhallett . The PR looks great! 
Kindly find some few comments from me. Thanks

src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/scenario_impact_of_healthsystem.py
```diff
@@ -48,7 +48,8 @@ def log_configuration(self):
         }
 
     def modules(self):
-        return fullmodel(resourcefilepath=self.resources) + [ScenarioSwitcher(resourcefilepath=self.resources)]
+        return fullmodel(resourcefilepath=self.resources) + [
+            ImprovedHealthSystemAndCareSeekingScenarioSwitcher(resourcefilepath=self.resources)]
```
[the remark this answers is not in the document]
tbhallett: It's because we don't  _need_ to have `ImprovedHealthSystemAndCareSeekingScenarioSwitcher` in the `full_model` (although it wouldn't do any harm if it was, to be honest!). So, as it's not, I have to concatenate list of modules to pass in at this point.

We could add 'ScenarioSwticher' to the full model, but I think it could add complexity to the full model that we don't need.

tests/test_analysis.py
tbhallett: Good spot again - see above my response. I think I'm happy to leave it as is.

tests/test_analysis.py
tbhallett: I see what you mean and think you make a good point. Let me see if I can make a better test.

src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/scenario_impact_of_healthsystem.py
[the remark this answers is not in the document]
mnjowe: I see. You made a good choice. Indeed full model is already heavy and complex, we don have to add another complexity.

tests/test_analysis.py
tbhallett: Next commit I think does this! :-)

mnjowe (commented):
I'm happy with the changes Tim. Great work on tests. You can proceed to merging the PR. Thanks

PR #1329, opened 2024-05-07
Generate and add equipment availability data

### LINKED ISSUE #1305: Equipment Availability Estimates
We (will shortly) have a ResourceFile providing specifying the availability in the model simulation of each piece of equipment at each facility_level. 

Data to inform that ResourceFile are available from the HHFA. Indeed, @sakshimohan already has a formatted version of it.

The work is to link the equipment items in the model to those in the dataset, and to summarise the dataset into one or more scenarios that can be used in model runs.

This is a modest amount of work, and would be suitable for a student as the description of the data is useful (and publishable) in its own right.

sakshimohan:
Fixes #1305  

This PR generates availability data on equipment using HHFA 2018-19 and maps it to TLO model equipment codes.


sakshimohan:
Hi @tbhallett. I'll add the equipment availability resource file to this PR. There are a couple of steps of cleaning the data left. It should be ready by Friday. But the format will be as follows:


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">
<meta charset=utf-8>
<span style='font-variant-ligatures: normal;font-variant-caps: normal;
orphans: 2;text-align:start;widows: 2;-webkit-text-stroke-width: 0px;
text-decoration-thickness: initial;text-decoration-style: initial;text-decoration-color: initial'>


Equipment_code | Facility_Level | District | Proportion of facilities which reported   equipment as available | Proportion of facilities which reported   equipment as available & functional | Proportion of facilities which reported   equipment as available, functional, and calibrated/prepared
-- | -- | -- | -- | -- | --
1 | 1a | Lilongwe | 0.9 | 0.85 | 0.8
1 | 1b | Lilongwe | 0.8 | 0.8 | 0.8
2 | 1a | Blantyre | 0.9 | 0.9 | 0.8
2 | 1b | Blantyre | 0.7 | 0.6 | 0.5
... | ... | ... | ... | ... | ... | 


</body>

</html>

tbhallett:
> Hi @tbhallett. I'll add the equipment availability resource file to this PR. There are a couple of steps of cleaning the data left. It should be ready by Friday. But the format will be as follows:
> 
> <style> </style>
> Equipment_code	Facility_Level	District	Proportion of facilities which reported equipment as available	Proportion of facilities which reported equipment as available & functional	Proportion of facilities which reported equipment as available, functional, and calibrated/prepared
> 1	1a	Lilongwe	0.9	0.85	0.8
> 1	1b	Lilongwe	0.8	0.8	0.8
> 2	1a	Blantyre	0.9	0.9	0.8
> 2	1b	Blantyre	0.7	0.6	0.5
> ...	...	...	...	...	...

Hi @sakshimohan  and @nchagoma503  -- this is going to be brilliant. Thank you so much.

Just one comment.... for the column `District`, it would be really handy if this could match as well as possible with the District names defined in `resources/demography/ResourceFile_PopulationSize_2018Census.csv`. These are:
```python
{'Chitipa', 'Karonga', 'Nkhata Bay', 'Rumphi', 'Mzimba', 'Likoma',
'Mzuzu City', 'Kasungu', 'Nkhotakota', 'Ntchisi', 'Dowa', 'Salima',
'Lilongwe', 'Mchinji', 'Dedza', 'Ntcheu', 'Lilongwe City',
'Mangochi', 'Machinga', 'Zomba', 'Chiradzulu', 'Blantyre',
'Mwanza', 'Thyolo', 'Mulanje', 'Phalombe', 'Chikwawa', 'Nsanje',
'Balaka', 'Neno', 'Zomba City', 'Blantyre City'}
```

tbhallett:
Also, please could I suggest that the destination for the ResourceFiles that will be created to be `resources/healthsystem/infrastructure_and_equipment`

tbhallett:
Thanks so much for this @sakshimohan 

I've just done a quick change to update (ever so slightly) the name of the resulting file. 

I've tried running the script, `equipment_availability_estimation.py` but get an error on line 99: `TypeError: '>=' not supported between instances of 'float' and 'str'`.

I think this comes from the aggregation using max on the groupby object. It could be the columns that are strings need to be coerced into floats.

I assume it runs cleanly from top-to-bottom on your machine?

sakshimohan:
Hi @tbhallett. Thanks for looking at this. The script did run cleanly for me but you're right that there were two string value columns (`calibrated` and `prepared`) which were not really passing through the aggregate max command. I have removed these from the script altogether because there is very limited data on these for these to be of value (I was already not extracting them into the resource file). Hopefully the script should run on your system as well now?

tbhallett:
> Hi @tbhallett. Thanks for looking at this. The script did run cleanly for me but you're right that there were two string value columns (`calibrated` and `prepared`) which were not really passing through the aggregate max command. I have removed these from the script altogether because there is very limited data on these for these to be of value (I was already not extracting them into the resource file). Hopefully the script should run on your system as well now?
> 
> I should note that I'm still using python 3.10 not 3.11. Will update this soon!

Thanks @sakshimohan. Now runs perfectly on my machine too.

tbhallett:
Thanks again for this @sakshimohan.

I've added some stuff onto the end of your script that
* formats exported dataset into what is required for reading into the `Equipment` class
* does the extrapolation to all item_codes and facilities (Rather than having this hidden in the `Equipment` class).

I'll add some comments onto the edits about both these points.

src/tlo/methods/equipment.py
tbhallett, line 1: Moving the interpolation into the data processing scripts and now insisting that the imported data on availabiltity has an estimate for every item_code and facility_id

tests/test_equipment.py
tbhallett, line 1: Remove checks on handling of item_codes/facility_id that are not in the availability data (we now insist that they all are).

resources/healthsystem/infrastructure_and_equipment/ResourceFile_Equipment_Availability_Estimates.csv
tbhallett, line 1: recreated file from update script

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
```diff
@@ -0,0 +1,273 @@
...
+final_equipment_availability_export = final_equipment_availability_export.rename(columns={'Facility_level': 'Facility_Level'})
+
+# Check that every row specifies a facility_level
+assert final_equipment_availability_export['Facility_Level'].isin([
+    '0', '1a', '1b', '2', '3', '4', '5'
+]).all()
+assert final_equipment_availability_export['Facility_Level'].notnull().all()
+
+# - 'Pr_Available' is the name expected in the Equipment class, and is taken to be the probability that an item of equipment is available for use
+#   which is taken to be equivalent to it to being "functional"
```
tbhallett, line 163: I think we can use the 'functional' column as the final probability of piece of equipment actually being "available" _and_ "usable", right @sakshimohan?

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
```diff
@@ -0,0 +1,273 @@
...
+    right_on=['Facility_Level', 'District'],
+    how='left',
+)
+
+# Merge in region-level facilities (level 3) (matching on region only)
+
+# - First, ensure that there is only one set of facility_level 3 entries for each region. (In the current dataframe
+#   entries for level 3 of Blantyre City and Zomba City could both inform on level 3 in the Southern region.)
+#   Do this by taking the mean of Pr_Available across those rows that are facility_level 3 in the same region.
+#   Create a 'reconciled' set of data facility level 3.
```
tbhallett, line 194: As facility_level = 3 is defined for the region, not the district, we needed to do this reconciliation here. Does it seem ok to you @sakshimohan ?

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
```diff
@@ -0,0 +1,273 @@
...
+    mfl.set_index('Facility_ID')['Facility_Level'].to_dict()
+)
+final_equipment_availability_export_full = final_equipment_availability_export_full.groupby(['Item_Code', facility_levels]).transform(lambda x: x.fillna(x.mean()))
+
+# Remaining misising data include those item_codes that are never seen anywhere: interpolate these from the average availability of all other items, stratified by level
+final_equipment_availability_export_full = final_equipment_availability_export_full.groupby([facility_levels]).transform(lambda x: x.fillna(x.mean()))
+
+# Remaining missing data are for items_codes in facility_ids for which there is no information at all in the facility_level
+# Impute the availability for these items in facilities as ... the average of other items available at other facilities
+final_equipment_availability_export_full = final_equipment_availability_export_full.groupby("Item_Code").transform(lambda x: x.fillna(x.mean()))
```
tbhallett, line 256: This is my proposal for doing the extrapolation to item_codes/ facility_id that are not directly observed in the data.
Please let me know if it looks ok, or you see anything suspicious in the resulting data....?

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
sakshimohan: Yes, that makes sense.

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
sakshimohan: Ah I didn't know this was the case - noted for future analyses. The reconciliation looks good.

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
sakshimohan: This looks good. The only thing I'm concerned about is interpolating the value of cheaper items like thermometer from expensive items like X-rays which are likely to have a lower level of availability. Because equipment costing in the HSSP-III differentiates between equipment which cost <$1000 and those which cost more, we could use the same distincition to interpolate availability. Let me try to add this change.

tbhallett: Very good point! :)

If it becomes complicated to do the interpolation using these simples rules when using multiple criteria, we could always estimate a linear model for the availability, and use that to fill in the blanks;

e.g.,
availability ~ item_code + facility_level + district + price_is_high

sakshimohan: Thanks, Tim. I've pushed the update now. I just updated the interpolation method to include (binary) equipment category and facility levels (instead of just facility levels), as it seemed more straightforward that using a linear model.

sakshimohan, line 267: On the final level of interpolation which is done based on 'Item_code' alone - This is interpolating values only for facility level 5, where there is no service delivery taking place. I assume this is needed to be done for the assertions in `equipment.py` to work?  
I wonder if more reasonable options might be - 
1. Either we just drop level 5 completely from  `ResourceFile_Equipment_Availability_Estimates.csv`
2. Or we force all `Pr_Available` = 0 for`Facility_Level` == 5

src/tlo/methods/equipment.py
```diff
@@ -104,35 +104,25 @@ def availability(self, value: Literal["all", "default", "none"]):
...
+        Extracting them once and storing the result allows us to avoid repeating this
         calculation if the equipment availability change event occurs during the simulation.
         """
-        # Create "full" dataset, where we force that there is probability of availability for every item_code at every
-        # observed facility
-        dat = pd.Series(
-            index=pd.MultiIndex.from_product(
+        dat = self.data_availability.set_index(
+            [self.data_availability["Facility_ID"].astype(int), self.data_availability["Item_Code"].astype(int)]
+        )["Pr_Available"]
```
sakshimohan, line 119: We had a conversation before about whether equipment should have a probability of availability with values between 0 to 1 or just have a binary value to indicate whether each equipment is available or not at each Facility_ID.  And I believe we decided to go with the latter. @tbhallett, should we update the equipment availability with a rule like this - 
`final_equipment_availability_export_full[final_equipment_availability_export_full < 0.5] = 0
final_equipment_availability_export_full[final_equipment_availability_export_full >= 0.5] = 1`

Note that this increases the average availability of equipment from ~60% to ~ 73%.

tbhallett: Good question.

The implementation I've proposed (and which is in `master`) [see below] at the moment would need us to retain these probabilities. 

I think working on the availability of a piece of equipment probabilistically for the HSI_Event is a better way to represent that these data estimate are really telling us about the *proportion* of facilities in a particular level/district that have that piece of equipment.

https://github.com/UCL/TLOmodel/blob/67c80589a2e0f274ac12c87cd76876df3010d51d/src/tlo/methods/equipment.py#L13-L61

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
sakshimohan: Thanks Tim. I've pushed a change to implement option 2.

sakshimohan:
@tbhallett This should be ready to merge now.

src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py
```diff
@@ -0,0 +1,289 @@
...
+final_equipment_availability_export_full = final_equipment_availability_export_full.groupby([equipment_categories, facility_levels]).transform(lambda x: x.fillna(x.mean()))
+final_equipment_availability_export_full = final_equipment_availability_export_full.droplevel(level =2) # drop equipment category from the index
+
+# Force availability to be 0 for level 5, where no service delivery occurs
+level_5_lookup = mfl.loc[mfl['Facility_Level'].isin(['5'])].set_index('Facility_Level')['Facility_ID'].to_dict()
+final_equipment_availability_export_full[final_equipment_availability_export_full.index.get_level_values(level = 0) == list(level_5_lookup.values())[0]] = 0
+
+# Remaining missing data are for items_codes in facility_ids for which there is no information at all in the facility_level
+# Impute the availability for these items in facilities as ... the average of other items available at other facilities
+final_equipment_availability_export_full = final_equipment_availability_export_full.groupby("Item_Code").transform(lambda x: x.fillna(x.mean()))
```
tbhallett, line 272: I think this average should be computed from across facility levels 0-3. (Otherwise it gets dragged down by the zeros at level 5). 

I'll make that last change

sakshimohan: That's a good point. Thanks, Tim. Just as an FYI, there isn't any data being imputed by this line of code as the series is complete at this stage. But I have deleted the imputation in case any new source of data requires this step.

tbhallett (approved):

PR #1330, opened 2024-05-07
Add option to specify image tag when submitting batch tasks

### LINKED ISSUE #1328: Branches using older versions of python cannot submit jobs to azure
Discussed with @tamuri last week that branches using versions of python before 3.11 can no longer submit jobs to azure. This is an issue for ongoing paper submissions which may/have been required to do additional runs of new/updated scenarios at review process.

tamuri:
- overrides default image tag, which is set in batch configuration
- allows user to run tasks on an older image (e.g. with python 3.8 support)

Fixes #1328


giordano (approved):
Looks good to me, but do I understand correctly that the image tag could also be changed in the config file?  If so, should we recommend using that?

tamuri:
Thanks. These particular settings are read from the Azure Batch resource, not the user's personal config. I think it's better to set the run environment on the server-side, and let the user override for specific runs because e.g. they are running some old code. In the general case, they should always be using the one we set.

PR #1331, opened 2024-05-09
allocate population equally across districts at initialisation

tdm32:
This PR uses the initial specified population size (n) and produces a population dataframe which creates n entries within each district. 

The rationale is that for highly heterogeneous diseases which occur in sparsely populated districts, this offers a way to create large enough populations in each district to reach endemic equilibria without requiring huge total population sizes. 

The argument equal_allocation_by_district is added to the Simulation class which is called during make_initial_population to create the correct sized dataframe.

Demography then uses this argument to allocate the population equally across the districts, retaining the age structure from the initial population data. 

A test is created in test_demography to ensure population size is correctly allocated. 

TO DO:
One issue is that the scaling factor now needs to be calculated per district. This will be needed to scale the outputs to full population size and to scale HCW capabilities in the case that mode 2 runs are required.

### DIFF
--- a/src/tlo/methods/demography.py
+++ b/src/tlo/methods/demography.py
@@ -87,6 +87,7 @@ class Demography(Module):
         self.districts = None  # will store all the districts in a list
 
     OPTIONAL_INIT_DEPENDENCIES = {'ImprovedHealthSystemAndCareSeekingScenarioSwitcher'}
+    # <-- this forces that module to be the first registered module, if it's registered.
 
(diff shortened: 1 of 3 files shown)


tbhallett:
Thanks for this @tdm32 
I have three quick suggestions
 - make this contained wholly within demography
 - follow the factorised design with demography, so that all the new calcs in their own function
 - make the test agnostic to the number of districts defined.

I'll push a commit with both changes in a minute.

The actual guts of the calculation seem fine and very elegant using `groupby`.

tbhallett (approved):

PR #1359, opened 2024-05-17
Revert "[CI] Change label of runners for profiling workflows"

giordano:
Reverts UCL/TLOmodel#1358.  The Azure Kubernetes cluster is up and running, but it's not able to create new runners.  I need to investigate what's going on.

***Edit***: wait, maybe I found the issue.

### DIFF
--- a/.github/workflows/run-profiling.yaml
+++ b/.github/workflows/run-profiling.yaml
@@ -82,7 +82,7 @@ jobs:
     needs: set-variables
     uses: ./.github/workflows/run-on-comment.yml
     with:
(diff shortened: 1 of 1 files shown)


giordano:
Issue sorted out, I had simply created the ARC runners with a PTA targeting the wrong repository :smiling_face_with_tear:  I destroyed the old Kubernetes cluster and created a new one with the right PTA, and now this [seems to be working fine](https://github.com/UCL/TLOmodel/actions/runs/9131756469), so all is good!

PR #1361, opened 2024-05-20
`HSI_Event` uses separate private member function (`_run_after_hsi_event`) instead of `post_apply_hook`

### LINKED ISSUE #1360: Over-writing of `post_apply_hook` by some `HSI_Events`
There are some instance of disease module writing instances of `HSI_Events` with a `post_apply_hook` method defined. This will over-write the method of the same name in the base class, preventing the functionality contained therein to operate ([post_apply_hook in HSI_Event base class](https://github.com/UCL/TLOmodel/blob/67c80589a2e0f274ac12c87cd76876df3010d51d/src/tlo/methods/hsi_event.py#L174)). It was a sort of "unwritten rule" (!) that the `post_apply_hook` part of an HSI_Event is reserved for use by the HSI_Event itself.

Examples of HSI_Event instance over-writing this method are:
https://github.com/UCL/TLOmodel/blob/67c80589a2e0f274ac12c87cd76876df3010d51d/src/tlo/methods/tb.py#L2236
https://github.com/UCL/TLOmodel/blob/67c80589a2e0f274ac12c87cd76876df3010d51d/src/tlo/methods/contraception.py#L1215

There may be others...!

A simple fix would be for this logic to be placed at the end of the `apply` function in the respective HSI_Event.

But, a better fix would be allow the disease module do over-write the `post_apply_hook` method and create another method that is only called by the HealthSystem. e.g. in the `HSI_Event` base class:
* put the current content of the `post_apply_hook` into `_run_after_hsi_event`,
* over-write run (from the Event base class) to run `apply`, then `post_apply_hook`, then `_run_after_hsi_event`

tbhallett:
Fixes #1360 

Here `HSI_Event` uses a separate private member function (`_run_after_hsi_event`) rather than `post_apply_hook` to do the jobs expected after the `HSI_Event` has run.

This prevents disease modules using `HSI_Events` that over-write the `post_apply_hook` member function and so inadvertently removing the expected behaviour of the `HSI_Event`.

### DIFF
--- a/src/tlo/methods/hsi_event.py
+++ b/src/tlo/methods/hsi_event.py
@@ -171,7 +171,7 @@ class HSI_Event:
         """Called when this event is was entered to the HSI Event Queue, but was never run."""
         logger.debug(key="message", data=f"{self.__class__.__name__}: was never run.")
 
-    def post_apply_hook(self) -> None:
+    def _run_after_hsi_event(self) -> None:
         """
         Do things following the event's `apply` function running.
          * Impose the bed-days footprint (if target of the HSI is a person_id)
@@ -193,6 +193,7 @@ class HSI_Event:
         """Make the event happen."""
         updated_appt_footprint = self.apply(self.target, squeeze_factor)
         self.post_apply_hook()
+        self._run_after_hsi_event()
         return updated_appt_footprint
 
     def get_consumables(
@@ -283,7 +284,7 @@ class HSI_Event:
     def add_equipment(self, item_codes: Union[int, str, Iterable[int], Iterable[str]]) -> None:
         """Declare that piece(s) of equipment are used in this HSI_Event. Equipment items can be identified by their
         item_codes (int) or descriptors (str); a singular item or an iterable of items (either codes or descriptors but
-        not a mix of both) can be defined at once. Checks are done on the validity of the item_codes/item 
+        not a mix of both) can be defined at once. Checks are done on the validity of the item_codes/item
         descriptions and a warning issued if any are not recognised."""
         self._EQUIPMENT.update(self.healthcare_system.equipment.parse_items(item_codes))
 
--- a/src/tlo/methods/hsi_event.py
+++ b/src/tlo/methods/hsi_event.py
@@ -173,7 +173,6 @@ class HSI_Event:
 
     def post_apply_hook(self) -> None:
         """Do any required processing after apply() completes."""
-        pass
 
     def _run_after_hsi_event(self) -> None:
         """
--- a/src/tlo/methods/hsi_event.py
+++ b/src/tlo/methods/hsi_event.py
@@ -173,7 +173,7 @@ class HSI_Event:
 
     def _run_after_hsi_event(self) -> None:
         """
-        Do things following the event's `apply` function running.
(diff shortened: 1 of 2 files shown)


src/tlo/methods/hsi_event.py
```diff
@@ -283,7 +284,7 @@ def make_appt_footprint(self, dict_of_appts) -> Counter:
     def add_equipment(self, item_codes: Union[int, str, Iterable[int], Iterable[str]]) -> None:
         """Declare that piece(s) of equipment are used in this HSI_Event. Equipment items can be identified by their
         item_codes (int) or descriptors (str); a singular item or an iterable of items (either codes or descriptors but
-        not a mix of both) can be defined at once. Checks are done on the validity of the item_codes/item 
+        not a mix of both) can be defined at once. Checks are done on the validity of the item_codes/item
```
tbhallett, line 287:
```suggestion
        not a mix of both) can be defined at once. Checks are done on the validity of the item_codes/item 
```

tbhallett: This is an inadvertent change, but I can't seem to make a commit to roll it back (git errors saying "cannot commit"). So, I think I'm stuck with this little change hitchhiking its way on this PR.

src/tlo/methods/hsi_event.py
```diff
@@ -171,9 +171,13 @@ def never_ran(self) -> None:
         """Called when this event is was entered to the HSI Event Queue, but was never run."""
         logger.debug(key="message", data=f"{self.__class__.__name__}: was never run.")
 
-    def post_apply_hook(self) -> None:
+    def post_apply_hook(self):
```
tbhallett, line 174:
```suggestion
    def post_apply_hook(self) -> None:
```

src/tlo/methods/hsi_event.py
```diff
@@ -172,8 +172,12 @@ def never_ran(self) -> None:
         logger.debug(key="message", data=f"{self.__class__.__name__}: was never run.")
 
     def post_apply_hook(self) -> None:
+        """Do any required processing after apply() completes."""
+        pass
```
matt-graham, line 176:
```suggestion
```
[The `pass` is not strictly required here as the docstring suffices to make this syntactically correct as a function definition by having an indented block](https://docs.astral.sh/ruff/rules/unnecessary-placeholder/).

src/tlo/methods/hsi_event.py
matt-graham: This is possibly just your editor autostripping whitespace at ends of line, which is generally not a bad thing so I think fine to leave, but thank you for flagging!

matt-graham (approved):
This looks good to me! Have left one very minor suggestion with regards to moving unnecessary `pass` statement but this is just a nitpick.

src/tlo/methods/hsi_event.py
tbhallett: Okay thanks for confirming.

tbhallett:
> This looks good to me! Have left one very minor suggestion with regards to moving unnecessary `pass` statement but this is just a nitpick.

Thanks very much. And I'm always grateful for sharped-eyed "nitpicking"!

PR #1298, opened 2024-03-13
Editing units of consumable required for each HSI

sakshimohan:
There are a few challenges with the way in which units of consumables required are coded in modules - 

1. Inconsistency in coding the quantity of consumables needed - Some modules express the units needed in terms of the proportion of the unit of issue (package sold) as per the name of the consumable in the original consumable name list from the 2016 One Health Tool. For instance, see line 1040 of `alri.py` where 0.01 of a package is required per HSI. Others specify these in terms of the number of individual units (smallest divisible unit). For instance, the contraception module assume *30* condoms issued per HSI. 

2. Using  the unit of issue as defined in the 2016 One Health Tool might lead to errors in costing especially if these units of issue change over time and costs are expressed in terms of a different unit of issue. For eg., costs expressed in terms of $/pack of 30 tablets whereas module coded to provide quantities assuming a pack  of 60 tablets. 

In order to address the inconsistency and ensure accuracy of costing, we wish to recode consumable quantities required by HSIs within modules to represent the quantity of the chosen unit of the consumable in terms of which price is expressed, such as a single vial, single ampoule, 1 gram of drug, 1ml of liquid etc. Some examples below - 

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Consumable_name_tlo | Chosen_unit_for_price | Final_price_per_chosen_unit   (USD, 2023)
-- | -- | --
Gentamycin,   injection, 40 mg/ml in 2 ml vial | 2 ml vial (40 mg/ml) | 0.053693068  | 
Ciprofloxacin   250mg_100_CMST | 1 mg | 0.000058701 | 
[...]

PR #1279, opened 2024-02-20
Costing Module

sakshimohan:
This PR adds a costing module to the TLO model

PR #1362, opened 2024-05-20
Editing units of consumable required: Contraception module

EvaJanouskova:
units per case updated to relate to the chosen units in #1298 for consumables in contraception module
(and item codes get from item names instead of pkg name, hence numbers of units per case defined within the code)

### DIFF
--- a/src/tlo/methods/contraception.py
+++ b/src/tlo/methods/contraception.py
@@ -155,6 +155,7 @@ class Contraception(Module):
 
         self.processed_params = dict()  # (Will store the processed data for rates/probabilities of outcomes).
         self.cons_codes = dict()  # (Will store the consumables codes for use in the HSI)
+        self.cons_alternatives_probs = dict()  # (Will store the probabilities of use for consumable alternatives)
         self.rng2 = None  # (Will be a second random number generator, used for things to do with scheduling HSI)
 
         self._women_ids_sterilized_below30 = set()  # The ids of women who had female sterilization initiated when they
@@ -257,9 +258,10 @@ class Contraception(Module):
         # Schedule first occurrences of Contraception Poll to occur at the beginning of the simulation
         sim.schedule_event(ContraceptionPoll(self, run_update_contraceptive=self.run_update_contraceptive), sim.date)
 
-        # Retrieve the consumables codes for the consumables used
+        # Retrieve the consumables codes for the consumables used and probabilities for alternatives
         if self.use_healthsystem:
-            self.cons_codes = self.get_item_code_for_each_contraceptive()
+            self.cons_codes, self.cons_alternatives_probs =\
+                self.get_item_codes_for_each_contraceptive_and_probs_for_cons_alternatives()
 
         # Create second random number generator
         self.rng2 = np.random.RandomState(self.rng.randint(2 ** 31 - 1))
@@ -644,11 +646,13 @@ class Contraception(Module):
         # Do the change in contraceptive
         self.schedule_batch_of_contraceptive_changes(ids=[mother_id], old=['not_using'], new=[new_contraceptive])
 
-    def get_item_code_for_each_contraceptive(self):
-        """Get the item_code and numbers of units per case for each contraceptive and for contraceptive initiation."""
+    def get_item_codes_for_each_contraceptive_and_probs_for_cons_alternatives(self):
+        """Get the item_code and numbers of units per case for each contraceptive and for contraceptive initiation.
+        Save the probs for consumable alternatives."""
         # TODO: update with optional items (currently all considered essential)
 
         # ### Get item codes and number of units per case from package name
+        # (note: this version does not deal with consumable alternatives)
         # get_items_from_pkg = self.sim.modules['HealthSystem'].get_item_codes_from_package_name
 
         # _cons_codes = dict()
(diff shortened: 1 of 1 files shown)


EvaJanouskova:
@tbhallett, @sakshimohan, Some `# TODO: @TimH` and `# TODO: @Sakshi` in the code to be addressed before merged into #1298.

src/tlo/methods/contraception.py
```diff
@@ -645,24 +645,89 @@ def select_contraceptive_following_birth(self, mother_id, mother_age):
...
-        _cons_codes['injections'] = get_items_from_pkg('Injectable')
-        _cons_codes['implant'] = get_items_from_pkg('Implant')
-        _cons_codes['female_sterilization'] = get_items_from_pkg('Female sterilization')
+        # # items for each method that requires an HSI to switch to
+        # TODO: @TimH Pills are alternatives, we weight their use by the probabilities (as approximation of use in 80:20
+        #  cases, for each case 0.8 and 0.2 proportion of units is used). Is that okay, or do we want to choose for each
+        #  case only one type of pills? Total usage should be approximately same.
+        # TODO: @Sakshi what are the numbers of pills in monthly packets for both following items?
+        _cons_codes['pill'] =\
+            {get_item_code("Levonorgestrel 0.0375 mg, cycle"): 28*3.75*0.2,  # progesterone-only pills used in 20% cases
```
tbhallett, line 678: This will make any checks on availability require BOTH to be available. Probably better to choose one here OR chose between these probabilistically at point of dispensation in hsi rather than like this.

src/tlo/methods/contraception.py
```diff
@@ -645,24 +645,89 @@ def select_contraceptive_following_birth(self, mother_id, mother_age):
...
+             }
+        _cons_codes['male_condom'] =\
+            {get_item_code("Condom, male"): 30}
+        _cons_codes['other_modern'] =\
+            {get_item_code("Female Condom_Each_CMST"): 30}
+        _cons_codes['IUD'] =\
+            {get_item_code("Glove disposable powdered latex medium_100_CMST"): 2,
+             get_item_code("IUD, Copper T-380A"): 1}
+        # TODO: originally used "Medroxyprogesterone acetate injection 150mg/mL, 1mL vial with 2ml syringe with 22g 0.7
+        #  X 25mm needle_each_CMST", which is not within the RF_Costing anymore, does it mean we need to add a syringe
```
tbhallett, line 690: These look very similar to me. In principle, you can use items from the original consumables resource file (don't have to limit to the costing file, right Sakshi?)

src/tlo/methods/contraception.py
```diff
@@ -645,24 +645,89 @@ def select_contraceptive_following_birth(self, mother_id, mother_age):
...
+            {get_item_code("Depot-Medroxyprogesterone Acetate 150 mg - 3 monthly"): 1,
+             get_item_code("Glove disposable powdered latex medium_100_CMST"): 1,
+             get_item_code("Water for injection, 10ml_Each_CMST"): 1,
+             get_item_code("Povidone iodine, solution, 10 %, 5 ml per injection"): 5,
+             get_item_code("Gauze, swabs 8-ply 10cm x 10cm_100_CMST"): 1}
+        # TODO: trocar may be reusable (we used 0.1 units, indicating 10 uses on average) - need to find out for what
+        #  kind of trocar (reusable or not) we have the cost
+        # Jadelle & Implanon are alternatives, we weight their use by the probabilities (as approximation of use in
+        # 50:50 cases, for each case 0.5 and 0.5 proportion of units is used)
+        # TODO: @TimH Is that okay, or do we want to choose for each case only one type of implant? Total usage should
```
tbhallett, line 702: As above, doing this way will require both items to be available in any checks on availability. So I would recommend choosing one here.

src/tlo/methods/contraception.py
EvaJanouskova: @tbhallett, Good point, the alternatives are now chosen by probs, including the item Atropine that is used only in 50% cases (with no altrenative).

src/tlo/methods/contraception.py
EvaJanouskova: @sakshimohan, what is better for you if I use? The one which is there now--Medroxyprogesterone Acetate only--, or the one including syringe and needle? If the former, do I need to add a syringe and a needle or is it covered within the cost of "Depot-Medroxyprogesterone Acetate 150 mg - 3 monthly" even if not included in the name? The "Medroxyprogesterone acetate injection 150mg/mL, 1mL vial with 2ml syringe with 22g 0.7 X 25mm needle_Each_BB049500_CMST" is in the `RF_Costing/consumables` in the column `Item_EHP` in the line with item "Depot-Medroxyprogesterone Acetate 150 mg - 3 monthly".

sakshimohan: Hi @EvaJanouskova. I've current got the cost of this item `Medroxyprogesterone acetate injection 150mg/mL, 1mL vial with 2ml syringe with 22g 0.7 X 25mm needle_Each_BB049500_CMST` from CMST which already includes the syringe and injection. This is for `item_code` = 3 in the model. So perhaps easiest to include the one including syringe and needle?

sakshimohan: Sorry if I'm missing something here Eva. The name of the consumable for item_code 3 in ResourceFile_Items_and_Packages  is "Medroxyprogesterone acetate injection 150mg/mL, 1mL vial with 2ml syringe with 22g 0.7 X 25mm needle_Each_BB049500_CMST, ampule". For item_code 3, there is availability data in in ResourceFile_Consumables_availability_small and there is cost data in ResourceFile_Costing.
The cost for this item is based on the item with the same name in the CMST costing catalogue which includes the cost of the syringe and needle.
I'm not sure why the name of the consumable in ResourceFile_Costing appears to be Depot-Medroxyprogesterone Acetate 150 mg - 3 monthly but in either case, it's fine for you to go ahead and use item_code 3 for your purposes. If in your scrip for the contraception.py module, you're calling teh consumable by its name, this will be obtained from  ResourceFile_Items_and_Packages so you can ignore the name in ResourceFile_Costing .

sakshimohan: So the preference is for you to use item code 3. However, if for any other reason, you wish to use item code 1839, I will need to make a few changes to make sure that there is availability data for 1839 in `ResourceFile_Consumables_availability_small` and add cost data to `ResourceFile_Costing`.

sakshimohan:
> @tbhallett, @sakshimohan, Some `# TODO: @TimH` and `# TODO: @Sakshi` in the code to be addressed before merged into #1298.

Hi Eva. I've updated the unit `Gauze, swabs 8-ply 10cm x 10cm_100_FF010800_CMST` to 1 swab in the costing PR - see commit [5d2c813](https://github.com/UCL/TLOmodel/pull/1279/commits/5d2c8138c4034d8be1f85370a14d15aa5c221033)

PR #1367, opened 2024-05-22
Health Impact of Realistic Consumable Availability Scenarios

sakshimohan:
This PR estimates the health impact of increasing consumable availability to realistic levels. Realistic levels lie somewhere between `default` and `all` as predicted by the regression model or by assuming that all facilities perform as well as the facility at the nth percentile in terms of consumable availability. 

Changes:
1. Add scripts to prepare HHFA data and run regression analysis to determine factors associated with consumable availability
2. Update `ResourceFile_consumables_amatched.csv` to include a crosswalk between names of consumables in the TLO model and that under HHFA
3. Predict change in consumable availability under various scenarios
4. Change `healthsystem.py` and `consumables.py` to accommodate specification of which consumable availability estimates to use
5. Introduce a scenario which runs the model with a different set of availability estimates from the original estimates.

### DIFF
--- a/src/scripts/data_file_processing/healthsystem/consumables/consumable_resource_analyses_with_hhfa/prepare_hhfa_consumables_data_for_inferential_analysis.py
+++ b/src/scripts/data_file_processing/healthsystem/consumables/consumable_resource_analyses_with_hhfa/prepare_hhfa_consumables_data_for_inferential_analysis.py
@@ -1,0 +1,811 @@
+"""
+This script prepares HHFA data on consumable availability for regression analysis on R.
+
+Inputs:
+1. Raw HHFA data - Q1.dta (~Dropbox/Thanzi la Onse/07 - Data/HHFA_2018-19/0 raw/2_Final data/)
+2. Cleaned variable names for HHFA data - variable_list.csv (~Dropbox/Thanzi la Onse/07 - Data/HHFA_2018-19/1 processing)
+3. Relevant distance calculations for regression analysis - facility_distances_hhfa.csv (~Dropbox/Thanzi la Onse/07 - Data/HHFA_2018-19/2 clean/)
+4. Consumable categorisations - items_hhfa.xlsx (~Dropbox/Thanzi la Onse/07 - Data/HHFA_2018-19/1 processing/)
+
+Outputs:
+1. Cleaned consumable availability data ready for regression analysis - cleaned_hhfa_2019.csv (~Dropbox/Thanzi la Onse/07 - Data/HHFA_2018-19/2 clean/)
+
+Consumable availability is measured as probability of stockout at any point in time.
+"""
+
+import calendar
+import datetime
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from tabulate import tabulate
+import copy
+
+from collections import defaultdict
+
+# Set local Dropbox source
+path_to_dropbox = Path(  # <-- point to the TLO dropbox locally
+     '/Users/sm2511/Dropbox/Thanzi la Onse'
+)
+
+path_to_files_in_the_tlo_dropbox = path_to_dropbox / "07 - Data/HHFA_2018-19/" # <-- point to HHFA data folder in dropbox
+
+# define a timestamp for script outputs
+timestamp = datetime.datetime.now().strftime("_%Y_%m_%d_%H_%M")
+
+# print the start time of the script
+print('Script Start', datetime.datetime.now().strftime('%H:%M'))
+
+## 1. DATA IMPORT ##
+#########################################################################################
+raw_hhfa = pd.read_csv(path_to_files_in_the_tlo_dropbox / '0 raw/2_Final data/Q1.csv',
+                       low_memory=False)  # import 2018 data
+varnames = pd.read_csv(path_to_files_in_the_tlo_dropbox / '1 processing/variable_list.csv',
+                       encoding="ISO-8859-1")  # import file with cleaned variable names
+
+# Rename HHFA columns using variable name mapping in loaded .csv
+old_var_name = varnames['var']
+new_var_name = varnames['new_var_name']
+
+hhfa = copy.deepcopy(raw_hhfa)
+for i in range(len(new_var_name)):
+    if new_var_name[i] != np.nan:
+        hhfa.rename(columns={old_var_name[i]: new_var_name[i]},
+                    inplace=True)
+    else:
+        pass
+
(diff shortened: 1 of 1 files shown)


sakshimohan:
Current consumable availability under various scenarios. All scenarios only affect levels 1a and 1b:
1. Scenario 1: All items perform as well as **consumables other than drugs/diagnostic tests**
2. Scenario 2: 1 + All items perform as well as consumables classified as **'Vital' in the Essential Medicines List**
3. Scenario 3: 2 + All facilities perform as well as those in which consumables stock is managed by **pharmacists**
4. Scenario 4: 3 + Level 1a facilities perform as well as **level 1b**
5. Scenario 5: 4 + All facilities perform as well as **CHAM** facilities
6. Scenario 6: All facilities have the same probability of consumable availability as the **75th percentile** best performing facility for each individual item
7. Scenario 7: All facilities have the same probability of consumable availability as the **90th percentile** best performing facility for each individual item
8. Scenario 8: All facilities have the same probability of consumable availability as the **99th percentile** best performing facility for each individual item


availability_by_level
availability_by_category

sakshimohan:
Overall change in availability - 
available_prop                       **0.49289695774871195**
available_prop_scenario1    **0.520910**
available_prop_scenario2    **0.523982**
available_prop_scenario3    **0.527928**
available_prop_scenario4    **0.529791**
available_prop_scenario5    **0.531345**
available_prop_scenario6    **0.513739**
available_prop_scenario7    **0.540825**
available_prop_scenario8    **0.575753**

sakshimohan:
Hi @tbhallett. Following our discussion of whether to add any scenarios for the FCDO analysis, this branch contains the consumable availability scenarios which I used to run the analysis I presented at the Think Tank. The list of scenarios and their description is provided here - https://github.com/UCL/TLOmodel/pull/1367#issuecomment-2152722059

One change I have made since last week is updating scenarios 10-12 so that the final probability of availability is the maximum or minimum of the original value and the average of the relevant high- (or low) performing supply chain, as opposed to simply the latter regardless of the original probability.

sakshimohan:
Hi @marghe-molaro

sakshimohan:
Hi @marghe-molaro. Responding to your question on slack here. 

> Can I double check what the assumptions for other consumables are (non-HIV) in scenario_12? I'm concerned in > particular about TB consumables, because these might have spill-over effects into HIV/AIDS DALYs. If we are assuming that all vertical programmes are transitioning from vertical to govt funding then  it would be more consistent to assume that TB consumables are also similarly affected (and maybe also malaria + epi)? We could also consider them in combination (e.g. all of them govt funded availability, HIV/AIDS govt funded but TB/malaria/epi vertically funded availability, etc)

As you'll see Figure 1 in the comment here (https://github.com/UCL/TLOmodel/pull/1367#issuecomment-2152722059), I have only updated the availability of HIV consumables to be equal to Minimum(original availability, average availability of consumables other than HIV, EPI and Cancer* (at the corresponding Facility_Level)).
The availability of TB, malaria and EPI consumables is assumed to be the same (See `Actual` and `HIV moved to Govt. Supply Chain` columns in the heatmap).
 
I see your point about updating TB and malaria consumables as well. But the challenge is that it is unclear to me whether the distribution supply chain for these is actually parallel. The supply chain can be parallel either for **procurement** or **distribution**. Parallel procurement exists across several programs - for eg. UNICEF procures IMCI and nutrition drugs. Parallel distribution is more limited. I have only been able to confirm parallel distribution (including maintenance of stock records) for HIV and EPI consumables and, hence, I had limited the scenario only to HIV (considering the perspective of Global Fund) - EPI consumables are managed by UNICEF as I understand it. 
Happy to take your advice here if you think the scenario needs updating. 

References - https://www.theglobalfund.org/media/9060/oig_gf-oig-19-024_report_en.pdf;
https://www.refworld.org/reference/countryrep/odi/2013/en/93985

tbhallett:
> Hi @tbhallett. Following our discussion of whether to add any scenarios for the FCDO analysis, this branch contains the consumable availability scenarios which I used to run the analysis I presented at the Think Tank. The list of scenarios and their description is provided here - [#1367 (comment)](https://github.com/UCL/TLOmodel/pull/1367#issuecomment-2152722059)
> 
> One change I have made since last week is updating scenarios 10-12 so that the final probability of availability is the maximum or minimum of the original value and the average of the relevant high- (or low) performing supply chain, as opposed to simply the latter regardless of the original probability.

Thnaks VERY much @sakshimohan. Tagging in @tdm32 as she will incorporate these scenarios into her analyses.

sakshimohan:
> Hi @sakshimohan , thanks so much for producing these. For the PR, would it be possible to rename the scenarios for the cons_availability argument with an informative name, such that they can be called from a script without needing to reference a list of definitions somewhere?

Hi @tdm32. That's definitely what I plan to do. Would it be helpful if I make this change soon or should I prioritise it after I have made more progress with costing in the coming week?
It seems like a quick fix but because of the way the scenarios are generated and coded across different scripts, I might need an afternoon to implement and test the changes.

sakshimohan:
Hi @tbhallett. I have just done a final update of the ResourceFiles after running the data_file_processing scripts. I've compared the latest files with the older version and a change is recorded because some item_codes undergo a <1.5% change in the probability of availability in scenarios 12-15, which I think is due to a minor update in the categorisation of consumables into programs. 
These ResourceFiles are final and should be good to merge into master.

resources/costing/ResourceFile_Consumables_Inflow_Outflow_Ratio.csv
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
tbhallett, line 1: should this file be on this PR, or the costing one?

sakshimohan: Good point, Tim. The script that generates this resource file (`consumable_availability_estimation.py`) is in under this PR (#1367) but it's used in the costing module and I had manually imported it there (which is probably not best practice). I can undo this commit in the costing PR #1637 and merge it in from here perhaps? The RF on this PR (PR #1367) is the final one and should override the one in costing.

tbhallett: Ok -- if this is the final version of the file we want then I would propose the simplest thing to do would be :
* we leave this one here and we'll merge it imminently.
* then in the costing branch, when we merge in master, we make sure this version (which becomes the master version) over-writes the out-dated version on the costing branch.

Does that sound ok?

tbhallett (approved):

tamuri (approved):

Issue #1363, opened 2024-05-20
Pass arguments to a Scenario when running `tlo scenario-run`


tamuri:
We need to allow scenarios to modify behaviour based on command-line arguments. A few use cases:

- Suspend scenario at date `tlo scenario-run /this/is/my/scenario.py --suspend-date 2020-01-01`
- Resuming scenario from the path `tlo scenario-run /this/is/my/scenario.py --resume-from-dir suspended-scenario-20240520`
- Change the mode of the healthsystem `tlo scenario-run /this/is/my/scenario.py --healthsystem-mode 2`

The first two examples modify the internal sim behaviour, whilst the last example is for the epi. 

I have to read how the [Click package handles unknown arguments](https://click.palletsprojects.com/en/8.1.x/advanced/#forwarding-unknown-options) - whether we can bundle them up and pass them to the scenario. It seems to discourage it. I don't think it's right to put all the options in `cli.py` itself; we'd have to update it everytime someone wanted to add an option to the scenario. I'm leaning toward injecting the unknown arguments into the Scenario class into, say, a field called `args` that users can inspect and use as they need.

Other suggestions welcome.

Issue #1335, opened 2024-05-10
Do scaling runs for mode_appt_constraints = 2 too


marghe-molaro:
In order to weed out errors before large jobs in this mode are submitted, ensure scaling runs also consider the mode_appt_constraints = 2 set-up.

matt-graham:
Question from meeting with @tamuri, @giordano, @matt-graham and @willGraham01 

- Do we want to do this on a single (larger) machine along with mode 1 runs or a separate machine?

Seems like we're leaning towards former.

tamuri:
@marghe-molaro / @tbhallett - we're working on adding `mode_appt_constraints=2` to the nightly long runs. Couple of questions:

1. If we're using the `long_run_all_diseases.py` scenario file to run the mode=2 runs, is the only necessary change overriding the `mode_appt_constraints` parameter? This parameter is set to 1 by `get_parameters_for_status_quo()` but we now have a way to change it to 2 on-demand.
2. Will the post-processing scripts (that create the plots) require any modification?

marghe-molaro:
Hi @tamuri, 

1. I think the "dream" nightly long runs scenarios for mode 2 would be as follows:
```
                "HealthSystem": {
                    "mode_appt_constraints": 1,                            # <-- Start in mode 1
                    "mode_appt_constraints_postSwitch": 2,      # <-- Include a transition to mode 2, to pick up any issues with this
                    "year_mode_switch": 2012,                              # <-- Could make this quite soon, but I'd say >1 year
                    "tclose_overwrite": 1,                                        # <-- In most of our runs in mode 2, we chose to overwrite tclose
                    "tclose_days_offset_overwrite": 7,                  # <-- and usually set it to 7.
                }
```
Especially the latter two, however, are not crucial. If it were much simpler to run in a simple ```mode_appt_constraints =2 ``` I think it would be fine.

2. I think maybe a time evolution of HSIs delivered, either as total number or fraction of requested that are delivered, and a time evolution of DALYs. Both could be done for the case of HIV/AIDS only, which are the main drivers of HSIs and DALYs initially. 

@tbhallett, what do you think?

tbhallett:
1) that seems fine to me. (If that's the target for most of the runs we use, which I think it is)

2) I think the original processing scripts can be used but it sounds like - in addition - Margherita would like to see the time trends in certain things which the original scripts don't look at. 

I think the run is only from 2010-2019, and the target period for calibration is 2010-2019. 

[...]

PR #1227, opened 2023-12-11
Add support for saving and loading simulation state to / from files

matt-graham:
Potentially resolves #86 though currently this doesn't deal with exposing this functionality with scenarios.

Adds new methods `save_to_pickle` and `load_from_pickle` that respectively save the current simulation state to a pickle file  and load the simulation state from a pickle file (with the latter being a class method to allow using to directly load a simulation as `Simulation.load_from_pickle`. Pickling is dealt with by `dill` as this supports a much wider range of Python objects than built in `pickle` implementation. `dill` is added to package dependencies here, but the import is currently wrapped with logic to avoid `ImportError`s in environments which do not have `dill` available (with `save_to_pickle` and `load_from_pickle` raising informative exceptions in this case).

The contents of the current `Simulation.simulate` method have also been factored out in to three separate methods `Simulation.initialise`, `Simulation.run_simulation_to` and `Simulation.finalise` that deal allow initialising, running and finalising the simulation separately, with `Simulation.simulate` retaining the same behaviour by just calling the three in sequence. This allows simulations to be partially run to an intermediate date before the simulation end date, saved to file and then reloaded and continued. 

As there is also global state recorded in `tlo.logging` we also need to reconfigure logging on loading a simulation from file. I initially included some logic in `load_from_pickle` to inject loaded simulation in to `tlo` logger and set output file to previous logging path (logging `FileHandler` loaded by `dill` is not able to acquire a lock causing deadlocks if used), but decided it would be better to make this step explicit, particularly as I'd guess we would often want to write to a new log file when resuming a loaded simulation.

A set of tests that check for consistency of simulations when saving and loading from file, including using to resume a partially run simulation, are also added.

Apologies for the unrelated formatting changes `src/tlo/simulation.py`, I was using `black` to autoformat my changes and forgot it would also reformat rest of module - I can revert these bits if it makes a pain to review.

## Questions

* Instead of `save_to_pickle` and `load_from_pickle` and should we just use the names `save` and `load`?
* Should we explicitly set `simulation.output_file` to `None` in `Simulation.load_from_pickle` to guard against accidental use of previous log file (and potential deadlock issues)?

PR #1373, opened 2024-05-26
Allow users to pass command-line arguments to scenarios

tamuri:
For #1363 and #1335

- allow users to add their own arguments to scenarios, which will override attributes in the subclass
- add `--resume-simulation` and `--suspend-date` arguments, in anticipation of suspend/resume functionality, which is handled by BaseScenario.

I could not figure out a nice way to use the [Click](https://click.palletsprojects.com/en/8.1.x/) argument handling for partial handling of arguments inside of subclasses, so have implemented using argparse. Click will pass on any unhandled arguments to the scenario for handling upstream.

*Example*

`tlo scenario-run --draw-only src/scripts/dev/scenarios/playing_22.py --pop-size 200`

Here, only `--draw-only` is known and handled by the click decorators in `cli.py`. `--pop-size` is passed to a new scenario `parse_arguments()` method. In order for it to work, the scenario writer should have added the following to their scenario class:

```
    def add_arguments(self, parser):
        parser.add_argument('--pop-size', type=int)
```

This allows the scenario to accept the `--pop-size` argument which, in turn, sets `self.pop_size` (overriding any existing value).

### DIFF
--- a/src/tlo/scenario.py
+++ b/src/tlo/scenario.py
@@ -66,7 +66,7 @@ import json
 import pickle
 from itertools import product
 from pathlib import Path, PurePosixPath
-from typing import Optional, List
+from typing import List, Optional
 
 import numpy as np
 
--- a/src/tlo/scenario.py
+++ b/src/tlo/scenario.py
@@ -60,6 +60,7 @@ In summary:
 """
 
 import abc
+import argparse
 import datetime
 import json
 import pickle
@@ -71,6 +72,7 @@ import numpy as np
 
 from tlo import Date, Simulation, logging
 from tlo.analysis.utils import parse_log_file
+from tlo.util import str_to_pandas_date
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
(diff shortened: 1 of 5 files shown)


tamuri:
Note: I haven't updated the `batch-submit` subcommand to handle these scenario arguments in this PR - that's for a future PR once this local-running has been tested (and suspend/resume is in??)

src/tlo/scenario.py
```diff
@@ -125,6 +127,43 @@ def __init__(
...
+        self.add_arguments(parser)
+
+        arguments = parser.parse_args(self.arguments)
+
+        # set the arguments as attributes of the scenario
+        for key, value in vars(arguments).items():
+            if value is not None:
+                if hasattr(self, key):
+                    logger.info(key="message", data=f"Overriding attribute: {key}: {getattr(self, key)} -> {value}")
+                setattr(self, key, value)
```
matt-graham, line 152: This assumes we never want an argument to have value `None` - this is probably fine but we might want to document this in the `add_arguments` docstring?

matt-graham (changes requested):
Thanks this looks good in general. Have suggested some refactoring of the new `test_scenario.py` module to avoid the dependencies between the tests and a couple of other minor comments.

tests/test_scenario.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.fixture
+def scenario_path():
+    return Path(f'{os.path.dirname(__file__)}/resources/scenario.py')
+
+
+@pytest.fixture
+def suspend_date():
+    return Date(2012, 3, 4)
+
+
```
matt-graham, line 20:
```suggestion
```
As no longer used

tests/test_scenario.py
```diff
@@ -0,0 +1,75 @@
...
+    return 100
+
+
+@pytest.fixture
+def loaded_scenario(scenario_path):
+    return ScenarioLoader(scenario_path).get_scenario()
+
+
+@pytest.fixture
+def arguments(pop_size, suspend_date):
```
matt-graham, line 32:
```suggestion
def arguments(pop_size):
```

tests/test_scenario.py
```diff
@@ -0,0 +1,75 @@
...
+    assert isinstance(loaded_scenario, BaseScenario)
+    assert loaded_scenario.scenario_path == scenario_path
+    assert hasattr(loaded_scenario, "pop_size")  # Default value set in initialiser
+
+
+def test_parse_arguments(loaded_scenario_with_parsed_arguments, pop_size):
+    """Check we can parse arguments related to the scenario. pop-size is used by our scenario,
+    suspend-date is used in base class"""
+    assert loaded_scenario_with_parsed_arguments.pop_size == pop_size
+    assert not hasattr(loaded_scenario_with_parsed_arguments, 'resume_simulation')
```
matt-graham, line 53:
```suggestion
```

tests/test_scenario.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_config(tmp_path, loaded_scenario_with_parsed_arguments, arguments):
+    """Create the run configuration and check we've got the right values in there."""
+    config = loaded_scenario_with_parsed_arguments.save_draws(return_config=True)
+    assert config['scenario_seed'] == loaded_scenario_with_parsed_arguments.seed
+    assert config['arguments'] == arguments
+    assert len(config['draws']) == loaded_scenario_with_parsed_arguments.number_of_draws
+
+
+def test_runner(tmp_path, loaded_scenario_with_parsed_arguments, pop_size, suspend_date):
```
matt-graham, line 64:
```suggestion
def test_runner(tmp_path, loaded_scenario_with_parsed_arguments, pop_size):
```

matt-graham (approved):
Thanks for making the changes! Looks good to me now - I've suggested some minor clean-ups to the tests to remove now used `suspend_date` fixture and check for absence of `resume_simulation` attribute, but as I think the tests will still be valid with these present if you think better to leave in place for adding in those arguments once we get #1227 in then happy for this to go in as is.

PR #1378, opened 2024-05-28
equipment pkgs

EvaJanouskova:


tbhallett (commented):
As these call (to `lookup_item_codes_from_pkg_name()`) are relatively "expensive", I think we want to avoid repeating them. However, as written currently, every time the `HSI_Event` is run, we have to do the look-up. 

So my proposal would be to do this:

1) In `Equipment.__init__()`, add something like:
```python
self._pkg_lookup = self._create_pkg_lookup()
```

2) In `self._create_pkg_lookup()`, use the ResourceFile to return a `dict` of the form: `{PackageName: Set(item_codes_in_package)}`.

3) Re-write [`lookup_item_codes_from_pkg_name()`](https://github.com/UCL/TLOmodel/blob/6494277376ad3c66d27ddb547a6a61ac3f3012d1/src/tlo/methods/equipment.py#L251-L260) so that it is looking inside that dict (`self._pkg_lookup`) rather than searching through the ResourceFile. e.g.
```python
def lookup_item_codes_from_pkg_name(self, pkg_name: str) -> Set[int]:
   if pkg_name in self._pkg_lookup:
        return self._pkg_lookup.get(pkg_name)
   else:
        raise ValueError(f"Package Name not recognised: {pkg_name}")
```

4) Lastly, we could rename `lookup_item_codes_from_pkg_name` to `from_pkg_name` to make it slicker.

What do you think?

EvaJanouskova:
@tbhallett, I mostly agree. Only to call the pkgs, it is still too long:
`self.add_equipment(self.healthcare_system.equipment.from_pkg_name('Major Surgery'))`

I think, we could make it more user friendly, if we have either
- two fncs: `add_equipment()` and `add_equipment_from_pkg()`, or
- an input parametr for `add_equipment()` fnc, allowing to use pkg_names instead of item_names, smt like 
`def add_equipment(self, item_codes: Union[int, str, Iterable[int], Iterable[str]], from_pkg=False) -> None:`

_Also, shouldn't we call the input `item_codes` differently since it can be code(s) or name(s) of the item(s)? Maybe `equip_items`?_

tbhallett:
> @tbhallett, I mostly agree. Only to call the pkgs, it is still too long:
> 
> `self.add_equipment(self.healthcare_system.equipment.from_pkg_name('Major Surgery'))`
> 
> 
> 
> I think, we could make it more user friendly, if we have either
> 
> - two fncs: `add_equipment()` and `add_equipment_from_pkg()`, or
> 
> - an input parametr for `add_equipment()` fnc, allowing to use pkg_names instead of item_names, smt like 
> 
> `def add_equipment(self, item_codes: Union[int, str, Iterable[int], Iterable[str]], from_pkg=False) -> None:`
> 

I know what you mean, but I think it's ok. That statement may be a bit long to write but is the most transparent and avoids us making big changes to the equipment module at this point. 

(Originally the two steps were intended to be separate (like consumables) where the module looks up the item codes and save them, and the HSI EVENT just refers to the save information on the module. But now we have the cached look-up this is not necessary.)

I think the ultimate thing would be just to pass in a package name string as any other string and for it to be automatically detected as being a package. 

Please could you raise this as an issue? (Could click "create issue from comment" on this box!)


> 
> _Also, shouldn't we call the input `item_codes` differently since it can be code(s) or name(s) of the item(s)? Maybe just `items` or `equip_items`?_

Yes, good point. We could call the argument where we accept int
or str just 'items'. This can be a separate PR to tidy that up if you'd like to raise it. Thanks.

EvaJanouskova:
_I did rebase this on hallett/equipment_changes_and_structure and updated the  `test_core_functionality_of_equipment_class` in `test_equipment` along with that._

It seems my commits broke the tests. I'm looking into it but so far no success.

EvaJanouskova:
> _I did rebase this on hallett/equipment_changes_and_structure and updated the `test_core_functionality_of_equipment_class` in `test_equipment` along with that._
> 
> It seems my commits broke the tests. I'm looking into it but so far no success.

I found at least one thing: I added two items to the `RF_EquipmentCatalogue`, but it also needs to be added in the `data_availability` resource file. Feel free to proceed with that if you find time before I do. (Will continue tomorrow in the late morning/early afternoon.)

joehcollins:
@EvaJanouskova - i've added ANC packages but the tests are failing due to an assert error at line 126 of equipment.py. Are test passing on your machine?

EvaJanouskova:
> @EvaJanouskova - i've added ANC packages but the tests are failing due to an assert error at line 126 of equipment.py. Are test passing on your machine?

Thank you Joe.

No, the tests are failing because of what I state in the comment above.

EvaJanouskova:
_rebased on `hallett/equipment_changes_and_structure` branch_

tbhallett:
> > > _I did rebase this on hallett/equipment_changes_and_structure and updated the `test_core_functionality_of_equipment_class` in `test_equipment` along with that._
> > > It seems my commits broke the tests. I'm looking into it but so far no success.
> > 
> > 
> > I found one thing: I added two items to the `RF_EquipmentCatalogue`, but they also need to be added in the `data_availability` resource file. Feel free to proceed with that if you find time before I do. _(I'll continue tomorrow in the late morning/early afternoon.)_
> 
> @sakshi, @tbhallett, I added the new items (last 2 lines: Endoscope, item code 402; Electrocardiogram, item code 403) of equipment in the `ResourceFile_EquipmentCatalogue.csv` which were added in the original PR of equipment implementation but were lost in the process of creating the new PR.
> 
> Could you please add these items along with estimated availability into the script generating the `ResourceFile_Equipment_Availability_Estimates.csv` and regenerate this RF?

The file is this one: `src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py`. It should run for you. It will automatically incorproate item_codes you have added to the catalogue.

EvaJanouskova:
> > > > _I did rebase this on hallett/equipment_changes_and_structure and updated the `test_core_functionality_of_equipment_class` in `test_equipment` along with that._
> > > > It seems my commits broke the tests. I'm looking into it but so far no success.
> > > 
> > > 
> > > I found one thing: I added two items to the `RF_EquipmentCatalogue`, but they also need to be added in the `data_availability` resource file. Feel free to proceed with that if you find time before I do. _(I'll continue tomorrow in the late morning/early afternoon.)_
> > 
> > 
> > @sakshi, @tbhallett, I added the new items (last 2 lines: Endoscope, item code 402; Electrocardiogram, item code 403) of equipment in the `ResourceFile_EquipmentCatalogue.csv` which were added in the original PR of equipment implementation but were lost in the process of creating the new PR.
> > Could you please add these items along with estimated availability into the script generating the `ResourceFile_Equipment_Availability_Estimates.csv` and regenerate this RF?
> 
> The file is this one: `src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py`. It should run for you. It will automatically incorproate item_codes you have added to the catalogue.

Doesn't it require to define availability for those items somewhere? If so, where and to what values?

It gives me this error message:
```
Traceback (most recent call last):
  File "/home/eva/PycharmProjects/TLOmodel/src/scripts/data_file_processing/healthsystem/equipment/equipment_availability_estimation.py", line 337, in <module>
    assert final_equipment_availability_export_full.notnull().all()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Issue #1325, opened 2024-04-25
Module Dependency Errors


mnjowe:
****** For such moments where you don't want to run a full model analyses **************
@tbhallett , @tamuri  and @matt-graham 
With modules declaring dependencies and analyses wanting the modules where these dependencies are declared registered in `sim.register()`, I think it is somewhat difficult for someone new to TLM to setup a small analyses and run it as it will keep crashing with module dependency errors. see below for an example

`tlo.dependencies.ModuleDependencyError: Module Tb depends on Epi which is missing from modules to register as are any alternatives to Epi.`

Wouldn't it be nice if we can automatically register modules that are missing in simulation? i.e. when disease modules are declaring module dependencies they also declare/include the module name where this dependency is found. something like  `INIT_DEPENDENCIES = {'enhanced_lifestyle': 'Lifestyle'}`. That way we could use the module name to automatically register the module if not already registered in `sim.register()`?

I know there might be some better ways to doing it BUT I just feel it would simplify the process of setting up a smaller analyses if all missing modules could be automatically registered instead of crashing the simulation.

tbhallett:
I think this is a good idea @mnjowe

matt-graham:
Agreed - this would be a nice feature to add. I think how the module dependencies are already declared could already allow doing this on the presumption that there is only one subclass of `Module` with a particular name (for example that there aren't two `Lifestyle` classes defined in different Python module files) which at the moment is the case, though having a more explicit specification of where the class is defined would definitely be more robust and easier to work with. Specificlly the `get_module_class_map` helper function in `src/tlo/dependencies.py` 
https://github.com/UCL/TLOmodel/blob/581ebe45b752b2e97ad0ef3add5bd5e8b2bdc86b/src/tlo/dependencies.py#L175-L199
allows constructing a map from module name strings to module classes.

[...]

PR #1379, opened 2024-05-29
Auto register module dependencies

mnjowe:
This PR aims at introducing a mechanism to automatically register module dependencies if they are not explicitly registered in simulation. Once merged it will close issue https://github.com/UCL/TLOmodel/issues/1325.


src/tlo/dependencies.py
```diff
@@ -120,29 +126,35 @@ def depth_first_search(module):
...
-                            'missing from modules to register'
-                        )
-                        if len(alternatives_with_instances) == 0:
-                            message += f' as are any alternatives to {dependency}.'
-                        else:
-                            message += (
-                                ' and there are multiple alternatives '
-                                f'({alternatives_with_instances}) so which '
-                                'to use to resolve dependency is ambiguous.'
+                    if auto_register_dependencies:
```
matt-graham, line 132: Just realised that this might not quite works as expected when an 'alternative' module dependency (like `SimplifiedBirths`)  is included in the registered modules. If for example `SimplifiedBirths` is registered and the other full pregnancy related modules such as `Labour` and `Contraception` are not, then if `auto_register_dependencies` is set to `True` then the full pregnancy related modules will be automatically registered if they are a dependency of any of the other registered modules even though in this situation it is likely the user is expecting to use `SimplifiedBirths` instead. I think to resolve this we just need to move the logic here to something like the following

```Python
if dependency not in module_instance_map:
    alternatives_with_instances = [
        name for name, instance in module_instance_map.items()
        if dependency in instance.ALTERNATIVE_TO
    ]
    if len(alternatives_with_instances) == 0 and auto_register_dependencies:
        module_instance_map[dependency] = module_class_map[dependency](
            resourcefilepath=resourcefilepath
        )
        yield from depth_first_search(dependency)
    elif len(alternatives_with_instances) != 1:
        message = ... # This block should be same as previously
    else:
        yield from depth_first_search(alternatives_with_instances[0])
else:
    yield from depth_first_search(dependency)
```

This will only automatically register the dependency if it is not already registered _and_ there are no alternatives registered.

matt-graham (changes requested):
Thanks @mnjowe, this is looking great. After looking over this again I think there is an edge case we are not dealing with quite correctly when there are 'alternative' modules registered like `SimplifiedBirths` - I've commented below with more details and a suggested fix. I've also made a few small stylistic suggestions.

src/tlo/dependencies.py
mnjowe: @matt-graham . Just noting also that get dependencies function [here](https://github.com/UCL/TLOmodel/blob/1bcb1cbebd93cadabbd802b66f4530ef94f7db2d/src/tlo/dependencies.py#L126) only consider dependencies declared in `INIT_DEPENDENCIES` and `OPTIONAL_INIT_DEPENDENCIES`. it doesn't including those declared in `ADDITIONAL_DEPENDENCIES` which makes auto registering of dependencies [fail here ](https://github.com/UCL/TLOmodel/blob/1bcb1cbebd93cadabbd802b66f4530ef94f7db2d/src/tlo/dependencies.py#L292C5-L297C10) when you include modules like `HIV` which has additional dependencies. Any ideas? or should we re-do the registering of dependencies [here instead of throwing an error?](https://github.com/UCL/TLOmodel/blob/1bcb1cbebd93cadabbd802b66f4530ef94f7db2d/src/tlo/dependencies.py#L292C5-L297C10)

matt-graham: Ah that's a good point, I'd forgotten about the distinction between `INIT_DEPENDENCIES` and `ADDITIONAL_DEPENDENCIES`.

I think then it might make sense to split out the functionality for auto-registering missing dependencies in to a separate function from `topologically_sort_modules` that is called before the `if sort_modules` and `if check_all_dependencies` blocks, that is something like

```Python
        if auto_register_dependencies:
            modules = [
                *modules, 
                *initialise_missing_dependencies(modules, resourcefilepath=self.resourcefilepath)
            ]
        if sort_modules:
            modules = list(topologically_sort_modules(modules))
        if check_all_dependencies:
            check_dependencies_present(modules)
```

with the logic for getting set of missing dependencies factored out from `check_dependencies_present` into a new function `get_missing_dependencies` which can then also be used to implement the `initialise_missing_dependencies` as something like (haven't tested this works)

```Python
def get_missing_dependencies(
    module_instances: Iterable[Module],
    get_dependencies: DependencyGetter = get_all_dependencies,
) -> Set[str]:
    """Get the set of missing required dependencies if any from an iterable of modules.

    :param module_instances: Iterable of ``Module`` subclass instances to get missing 
        dependencies for.
    :param get_dependencies: Callable which extracts the set of dependencies to check
        for from a module instance. Defaults to extracting all dependencies.
    :return: Set of ``Module`` subclass names corresponding to missing dependencies.
    """
    module_instances = list(module_instances)
    modules_present = {type(module).__name__ for module in module_instances}
    modules_present_are_alternatives_to = set.union(
        # Force conversion to set to avoid errors when using set.union with frozenset
        *(set(module.ALTERNATIVE_TO) for module in module_instances)
    )
    modules_required = set.union(
        *(set(get_dependencies(module, modules_present)) for module in module_instances)
    )
    missing_dependencies = modules_required - modules_present
    return (
        missing_dependencies - modules_present_are_alternatives_to
    )


def initialise_missing_dependencies(modules, **module_kwargs):
    """Get list of initialised instances of any missing dependencies for an iterable of modules.

    :param modules: Iterable of ``Module`` subclass instances to get instances of missing 
        dependencies for.
    :param module_kwargs: Any keyword arguments to use when initialising missing
        module dependencies.
    :return: List of ``Module`` subclass instances corresponding to missing dependencies.
    """
    module_class_map = get_module_class_map(set())
    missing_dependencies = get_missing_dependencies(
        module_instances, get_all_dependencies
    )
    return [
        module_class_map[dependency](**module_kwargs)
        for dependency in missing_dependencies:
     ]

def check_dependencies_present(
    module_instances: Iterable[Module],
    get_dependencies: DependencyGetter = get_all_dependencies,
):
    """Check whether an iterable of modules contains the required dependencies.

    :param module_instances: Iterable of ``Module`` subclass instances to check.
    :param get_dependencies: Callable which extracts the set of dependencies to check
        for from a module instance. Defaults to extracting all dependencies.

    :raises ModuleDependencyError: Raised if any dependencies are missing.
    """
    missing_dependencies = get_missing_dependencies(
        module_instances, get_dependencies
    )
    if len(missing_dependencies) > 0:
        raise ModuleDependencyError(
            'One or more required dependency is missing from the module list and no '
            'alternative to this / these modules are available either: '
            f'{missing_dependencies}'
        )
```

We want to add dependencies before topologically sorting so that these are taking in to account when sorting, and for the topological sorting we need to only iterate over the `INIT_DEPENDENCIES` as by construction these don't allow any directed cycles to avoid a infinite recursion, while if we tried to also account for registering any `ADDITIONAL_DEPENDENCIES` in this recursion I think we might run the risk of get stucking in infinite recursions.

mnjowe: This is great @matt-graham . Now that we have `get_missing_dependencies` and `initialise_missing_dependencies` functions, do we still need the` check_dependencies_present` function? I was thinking we can do that when testing for these functions in `test_module_dependencies`. I'm I right?

mnjowe: With the above implementation, we will also have to think of a way of adding dependencies of missing dependencies i.e when `Hiv` is registered the implementation will capture `Tb` as a missing dependency of `Hiv` BUT will not capture `Epi` as a missing dependency of `Tb`

matt-graham: Hi @mnjowe - sorry for the delay responding here. You're absolutely right in your last comment that the current implementation will only capture direct missing dependencies not the indirect ([transitive](https://en.wikipedia.org/wiki/Transitive_dependency)) missing dependencies. To get these, one approach would be to use some form of recursion, for example

```Python
def get_missing_dependencies(
    module_instances: Iterable[Module],
    get_dependencies: DependencyGetter = get_all_dependencies,
) -> Set[str]:
    """Get the set of missing required dependencies if any from an iterable of modules.

    :param module_instances: Iterable of ``Module`` subclass instances to get missing
        dependencies for.
    :param get_dependencies: Callable which extracts the set of dependencies to check
        for from a module instance. Defaults to extracting all dependencies.
    :return: Set of ``Module`` subclass names corresponding to missing dependencies.
    """
    module_instances = list(module_instances)
    modules_present = {type(module).__name__ for module in module_instances}
    modules_present_are_alternatives_to = set.union(
        # Force conversion to set to avoid errors when using set.union with frozenset
        *(set(module.ALTERNATIVE_TO) for module in module_instances)
    )
    module_class_map = get_module_class_map(set())

    def add_required_dependencies(module_name, modules_required):
        modules_required.add(module_name)
        module_class = module_class_map[module_name]
        dependencies = get_dependencies(module_class, modules_present)
        for dependency in dependencies:
            if dependency not in modules_required:
                add_required_dependencies(dependency, modules_required)

    modules_required = set()
    for module in modules_present:
        add_required_dependencies(module, modules_required)

    missing_dependencies = modules_required - modules_present
    return (
        missing_dependencies - modules_present_are_alternatives_to
    )
```

mnjowe: This is great @matt-graham , thanks. Would you also look at what I have done already in my last commit to see if we can go by that?

src/tlo/dependencies.py
```diff
@@ -57,6 +57,69 @@ def get_all_dependencies(
...
+            dependencies = [
+                module_class_map[dependency](**module_kwargs)
+                for dependency in missing_dependencies
+            ]
+            all_module_instances.extend(dependencies)
+            get_all_missing_dependencies(all_module_instances)
+
+        return all_module_instances
+
+    missing_module_instances = get_all_missing_dependencies(module_instances)
```
matt-graham, line 119: `get_all_missing_dependencies` returns `all_module_instances` which is defined in scope of the outer function, so `missing_module_instances` here will be I think be pointing to the same list object as `all_module_instances` already points to. Having multiple names attached to the same object again makes it a bit difficult to track what is going on in the function.

src/tlo/dependencies.py
```diff
@@ -57,6 +57,69 @@ def get_all_dependencies(
...
+
+    :param modules: Iterable of ``Module`` subclass instances to get instances of missing
+        dependencies for.
+    :param module_kwargs: Any keyword arguments to use when initialising missing
+        module dependencies.
+    :return: List of ``Module`` subclass instances corresponding to missing dependencies.
+    """
+    module_instances = modules
+    module_class_map = get_module_class_map(set())
+    all_missing_dependencies = list()
```
matt-graham, line 99: I think `all_missing_module_names` might be more descriptive here as it's not clear from the current name this is not just the set difference between `all_module_instances` and `module_instances`

src/tlo/dependencies.py
```diff
@@ -57,6 +57,69 @@ def get_all_dependencies(
...
+    all_module_instances = list(module_instances)
+
+    def get_all_missing_dependencies(module_instance: Iterable[Module]):
+        """ get all missing dependencies and return their module instances """
+        missing_dependencies = get_missing_dependencies(
+            module_instance, get_all_dependencies
+        )
+
+        if not missing_dependencies.issubset(all_missing_dependencies):
+            all_missing_dependencies.extend(missing_dependencies - set(all_missing_dependencies))
```
matt-graham, line 109: If `all_missing_dependencies` was itself a set then you could simplify this to

```Python
all_missing_dependencies |= missing_dependencies
```

or equivalently

```Python
all_missing_dependencies = all_missing_dependencies | missing_dependencies
```

src/tlo/dependencies.py
```diff
@@ -57,6 +57,69 @@ def get_all_dependencies(
...
+            dependencies = [
+                module_class_map[dependency](**module_kwargs)
+                for dependency in missing_dependencies
+            ]
```
matt-graham, line 113: I think `missing_module_instances` would be more descriptive here for the variable name than `dependencies`.

src/tlo/dependencies.py
```diff
@@ -57,6 +57,69 @@ def get_all_dependencies(
...
+            all_module_instances.extend(dependencies)
+            get_all_missing_dependencies(all_module_instances)
+
+        return all_module_instances
```
matt-graham, line 117: As `all_module_instances` is defined outside this function, returning here is not needed an makes it difficult to track how variables are being assigned to. I would say it would be better to either not return anything and rely on `all_module_instances` already being available in the outer scope, or better still, removing the dependence of the `get_all_missing_dependencies` function on the `all_module_instances` variable in the outer scope and instead having it just depend on its arguments. As the `module_instances` variable passed as an argument to the call of the function below already has the same contents as `all_module_instances` (but is potentially not a list but another iterable type), this should be relatively simple to do.

src/tlo/dependencies.py
```diff
@@ -57,6 +57,67 @@ def get_all_dependencies(
...
+    :param modules: Iterable of ``Module`` subclass instances to get instances of missing
+        dependencies for.
+    :param module_kwargs: Any keyword arguments to use when initialising missing
+        module dependencies.
+    :return: Set of ``Module`` subclass instances corresponding to missing dependencies.
+    """
+    module_class_map: Mapping[str, Type[Module]] = get_module_class_map(set())
+    all_module_instances: list[Module] = list(modules)
+
+    def add_missing_module_instances(modules: Iterable[Module], all_missing_module_names: set[str]) -> None:
```
matt-graham, line 100:
```suggestion
    def add_missing_module_instances(modules: list[Module], all_missing_module_names: set[str]) -> None:
```

If the suggestion below is applied technically `modules` will always be a list (and is required to be) so this would be a better type hint.

matt-graham (approved):
Thanks @mnjowe and sorry for the delay in getting back to you on this. This looks great - I've added some minor suggestions which I think slightly improve the readability of the function but these are subjective so feel free to ignore them and I'll approve now so you can merge when ready.

mnjowe:
Hi @matt-graham for your detailed review on this PR. I have committed your last suggestions as I think they indeed improve code  readability.  When all tests have passes feel free to merge this PR on my behalf(I currently don't have merge rights). I'm happy having it merged in master

PR #1386, opened 2024-05-30
Fix running_total_footprint in mode 2

marghe-molaro:
The running_total_footprint is not properly updated in mode 2. This variable is not relevant for the functioning of this mode (it is used in mode 1 to compute the daily squeeze factor), so this wouldn't have created any problems in previous runs*. However, since it is still logged, this commit ensures that the value logged is meaningful even in mode 2. 

*Note: it is not a problem even in the rescaling of mode 2 to effective capabilities because the former is based on the running_total_footprint computed in mode 1, prior to switching to mode 2.


tbhallett (approved):

tbhallett:
Thanks for spotting this.

Issue #1348, opened 2024-05-15
Batch of runs submitted from PR #1308 before/after review&master merge inconsistent 


marghe-molaro:
PR #1308 allows user to rescale HR capabilities to capture effective capabilities when transitioning to mode 2.

Two batch of runs where submitted to Azure from this PR, effect_of_policy-2024-04-09T132919Z (2 runs per draw) and effect_of_capabilities_scaling-2024-04-29T074640Z (10 runs per draw). While the rescaling seems to be occurring in both sets of runs, the evolution of DALYs recovered is much different when all other factors (cons. availability, HR used, etc...) are identical, as shown in plots below. 
![Scaling of capabilities](https://github.com/UCL/TLOmodel/assets/48129834/63c3fd1d-c7fe-477a-a1bd-bb359833290c)

The two branches from which these jobs were submitted produce identical outputs to the test test_rescaling_capabilities_based_on_squeeze_factors, suggesting difference does not arise from stylistic changes in how rescaling is computed in more recent branch suggested by @matt-graham. While the more recent branch includes checks on refactoring not being infinite, which would have occurred in the case of some mental health care workers, this should not be affecting DALYs related to HIV/AIDS shown here which do not rely on this medical worker class. 

Other differences between branches include changes brought over by master including:
- Refactoring of generic first appointments
- Transition to python 3.11
- new scenario switcher class

@matt-graham any suggestions on things to double check would be very appreciated!

matt-graham:
Hi @marghe-molaro. Nothing immediately comes to mind with regards to things that could have caused such a substantial change in model behaviour. As far as we're aware the changes to HSI events and generic first appointments in #1289, #1292 and #1315 shouldn't have changed the model output behaviour at all, and similarly changing to running on Python 3.11 on Azure in #1323 also should not have effected the model outputs. But obviously something has so maybe I'm assuming wrongly 😬. 

Do you have the specific Git commit SHAs for the two sets of runs (I think should be recorded under the key `commit` in the JSON file in the top-level directory containing the batch job results)? With those we could try running (shorter) runs locally and try to see  where they start diverging

marghe-molaro:
I totally agree @matt-graham that none of these changes should have made a difference, that's why the difference in outputs is so frustrating!

The two Git commit SHAs are are follows:
for effect_of_policy-2024-04-09T132919Z
  "commit": "c61895880abf8e942635839d7d62ed38c83f6f89",
[...]

PR #1387, opened 2024-05-30
Ensure individual properties updated across modules in generic first appointments

### LINKED ISSUE #1350: `do_at_generic_first_appt` and `do_at_generic_first_appt_emergency` should not be in core.Module (probably)
Re. the refactoring that happened in bd0a763. Some `methods`-submodule implementation details have leaked into the top-level  part of the package. I'd argue this is not desirable, and we have tried to maintain a separation between what we considered the framework proper and the epi/modelling details which sit in the `methods` subpackage.

Some fixes may as simple as renaming e.g. in `population` module, a simple renaming of s/patient/person/ may be enough to make it general-purpose again. I guess it's called Patient because it was introduced in the context of HealthSystem improvements.

The more undersirable changes, IMO, are the changes in `core` where we've pulled in the HSI appointment methods into `core.Module`. Can we discuss some alternative ways to adopt Will's improvements but in a way that maintains SoC.

matt-graham:
Part of addressing #1348 and also fixes #1350

Specifically aims to fix issue identified by @marghe-molaro in https://github.com/UCL/TLOmodel/issues/1348#issuecomment-2134959827 and discussed further in https://github.com/UCL/TLOmodel/issues/1348#issuecomment-2139374633 by updating data structure used to provide cached (memoized) view of a population data frame row to allow both reading and writing.

Also moves the code related to generic first appointments out of `tlo.core` into the `tlo.method.hsi_generic_first_appts` module, creating a new subclass `GenericFirstApptsModule` that modules which wish to subscribe to this behaviour need to subclass from.

---

After the changes in this PR, the `Population` object has a new method `individual_properties` which returns an instance of `IndividualProperties` (previously `PatientDetails`) for a particular `person_id` row index and with option to choose whether the view on to a population dataframe row provided by the `IndividualProperties` instance is read-only or not.

The `IndividualProperties` instance allows reading the properties of the individual with square-bracket indexing notation for example

```Python
individual_properties = sim.population.individual_properties(person_id=person_id)
if individual_properties["is_alive"] and individual_properties["age_years"] > 5:
    # do something
```

Unlike the previous `PatientDetails` type, I've removed the ability to access properties using dot-based attribute access in favour of standardizing on indexing notation implemented using a `__getitem__` method as (i) I think its better to have one standard way of doing things for consistency, (ii) overriding the `__getattr__` special method can sometimes lead to infinite recursion bugs if not careful and (iii) using indexing to access properties allows adding further functionality in future to the `IndividualProperties` class exposed as attributes without worrying about these clashing with property names (for example we could add a `symptoms` property to allow easier checking / updating of symptoms as a set of strings).

As for the previous `PatientDetails` type, `IndividualProperties` uses a lazy memoized approach to reading properties from the dataframe. A reference to the `person_id` row index and population dataframe are stored internally when an instance is initialised along with a dictionary to as a cache. On the first attempt to access a property, the value will be read from the dataframe and stored in the cache dictionary, with subsequent accesses of the same property using the cache rather than reading again, with the assumption being that the population dataframe is now being changed between reads.

To allow writing the `IndividualProperties` class now also defines a `__setitem__` method. If an `IndividualProperties` instances has had the `read_only` argument to the initialiser set to `True`, any attempts to write properties using the `__setitem__` method will raise an exception. Otherwise the key of the property being updated is added to a internal set (empty on initialisation) recording any properties updated, and the new value written to the property cache dictionary, meaning any subsequent attempts to access this property will return the updated value rather than the original value in the dataframe. 

By default any updated property values will not be reflected back in the original population dataframe. To allow the updates to be synchronized back to the dataframe a new `synchronize_updates_to_dataframe` method has been added to `IndividualProperties` which writes all properties from the updated set with their new values.

To guard against this manual finalisation / synchronization step being missed, I've made `IndividualProperties` a context manager by implementing `__enter__` and `__exit__` method, with `synchronize_updates_to_dataframe` automatically called on exiting in this case. So for example the recommended way of creating and using writeable instance of `IndividualProperties` is something like

```Python
with sim.population.individual_properties(person_id=person_id, read_only=False) as individual_properties:
    individual_properties["is_alive"] = False
    individual_properties["age_years"] = 100
    ...
```

Optionally, we could also use the context manager to try to explicitly prevent the user being able to read / write to the population dataframe within the with block context by having `IndividualProperties` store a reference to the `Population` object rather than just its `props` dataframe attribute, and doing something like temporarily assigning a dummy value like `None` to the `props` attribute before restoring the reference to the dataframe when exiting. This would allow us to guard against incorrect usages where the population dataframe is being changed while an `IndividualProperties` instance is in use. I haven't currently implemented this, as we know there are currently a few cases where the initialisers for HSI events that scheduled in the `do_at_generic_first_appt` methods are reading properties of the target directly from the population dataframe. Ideally we want to refactor these cases to pass in the relevant target attributes directly and read from the `IndividualProperties` instance.

While at the moment the `IndividualProperties` class is only being used in the generic first appointments logic, it could potentially be used in a lot of other places in the model to mediate accesses to the dataframe and having some logic to guard against incorrect use in this case as suggested above would be more important then.

In addition to the changes to `IndividualProperties` this PR also further slightly refactors the `Population` class by removing the stored reference to the top-level `Simulation` object and changing the initializer to instead take as an argument a dictionary of the properties to use to create the population dataframe columns (keyed by the property names). Removing this coupling to the `Simulation` object made it much easier to write unit tests for the `Population` object (added in a new `test_population.py` module in this PR) as it can now be initialised without creating a simulation just by creating a dummy dictionary of `Property` instances. There were a few cases of population level events which were using the stored reference to the simulation object in the population to acccess the simulation date but these have been refactored to instead use the simulation reference stored in the module for the event.


src/tlo/methods/alri.py
```diff
@@ -55,7 +56,7 @@
 # ---------------------------------------------------------------------------------------------------------
 
 
-class Alri(Module):
+class Alri(GenericFirstApptModule):
```
tamuri, line 59: Do you prefer subclassing over a mixin?

src/tlo/population.py
```diff
@@ -1,39 +1,67 @@
...
-class PatientDetails:
-    """Read-only memoized view of population dataframe row."""
-    
-    def __init__(self, population_dataframe: pd.DataFrame, person_id: int):
+class IndividualProperties:
+    """Memoized view of population dataframe row that is optionally read-only."""
+
+    def __init__(
+        self,
+        population_dataframe: pd.DataFrame,
```
tamuri, line 20: Interesting question as to whether we should hold reference to population instance or props here. Insisting on using this class with a context manager should discourage holding the object for long periods of time. That'll be a problem because we switch out the population's props dataframe when we add rows for growing population.

src/tlo/methods/alri.py
matt-graham: Subclassing seemed more natural here to me - no need for multiple inheritance and I think we would only use the mix-in with `Module` as base class anyway so wasn't clear there would be any advantage to a mix-in over subclassing.

tamuri: I thought it'd be easier to pull in other interfaces or methods for \`Module\` (e.g. for disease vs non-disease modules) if needed.

tamuri: Thought about this some more, and I do think we should use mixins here rather than subclass. Flexibility of mixins offer more opportunity to pull out and refactor some common functionality across `Module` classes.

matt-graham: I've now refactored to make a mix-in

src/tlo/population.py
matt-graham: I've now refactored the code a bit to simplify allowing 'locking' the population dataframe when withing the context manager in future without having to store a reference to population in individual properties. The `Population.individual_properties` method itself now acts as a context manager (rather than the returned object) and so can easily in future do something like set a `self._locked` attribute which if true prevents direct access to the dataframe via the `props` attribute (making it a property) and also disallows `do_birth` being called. The object returned by the `Population.individual_properties` context manager will now also be 'finalized' on exiting the context, so that any subsequent attempts to read or update properties raise an error. The object also no longer stores even a 'private' reference to the population dataframe (with access instead mediated via closures defined in initialiser) which should further guard against direct accesses to population dataframe.

src/tlo/analysis/hsi_events.py
```diff
@@ -47,7 +47,7 @@ def get_hsi_event_classes_per_module(
         module = importlib.import_module(f'tlo.methods.{module_name}')
         tlo_module_classes = [
             obj for _, obj in inspect.getmembers(module)
-            if is_valid_tlo_module_subclass(obj, excluded_modules)
+            if is_valid_tlo_module_subclass(obj, {})
```
matt-graham, line 50: This change is I noticed the previous implementation was not quite correct as `excluded_modules` here is taken to be a set of Python module names (submodules of `tlo.methods` package) while `is_valid_tlo_module_subclass` takes a set of TLO `Module` subclass names for the `excluded_modules` argument.

tamuri (approved):
Looks good to me. Had a thought whether to wrap the `yield properties; properties.finalize()` in a `try: ... finally: ...` but I think it's better it raises exception if it happens to fail for some reason.

matt-graham:
@tamuri Do you we want further review here or is this good to merge?

tamuri:
Good to merge 👍.

PR #1366, opened 2024-05-22
Inconsistency fix between HEAD of master and older PR branches.

willGraham01:
For the investigation in https://github.com/UCL/TLOmodel/issues/1348

Aims to analyse why we are now seeing a divergence in the scenario results, between `master` and another branch. The contents of this PR are formed of the commits of #1292, applied one at a time, starting from the "trusted" commit `012efd6` and each time requiring the state of the repository to pass the following script:

```bash
#!/bin/bash

# Good commit: 012ef6d, checksum should be dc273b2e0ef27a3d61afb954ede48b8ba038a9a7
REFERENCE_CHECKSUM=dc273b2e0ef27a3d61afb954ede48b8ba038a9a7
YEARS=0
MONTHS=1
INITIAL_POPULATION=1000
RUN_CHECKSUM=$( \
    python src/scripts/profiling/scale_run.py \
    --years $YEARS --months $MONTHS --log-final-population-checksum \
    --initial-population $INITIAL_POPULATION | tail -n 1 \
    | grep -oP "(?<=checksum: ).*(?=\"])" \
)

if [ $? -ne 0 ]
then
    echo "Python script failed"
    exit 125
fi

echo "Run checksum was:  $RUN_CHECKSUM"
echo "Checksum to match: $REFERENCE_CHECKSUM"

if [[ "$RUN_CHECKSUM" == "$REFERENCE_CHECKSUM" ]]
then
    echo "That's a match"
else
    echo "No match, fail"
    exit 1
fi
```

Following the application of all the commits on #1292, `master` was merged in.

Overall `diff` is small (as should be expected) and consists of a few places where inconsistencies between RNG generation were spotted. Most glaringly, this PR [introduces an explicit ordering](https://github.com/UCL/TLOmodel/blob/452e624d578dfae3aaf9f32a5f3c6c93c94864ec/src/tlo/methods/hsi_generic_first_appts.py#L24-L73) for how HSI generic appointments run through each module to determine a follow up appointment. This ordering was implicit before the refactoring in #1292, and in #1348 it is believed that this ordering might have a long-term impact on the simulation results if allowed to snowball.

<details>
<summary>A commit-by-commit commentary of this process is available here.</summary>

Comments are made on each of the commits of #1292. New commit hashes (when conflicts needed to be resolved) are given for the corresponding state on this branch.

```bash
f776b57c4 Remove erroneously added "age_exact_years" in test_stunting # good
[...]

PR #1389, opened 2024-05-31
Fixes for diarrhoea generic first appointment logic and updating of individual properties across modules

matt-graham:
Related to #1348  and #1366.

As discussed in software engineering meeting today, this PR addresses two known issues with current refactored generic first appointments logic:

1. A change in the conditions on which the diarrhoea module does or does not perform an action on a generic first appointment introduced accidentally by the refactoring in #1292 is fixed to return to the previous behaviour (not performing any further action if the individual is over the age of 5 _or_ does not have any diarrhoea symptoms).
2. The issue identified by @marghe-molaro in https://github.com/UCL/TLOmodel/issues/1348#issuecomment-2134959827, that is that in 
  https://github.com/UCL/TLOmodel/blob/df25c764e6d6fc99f224a65d18a030c67630c5a4/src/tlo/methods/hsi_generic_first_appts.py#L80-L103
  updates to an individual properties by previous modules in the loop are not reflected necessarily in the `patient_details` structure passed to the `do_at_generic_first_appt` method of the current module in the loop as the updates are only made to the dataframe after the loop has completed. Due the subsequent changes to the `PatientDetails` type made in #1315, this will actually only be an issue if an update is made to a property by a module that has previously been read (and cached) by that or another module and then a subsequent module reads the same property again (as it will get the initial cached value). I'm not sure there are cases where that chain actually happens currently given only very few modules actually output updates to the individual properties and they are for properties usually specific to the module, so this possibly explains why @willGraham01 was able to get consistent dataframe checksums when working on #1366 even without this fix.

The fix for (2) here uses the simple approach I mentioned in https://github.com/UCL/TLOmodel/issues/1348#issuecomment-2139374633 which will loose some of the performance benefits we currently have by doing updates to the dataframe whenever a module returns properties to update (rather than doing all updates at once). In the longer term #1387 should fix the same issue in a more performant manner.

### DIFF
--- a/src/tlo/methods/diarrhoea.py
+++ b/src/tlo/methods/diarrhoea.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List
 
 import numpy as np
 import pandas as pd
@@ -944,13 +944,14 @@ class Diarrhoea(Module):
         self,
         patient_id: int,
         patient_details: PatientDetails,
+        symptoms: List[str],
         diagnosis_function: DiagnosisFunction,
         **kwargs,
     ) -> IndividualPropertyUpdates:
         # This routine is called when Diarrhoea is a symptom for a child
         # attending a Generic HSI Appointment. It checks for danger signs
         # and schedules HSI Events appropriately.
-        if patient_details.age_years > 5:
+        if patient_details.age_years > 5  or "diarrhoea" not in symptoms:
             return {}
 
         # 1) Assessment of danger signs
--- a/src/tlo/methods/hsi_generic_first_appts.py
+++ b/src/tlo/methods/hsi_generic_first_appts.py
@@ -72,10 +72,7 @@ class HSI_BaseGenericFirstAppt(HSI_Event, IndividualScopeEventMixin):
 
         # Dynamically create immutable container with the target's details stored.
         # This will avoid repeat DataFrame reads when we call the module-level functions.
-        df = self.sim.population.props
         patient_details = self.sim.population.row_in_readonly_form(self.target)
-
-        proposed_patient_details_updates = {}
 
         for module in modules.values():
             module_patient_updates = getattr(module, self.MODULE_METHOD_ON_APPLY)(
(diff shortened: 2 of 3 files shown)


matt-graham:
@marghe-molaro this PR should hopefully deal with the issues so once this is reviewed / merged to `master` you should be able to merge in to your branch for analysis runs and re-run to check if we restore original behaviour. If it does then we can close #1348.

tamuri (approved):

marghe-molaro:
Thank you very much @matt-graham!
Now that @tamuri has approved these changed I will merge these fixes into commit a91d60c518069ed11477f27fb3c81f1ee92302ca, which produced the plots in which we first spotted an issue. This will allow us to establish whether the issues fixed in this PR were behind the discrepancy.

marghe-molaro:
@matt-graham it appears there's an issue with the diarrhoea test

matt-graham:
It looks like Will [also updated the tests](https://github.com/UCL/TLOmodel/pull/1366/files#diff-ccf2792fb9d59e7d4b39bc97829dfcf18e72901eb0f2728c88350de927683015) that are failing here in #1366 to force individual to have diarrhoea symptoms - currently investigating why that was needed and will update test here if necessary.

matt-graham:
> It looks like Will [also updated the tests](https://github.com/UCL/TLOmodel/pull/1366/files#diff-ccf2792fb9d59e7d4b39bc97829dfcf18e72901eb0f2728c88350de927683015) that are failing here in #1366 to force individual to have diarrhoea symptoms - currently investigating why that was needed and will update test here if necessary.

I realized the manual setting of `diarrhoea` symptoms was now necessary as prior to the refactoring we called a method `Diarrhoea.do_when_presentation_with_diarrhoea` which assumed symptoms had already been checked, while post refactoring the logic of this method has been moved to the  `Diarrhoea.do_at_generic_first_appt` method which now with the fix in the PRs, exits straight away if there isn't `"diarrhoea"` present in the passed set of symptoms. Have now updated the test to manually create a set of symptoms with `"diarrhoea"` present where applicable so this should hopefully fix the failure.

matt-graham:
As this is a change to the test to keep it up to sync with the changes in the module, this shouldn't affect the validity of any runs with the changes here merged in before the last commit @marghe-molaro, just in case you've already set anything off!

marghe-molaro:
Hi @matt-graham, quick update: @tbhallett noticed that this PR was based on a version of master that had brought in changes to the HIV/TB/malaria modules which lead to a rescaling of DALYs and scheduled HSIs for those diseases (#1273). By comparing the original runs which included the error with the fixes brought in this PR (#1389), it has therefore become difficult to isolate the effect of modules reordering on the changes to the HIV DALYs evolution.

I will submit a new set of test runs from the current version of master where the only change added is the scrambling of the disease modules to recover the order in the original code vs the one introduced by PR #1292. Hopefully this will allow us to establish once and for all whether reordering the modules during the first generic appts leads to any statistically significant changes. Let me know if you have any thoughts on this.

PR #1396, opened 2024-06-04
Enforce order in which modules are dealt with in _do_on_generic_first_appt

marghe-molaro:
Quick PR to ensure we can specify in which order modules are dealt with in _do_on_generic_first_appt. This will have to be changed manually, and will only be used for testing (will not be merged in master).
@matt-graham could you just double check this looks all right to you?

### DIFF
--- a/src/tlo/methods/hsi_generic_first_appts.py
+++ b/src/tlo/methods/hsi_generic_first_appts.py
@@ -68,6 +68,9 @@ class HSI_BaseGenericFirstAppt(HSI_Event, IndividualScopeEventMixin):
         """
         # Make top-level reads of information, to avoid repeat accesses.
         modules: OrderedDict[str, "Module"] = self.sim.modules
+        print(modules)
+        print(modules.values())
+        exit(-1)
         symptoms = modules["SymptomManager"].has_what(self.target)
 
         # Dynamically create immutable container with the target's details stored.
--- a/src/tlo/methods/hsi_generic_first_appts.py
+++ b/src/tlo/methods/hsi_generic_first_appts.py
@@ -21,6 +21,43 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+fixed_modules_order = [
+    'ImprovedHealthSystemAndCareSeekingScenarioSwitcher',
+    'Demography',
+    'Lifestyle',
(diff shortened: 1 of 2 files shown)


src/tlo/methods/hsi_generic_first_appts.py
```diff
@@ -67,14 +69,16 @@ def _do_on_generic_first_appt(self, squeeze_factor: float = 0.) -> None:
         """
         """
         # Make top-level reads of information, to avoid repeat accesses.
-        modules: OrderedDict[str, "Module"] = self.sim.modules
-        symptoms = modules["SymptomManager"].has_what(self.target)
+        modules: OrderedDict[str, "Module"] = fixed_modules_order #self.sim.modules
+        filtered_modules = [item for item in modules if item in self.sim.modules]
+
+        symptoms = self.sim.modules["SymptomManager"].has_what(self.target)
```
matt-graham, line 75:
```suggestion
        filtered_and_sorted_modules = [
            self.sim.modules[name] for name in fixed_modules_order if name in self.sim.modules
        ]
        symptoms = self.sim.modules["SymptomManager"].has_what(self.target)
```

src/tlo/methods/hsi_generic_first_appts.py
```diff
@@ -67,14 +69,16 @@ def _do_on_generic_first_appt(self, squeeze_factor: float = 0.) -> None:
...
 
         # Dynamically create immutable container with the target's details stored.
         # This will avoid repeat DataFrame reads when we call the module-level functions.
         patient_details = self.sim.population.row_in_readonly_form(self.target)
-
-        for module in modules.values():
+        for name in filtered_modules:
+            module = self.sim.modules[name]
```
matt-graham, line 81:
```suggestion
        for module in filtered_and_sorted_modules:
```

matt-graham (commented):
This looks correct to me - you could do the filtering and ordering a bit more directly as suggested below, but I think what you have written is equivalent to this so no need to change this if the plan is just to keep this for testing and not merge.

marghe-molaro:
Thank you @matt-graham - jobs submitted, nothing left to do but keep our fingers crossed!

marghe-molaro:
Results of this PR were included in last comment to PR #1389. Order was confirmed to have no significant effect on outputs.

PR #1400, opened 2024-06-11
Reduce the Excessive Demand for TB Appointment Under the 'Max HealthSystem Functionality' Scenario Switch

tdm32:
This PR contains some small fixes which were identified when using max health system function scenarios. One key issue was huge numbers of TB HSIs being scheduled. The following fixes are suggested:

ResourceFile_Improved_Healthsystem_And_Healthcare_Seeking:
- Tb rate_testing_general_pop is removed from this sheet. This is the general (background) screening rate of non-symptomatic people and is currently set to 0.999 meaning every month almost the entire population will be scheduled for TB screening
- NTP2019 changed treatment coverage to 99.99 for all years. This treatment coverage is used to infer the testing rates of active TB cases, assuming perfect referral between testing and treatment
- Schisto delay_till_hsi_a_repeated and delay_until_hsi_b_repeated are both changed from 0.999 to 1. This should be an integer value of delay in days before seeking a repeat test. This will not materially affect results as the numbers required will be low. 
- Hiv hiv_testing_rates are applied to the whole population. This is set to 1 for adults, meaning the whole adult population will test every year. I suggest setting this to either the max value reported (which is 0.39 in 2018) or some slightly higher value, but definitely less than 1.

TB Module:
- HSI_Tb_ScreeningAndRefer has extra condition now to return a blank footprint if person has been tested within last 2 weeks. 

@marghe-molaro This should have an impact on your max health system function scenarios. Let me know if this all makes sense.


tbhallett (approved):
This seems very sensible to me.

The error we're getting is because the value of 1 (integer of days to delay) is being interpreted as a bool.

Issue #1333, opened 2024-05-10
Failing runs automatically resubmitted on Azure 


marghe-molaro:
In job effect_of_capabilities_scaling-2024-04-29T074640Z, two runs (draw_0_run_5 and draw_12-run_8) appear to be failing and automatically re-submitting. All logs are overwritten by the latest attempt, meaning there is no record of the error that caused them to crash.
The job was submitted with --more-memory option and at least one fails quite early, so RAM issue seems unlikely.

tamuri:
Some information \[here\](https://learn.microsoft.com/en-us/azure/batch/error-handling) about how Azure Batch handles errors.

tamuri:
And Batch [task retries and re-execution](https://learn.microsoft.com/en-gb/azure/batch/best-practices#design-for-retries-and-re-execution).

tamuri:
A bit of reading and digging reveals that the "max task retry count" is already defaulted to zero (i.e. no retries) at both job and task levels. However, the docs state (from the link above):

> Although rare, a task can be retried internally due to failures on the compute node, such as not being able to update internal state or a failure on the node while the task is running. The task will be retried on the same compute node, if possible, up to an internal limit before giving up on the task and deferring the task to be rescheduled by Batch, potentially on a different compute node.

It looks like Batch users don't have control over this internal retry limit.

I wonder whether there is some other way we can prempt an out-of-memory error before the task gets killed by the OOM killer. I suspect it's the Docker container that is getting killed so, perhaps, we can put a limit on the maximum memory the container can use (< VM memory), so the `tlo` process running inside the container is killed and we can then put some handling around that.

@giordano - would appreciate your thoughts on this.

tamuri:
Having said that, _this_ particular error was (I believe) due to some fault on the node rather than OOM because it only affected two of the many runs. Would have been useful to be able to control that internal retry count.

giordano:
> This is still worth thinking about to know we've experienced an OOM error.

OOMs are typically annoyingly hard to debug on Linux: unless you have some monitoring which points out memory is skyrocketing, you don't really get many indications from the operating system.  Azure does provide that monitoring service in the portal for running nodes, but for ephemeral nodes that's going to be complicated.

giordano:
[...]

PR #1402, opened 2024-06-12
[batch] Use `earlyoom` to terminate a process using too much memory

giordano:
Linux's built-in system to manage out-of-memory (OOM) errors is unreliable and non-informative, even if it kicks in (and if often it does too late, when the machine is already unresponsive) there's no easy way to tell that a process on a remote machine was terminated because it was using too much memory.  With `earlyoom` we bypass Linux's OOM management and have better control over what's printed to screen in case of error.

Ref: https://github.com/UCL/TLOmodel/issues/1333#issuecomment-2134898483.

Note: opening as draft because I haven't tested this in production yet.


src/tlo/cli.py
```diff
@@ -234,6 +234,25 @@ def batch_submit(ctx, scenario_file, asserts_on, more_memory, keep_pool_alive, i
...
     gzip_pattern_match = "{{txt,log}}"
     command = f"""
+    # Logfile for `earlyoom`
+    EARLYOOM_LOG=/tmp/earlyoom.log
+
+    function terminate_and_log() {{{{
+        # Terminate the `earlyoom` process
+        kill "${{{{EARLYOOM_PID}}}}"
+        # Print line(s) in the log file referencing terminated processes, if any.
+        grep "^sending SIGTERM to process" "${{{{EARLYOOM_LOG}}}}"
```
giordano, line 244: This catches also SIGKILL:
```suggestion
        grep "^sending \w+ to process" "${{{{EARLYOOM_LOG}}}}"
```

src/tlo/cli.py
```diff
@@ -234,6 +234,25 @@ def batch_submit(ctx, scenario_file, asserts_on, more_memory, keep_pool_alive, i
...
+    }}}}
+
+    # Add a trap at exit to terminate `earlyoom` and print relevant log lines.
+    trap terminate_and_log EXIT
+
+    # Delete existing log files, if any.
+    rm -f "${{{{EARLYOOM_LOG}}}}"
+    # Start the `earlyoom` daemon.
+    nohup earlyoom -m 95 -s 100 &> "${{{{EARLYOOM_LOG}}}}" &
```
giordano, line 253: I got the argument `-m 5` backward, the argument is the amount of memory _available_, not _used_:
```suggestion
    # Start the `earlyoom` daemon. Ignore swap usage, terminate process if less than 5% of main memory available.
    nohup earlyoom -m 5 -s 100 &> "${{{{EARLYOOM_LOG}}}}" &
```

PR #1407, opened 2024-06-28
Update ruff check command to fix failing checks

matt-graham:
Ruff now requires [using the `check` subcommand](https://docs.astral.sh/ruff/linter/#ruff-check) rather than running directly without `ruff <paths>`.


matt-graham:
Looks like new Ruff version has also got slightly stricter rules as a type comparison using `==` was now failing in `src/tlo/core.py` - now also fixed.

tamuri (approved):

Issue #1403, opened 2024-06-18
Recurrence of `HSI_Events` in `RTI` module when consumables not available


tbhallett:
In her analysis of the effects of consumables availability, @sakshimohan has noted that there are large numbers of recurring `HSI_Events` for the `RTI` module.

This appears to be due to logic in several of the HSI like this:
```python
# check for availability of consumables 
if consumables_available:
    # do something
else:
   # reschedule this event instance for tomorrow
   return self.make_appt_footprint({})  # return of blank footprint shows that the HSI did not really take up any time
```

e.g. https://github.com/UCL/TLOmodel/blob/4e983a583039844b85fcc7dd5b18f7a49698174c/src/tlo/methods/rti.py#L3856-L3862

This could be fixed by changing the logic in each place to be of the form:

```python
# in __init__() of the HSI_Event
self._number_of_times_this_event_has_run = 0
self._maximum_number_times_event_should_run = self.module.parameters['Maximum_Number_Of_Times_HSI_Events_Should_Run']  # module declares this parameter (value could be 5)

# at the top of the apply() of the HSI_Event
self._number_of_times_this_event_has_run += 1

# check for availability of consumables 
if consumables_available:
    # do something
else:
   if  self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
       # reschedule this event instance for tomorrow
   return self.make_appt_footprint({})


```

PR #1408, opened 2024-07-01
Fixing recurrence of HSIs in RTI

thewati:
Fixing recurring HSI_Events of RTI module. Reworked logic #1403

### DIFF
--- a/resources/ResourceFile_RTI.xlsx
+++ b/resources/ResourceFile_RTI.xlsx
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2c11ada2e8b77675950b61fc8e0efd1c4fa35dffaecaf1029eafd61892a7cefb
-size 13949
+oid sha256:d950c5d769848fb226db8c1a7d7796c8e43cc2590806f846b98a4bbef6840948
+size 13776
--- a/src/tlo/methods/rti.py
+++ b/src/tlo/methods/rti.py
@@ -1530,7 +1530,6 @@ class RTI(Module, GenericFirstAppointmentsMixin):
         # Begin logging the RTI events
         sim.schedule_event(RTI_Logging_Event(self), sim.date + DateOffset(months=1))
 
-
     def rti_do_when_diagnosed(self, person_id):
         """
         This function is called by the generic first appointments when an injured person has been diagnosed
@@ -3853,8 +3852,9 @@ class HSI_RTI_Shock_Treatment(HSI_Event, IndividualScopeEventMixin):
                 get_item_code('Blood, one unit'): 2,
                 get_item_code("Oxygen, 1000 liters, primarily with oxygen cylinders"): 23_040
             }
-            is_cons_available = self.get_consumables(item_codes=
-                  self.module.item_codes_for_consumables_required['shock_treatment_child'])
+            is_cons_available = self.get_consumables(
+                self.module.item_codes_for_consumables_required['shock_treatment_child']
+            )
         else:
             self.module.item_codes_for_consumables_required['shock_treatment_adult'] = {
                 get_item_code("ringer's lactate (Hartmann's solution), 1000 ml_12_IDA"): 2000,
@@ -3862,8 +3862,9 @@ class HSI_RTI_Shock_Treatment(HSI_Event, IndividualScopeEventMixin):
                 get_item_code('Blood, one unit'): 2,
                 get_item_code("Oxygen, 1000 liters, primarily with oxygen cylinders"): 23_040
             }
-            is_cons_available = self.get_consumables(item_codes=
-                  self.module.item_codes_for_consumables_required['shock_treatment_adult'])
+            is_cons_available = self.get_consumables(
+                self.module.item_codes_for_consumables_required['shock_treatment_adult']
+            )
 
         if is_cons_available:
             logger.debug(key='rti_general_message',
--- a/src/tlo/methods/rti.py
+++ b/src/tlo/methods/rti.py
@@ -4779,6 +4779,8 @@ class HSI_RTI_Major_Surgeries(HSI_Event, IndividualScopeEventMixin):
         self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({'MajorSurg': 1})
         self.ACCEPTED_FACILITY_LEVEL = '1b'
         self.BEDDAYS_FOOTPRINT = self.make_beddays_footprint({})
+        self._number_of_times_this_event_has_run = 0
+        self._maximum_number_times_event_should_run = self.module.parameters['maximum_number_of_times_HSI_events_should_run']
 
         p = self.module.parameters
         self.prob_perm_disability_with_treatment_severe_TBI = p['prob_perm_disability_with_treatment_severe_TBI']
@@ -4786,6 +4788,7 @@ class HSI_RTI_Major_Surgeries(HSI_Event, IndividualScopeEventMixin):
         self.treated_code = 'none'
 
     def apply(self, person_id, squeeze_factor):
+        self._number_of_times_this_event_has_run += 1
         df = self.sim.population.props
         rng = self.module.rng
         road_traffic_injuries = self.sim.modules['RTI']
@@ -5058,7 +5061,8 @@ class HSI_RTI_Major_Surgeries(HSI_Event, IndividualScopeEventMixin):
                 ['Treated injury code not removed', self.treated_code]
             df.loc[person_id, 'rt_date_death_no_med'] = pd.NaT
         else:
-            self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
             if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                 df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
             return self.make_appt_footprint({})
--- a/src/tlo/methods/rti.py
+++ b/src/tlo/methods/rti.py
@@ -42,7 +42,6 @@ class RTI(Module, GenericFirstAppointmentsMixin):
         self.resourcefilepath = resourcefilepath
         self.ASSIGN_INJURIES_AND_DALY_CHANGES = None
         self.item_codes_for_consumables_required = dict()
-        self.cons_item_codes = None  # dict containing the item_codes for consumables needed in the HSI Events.
 
     INIT_DEPENDENCIES = {"SymptomManager",
                          "HealthBurden"}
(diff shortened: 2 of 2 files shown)


src/tlo/methods/rti.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, name=None, resourcefilepath=None):
         self.resourcefilepath = resourcefilepath
         self.ASSIGN_INJURIES_AND_DALY_CHANGES = None
         self.item_codes_for_consumables_required = dict()
+        self.cons_item_codes = None  # dict containing the item_codes for consumables needed in the HSI Events.
```
tbhallett, line 45: no need for this change (I don't think, so roll back?)

src/tlo/methods/rti.py
```diff
@@ -1525,6 +1530,107 @@ def initialise_simulation(self, sim):
         sim.schedule_event(RTI_Check_Death_No_Med(self), sim.date + DateOffset(months=0))
         # Begin logging the RTI events
         sim.schedule_event(RTI_Logging_Event(self), sim.date + DateOffset(months=1))
+        # Look-up consumable item codes
```
tbhallett, line 1533: As discussed on slack, you're 100% right about this and it's a good spot. Currently the HSI does the (expensive) lookup of the item codes every time it's called, when it would be more efficient (and more readable) if the consumables are declared once, and all in one go, and the resulting item_codes saved on the module. The way you've set this up in the commented-out section looks right, but I would recommend we put it in a separate PR. So, please could you roll-it back on this branch (so this PR is focussed only on the change we want to make about HSI recurrence), but open another PR that makes that change?

src/tlo/methods/rti.py
```diff
@@ -3825,9 +3931,12 @@ def __init__(self, module, person_id):
         self.TREATMENT_ID = 'Rti_ShockTreatment'
         self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({'AccidentsandEmerg': 1})
         self.ACCEPTED_FACILITY_LEVEL = '1b'
+        self._number_of_times_this_event_has_run = 0
+        self._maximum_number_times_event_should_run = self.module.parameters['maximum_number_of_times_HSI_events_should_run']
```
tbhallett, line 3935: nice. I like the storing of the parameter as a local private variable.

src/tlo/methods/rti.py
```diff
@@ -3845,27 +3954,28 @@ def apply(self, person_id, squeeze_factor):
                 get_item_code('Blood, one unit'): 2,
                 get_item_code("Oxygen, 1000 liters, primarily with oxygen cylinders"): 23_040
             }
-            is_cons_available = self.get_consumables(
-                self.module.item_codes_for_consumables_required['shock_treatment_child']
-            )
+            # cons_available = self.get_consumables(item_codes=self.module.cons_item_codes['shock_treatment_child'])
+            cons_available = self.get_consumables(item_codes=
+                  self.module.item_codes_for_consumables_required['shock_treatment_child'])
```
tbhallett, line 3959: Not sure of the intention of these changes. I think the original is fine, isn't it?

Having read through the whole file, I am wondering if you're proposing this change in the name of the variable (from `is_cons_available` to `cons_available`) , because "is_X_available" sounds like a function that is testing something, whereas "X_available" sounds like the answer (a bool).

Is that right?

If so, then I guess I do agree with that in principle, but I also don't think we have to for this PR, that is focussed on something else, and I think it might confuse others / ourselves in the future, to make that incidental change throughout.

src/tlo/methods/rti.py
```diff
@@ -3845,27 +3954,28 @@ def apply(self, person_id, squeeze_factor):
...
             }
-            is_cons_available = self.get_consumables(
-                self.module.item_codes_for_consumables_required['shock_treatment_adult']
-            )
+            # cons_available = self.get_consumables(item_codes=self.module.cons_item_codes['shock_treatment_adult'])
+            cons_available = self.get_consumables(item_codes=
+                  self.module.item_codes_for_consumables_required['shock_treatment_adult'])
 
-        if is_cons_available:
+        if cons_available:
```
tbhallett, line 3971: if the change above is reversed, then the variable is called `is_cons_available`, so we'd need to roll back this change too.

src/tlo/methods/rti.py
```diff
@@ -3845,27 +3954,28 @@ def apply(self, person_id, squeeze_factor):
...
             logger.debug(key='rti_general_message',
                          data=f"Hypovolemic shock treatment available for person {person_id}")
             df.at[person_id, 'rt_in_shock'] = False
             self.add_equipment({'Infusion pump', 'Drip stand', 'Oxygen cylinder, with regulator', 'Nasal Prongs'})
         else:
-            self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
             return self.make_appt_footprint({})
```
tbhallett, line 3979: perfect

src/tlo/methods/rti.py
```diff
@@ -4017,12 +4131,13 @@ def apply(self, person_id, squeeze_factor):
...
-            self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-            logger.debug(key='rti_general_message',
-                         data=f"Person {person_id} has {fracturecastcounts + slingcounts} fractures without treatment"
-                         )
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4137: I think this needs to be un-indented, so that it's not inside the clause to do with the number of times the HSI has run. i.e. this need to happen, irrespective of whether or not we're rescheduling the HSI

src/tlo/methods/rti.py
```diff
@@ -4094,11 +4213,11 @@ def apply(self, person_id, squeeze_factor):
                     {get_item_code('Metronidazole, injection, 500 mg in 100 ml vial'): 1500}
                 )
         # Check that there are enough consumables to treat this person's fractures
-        is_cons_available = self.get_consumables(
+        cons_available = self.get_consumables(
             self.module.item_codes_for_consumables_required['open_fracture_treatment']
         )
 
-        if is_cons_available:
+        if cons_available:
```
tbhallett, line 4220: Same comment as above re renaming variables.

src/tlo/methods/rti.py
```diff
@@ -4131,12 +4250,13 @@ def apply(self, person_id, squeeze_factor):
...
-            logger.debug(key='rti_general_message',
-                         data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
-                         )
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
+                logger.debug(key='rti_general_message',
+                             data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
+                             )
```
tbhallett, line 4259: as per the comment above, I think this need to be outside of the new clause we're introducing about the rescheduling (or not) the HSI event.

src/tlo/methods/rti.py
```diff
@@ -4222,11 +4347,12 @@ def apply(self, person_id, squeeze_factor):
...
             else:
-                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-                logger.debug(key='rti_general_message',
-                             data="This facility has no treatment for open wounds available.")
+                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                    if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                        df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4353: same comment as above, about moving this outside the new clause

src/tlo/methods/rti.py
```diff
@@ -4305,10 +4435,10 @@ def apply(self, person_id, squeeze_factor):
                     {get_item_code("ringer's lactate (Hartmann's solution), 1000 ml_12_IDA"): 4000}
                 )
 
-            is_cons_available = self.get_consumables(
+            cons_available = self.get_consumables(
                 self.module.item_codes_for_consumables_required['burn_treatment']
             )
-            if is_cons_available:
+            if cons_available:
```
tbhallett, line 4441: same comment as above about not insisting on the renaming.

src/tlo/methods/rti.py
```diff
@@ -4346,11 +4476,12 @@ def apply(self, person_id, squeeze_factor):
...
             else:
-                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-                logger.debug(key='rti_general_message',
-                             data="This facility has no treatment for burns available.")
+                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                    if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                        df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4482: same comment as above about moving this outside the new clause.

src/tlo/methods/rti.py
```diff
@@ -4397,16 +4533,17 @@ def apply(self, person_id, squeeze_factor):
             self.module.item_codes_for_consumables_required['tetanus_treatment'] = {
                 get_item_code('Tetanus toxoid, injection'): 1
             }
-            is_tetanus_available = self.get_consumables(
+            tetanus_available = self.get_consumables(
                 self.module.item_codes_for_consumables_required['tetanus_treatment']
             )
-            if is_tetanus_available:
+            if tetanus_available:
                 logger.debug(key='rti_general_message',
```
tbhallett, line 4540: no need to rename?

resources/ResourceFile_RTI.xlsx
tbhallett, line 1: perfect

src/tlo/methods/rti.py
```diff
@@ -3845,27 +3954,28 @@ def apply(self, person_id, squeeze_factor):
...
                 get_item_code('Cannula iv  (winged with injection pot) 18_each_CMST'): 1,
                 get_item_code('Blood, one unit'): 2,
                 get_item_code("Oxygen, 1000 liters, primarily with oxygen cylinders"): 23_040
             }
-            is_cons_available = self.get_consumables(
-                self.module.item_codes_for_consumables_required['shock_treatment_adult']
-            )
+            # cons_available = self.get_consumables(item_codes=self.module.cons_item_codes['shock_treatment_adult'])
+            cons_available = self.get_consumables(item_codes=
+                  self.module.item_codes_for_consumables_required['shock_treatment_adult'])
```
tbhallett, line 3969: same comment as above about renaming of variables. (i.e. suggest to roll it back -- unless you feel very strongly?)

tbhallett (changes requested):
GREAT work, thanks very much.

src/tlo/methods/rti.py
thewati: This was for the change I had suggested. I should have removed it. I will roll back

src/tlo/methods/rti.py
thewati: I think the original `is_cons_available` is better. I had only changed it because your suggestion over [here](https://github.com/UCL/TLOmodel/issues/1403) used `consumables_available`. So, I thought you wanted me to change that too. In this case, I will roll it back

src/tlo/methods/rti.py
```diff
@@ -4017,12 +4028,13 @@ def apply(self, person_id, squeeze_factor):
...
-            self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-            logger.debug(key='rti_general_message',
-                         data=f"Person {person_id} has {fracturecastcounts + slingcounts} fractures without treatment"
-                         )
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4034:
```suggestion
            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```

src/tlo/methods/rti.py
```diff
@@ -4131,12 +4147,13 @@ def apply(self, person_id, squeeze_factor):
...
-            logger.debug(key='rti_general_message',
-                         data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
-                         )
+            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
+                logger.debug(key='rti_general_message',
+                             data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
+                             )
```
tbhallett, line 4156:
```suggestion
            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
            logger.debug(key='rti_general_message',
                             data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
                             )    
```

src/tlo/methods/rti.py
```diff
@@ -4222,11 +4244,12 @@ def apply(self, person_id, squeeze_factor):
...
             else:
-                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-                logger.debug(key='rti_general_message',
-                             data="This facility has no treatment for open wounds available.")
+                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                    if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                        df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4250:
```suggestion
                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```

src/tlo/methods/rti.py
```diff
@@ -4346,11 +4373,12 @@ def apply(self, person_id, squeeze_factor):
...
             else:
-                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
-                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
-                logger.debug(key='rti_general_message',
-                             data="This facility has no treatment for burns available.")
+                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                    if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
+                        df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```
tbhallett, line 4379:
```suggestion
                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
                if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                    df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
```

src/tlo/methods/rti.py
```diff
@@ -4404,9 +4437,10 @@ def apply(self, person_id, squeeze_factor):
                 logger.debug(key='rti_general_message',
                              data=f"Tetanus vaccine requested for person {person_id} and given")
             else:
-                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
-                logger.debug(key='rti_general_message',
-                             data=f"Tetanus vaccine requested for person {person_id}, not given")
+                if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
+                    self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
+                    logger.debug(key='rti_general_message',
+                                 data=f"Tetanus vaccine requested for person {person_id}, not given")
```
tbhallett, line 4443:
```suggestion
                logger.debug(key='rti_general_message', data=f"Tetanus vaccine requested for person {person_id}, not given")
```

tbhallett (commented):
Thanks again @thewati 

I just made those few other changes about the positioning of some of those places where there is writing the sim.populations.props dataframe. I think these changes ensure that we preserve the logic that was originally intended. Let me know if it looks right to you....?

src/tlo/methods/rti.py
thewati: Looks logical. Will implement and test it

src/tlo/methods/rti.py
thewati: @tbhallett I wanted your clarity on the `logger.debug()` upon reconsideration. Shouldn't `logger.debug()` be inside the `if pd.isnull()` condition. This implies if a person's `rt_date_death_no_med` is null, then we set the date and the logger message follows.  For instance, the code would look like this:

```suggestion
            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
                logger.debug(key='rti_general_message',
                             data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
                             )
```

thewati: @tbhallett I wanted your clarity on the `logger.debug()` upon reconsideration. Shouldn't `logger.debug()` be inside the `if pd.isnull()` condition. This implies if a person's `rt_date_death_no_med` is null, then we set the date and the logger message follows.  For instance, the code would look like this:

```suggestion
            if self._number_of_times_this_event_has_run < self._maximum_number_times_event_should_run:
                self.sim.modules['RTI'].schedule_hsi_event_for_tomorrow(self)
            if pd.isnull(df.loc[person_id, 'rt_date_death_no_med']):
                df.loc[person_id, 'rt_date_death_no_med'] = self.sim.date + DateOffset(days=7)
                logger.debug(key='rti_general_message',
                             data=f"Person {person_id}'s has {open_fracture_counts} open fractures without treatment",
                             )
```

tbhallett: Yes, you're right!!! Great spot.

I'll push commit to implement that change now

tbhallett:
Thanks again @thewati  -- this looks good to me now.

src/tlo/methods/rti.py
thewati: Right! I also had at the master code. I guess we can leave it and let it resemble the master in that case

PR #1409, opened 2024-07-01
Change logging level of message about footprint not being possible.

tbhallett:
This should prevent this message being included in logs many, many times.


matt-graham (approved):
Seems good to me! Sorry for the delay in reviewing

PR #1412, opened 2024-07-04
Check to see if the footprint created for an appointment is blank. 

### LINKED ISSUE #1401: possible inconsistencies in using get_blank_appt_footprint()
Currently get_blank_appt_footprint() is used within some modules (HIV, TB etc) when a HSI has been scheduled but not resources are actually occupied, e.g. person is not eligible for test or is already on treatment. These HSIs are logged as having been delivered and therefore contribute to the total numbers of HSIs in outputs, even though no person-time / consumables are required.

When no resources are required, should a return statement be used so that this HSI is not logged? The current set-up may be skewing results where HSI numbers requested / delivered are used.

@tbhallett @marghe-molaro What do you think?

RachelMurray-Watson:
Fixes #1401  
Add new logging keys in the healthsystem's "summary" logger for HSI Events that do have a non-blank footprint. (The existing log captures all HSI Events that run, including those that have a blank footprint (e.g. the generic first appointments).


src/tlo/methods/healthsystem.py
```diff
@@ -1725,7 +1725,7 @@ def write_to_hsi_log(
             },
             description="record of each HSI event"
         )
-        if did_run:
+        if did_run and event_details.appt_footprint:
```
tbhallett, line 1728: does this really check if the appt_footprint is blank (i.e. an empty dictionary, or a dictionary for which all values are 0)?

tbhallett (commented):
GREAT!

One comment.

If you want to prove your point, the next thing to do would be to add / elaborate one of the tests (see `test_healthsystem.py`)

src/tlo/methods/healthsystem.py
RachelMurray-Watson: It checks if it is an empty dictionary - or, least I think it does, in all of my tests. Would it also need to check if it is zero?

tbhallett: this is something I'm not quire sure about. 
Is it the case that using the following three things to express a blank footprint from the HSI_Event are all equivalent?

* `return sim.modules['HealthSystem'].get_blank_footprint()`
* `return self.make_appt_footprint({})`
* `return {}`

If not, we need to make sure that the logic at this place interprets them as such.

I think the code as written is checking for 'truthiness' of the footprint, and a footprint that has all values of 0's (which we think of as being blank), would seem truth, and that therefore, this would not work as intended.

RachelMurray-Watson: Ah okay, I think that I was under the impression from last weeks meeting that all the zeroes were seen as equivalent, as we only wanted to track what actually had a non-zero footprint.

tbhallett: the appt-footprint is always a dict. If all the values in that dict are zero, we say it's "blank".

But that's different to the python concept of it being 'falsey' versus 'truthy'.

So I think this check should be something like
```python
if did_run and sum(event_details.appt_footprint.values():
```
instead of 
```python
if did_run and event_details.appt_footprint
```

RachelMurray-Watson: Ah okay, understood

RachelMurray-Watson: So it looks like not all of the footprints are, in fact, dictionaries? Signalling a wider issue, or is it okay?

image

tbhallett: are you able to find which HSI returned that, and what is returned?

RachelMurray-Watson: Yes, I think it's coming from run_individual_level_events_in_mode_0_or_1 in healthsystems.py... Seems to be the common thread connecting all of the test failures. 

Could then either convert the tuples to dictionaries when logging or add another check for tuples? 

image

tbhallett: I've had a look at this more closely, and I can see that the property being addressed here `event_details` is actually passed in as a NamedTuple (from [here](https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/healthsystem.py#L1692-L1698)).

The conversion to NamedTuple is done [here](https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hsi_event.py#L406-L413).

Sorry to send you on a wild goose chase! 

This actually simplifies things, because I believe we can access all the values of NamedTuple using `list`. So therefore the check for "blankness" would be `sum(list(event_details.appt_footprint))`

RachelMurray-Watson: Ah okay, I was a bit behind you tracking it down. But I'll make that change now. 

Thank you for investigating!!

tbhallett:
@thewati  -- When @RachelMurray-Watson is finished with this, please could you provide a review?

src/tlo/methods/healthsystem.py
```diff
@@ -1697,6 +1715,14 @@ def record_hsi_event(self, hsi_event, actual_appt_footprint=None, squeeze_factor
                 did_run=did_run,
                 priority=priority,
             )
+            #if did_run and len(actual_appt_footprint) != 0 and not all(x == 0 for x in actual_appt_footprint):
```
thewati, line 1718: This commented out code can be removed

src/tlo/methods/healthsystem.py
```diff
@@ -1767,6 +1794,44 @@ def call_and_record_never_ran_hsi_event(self, hsi_event, priority=None):
...
 
+    def write_to_hsi_no_blank_footprint_log(
+        self,
+        event_details: HSIEventDetails,
+        person_id: int,
+        facility_id: Optional[int],
+        did_run: bool,
+        priority: int,
+    ):
+        """Write the log `HSI_Event` and add to the summary counter only if the appointment footprint is not blank."""
```
thewati, line 1805: The docstring of function `write_to_hsi_no_blank_footprint_log()` needs to be updated with a description of each of the parameters. I have given a rough example below:
```suggestion
        """Write the log `HSI_Event` and add to the summary counter only if the appointment footprint is not blank.
        :param event_details: What event_details is responsible for
        :param person_id: The person_id of interest
        :param facility_id: What this is responsible for
        :param did_run: What this is responsible for
        :param priority: The priority of for the HSI event. Highest is..., Lowest is...."""
```

src/tlo/methods/healthsystem.py
```diff
@@ -1924,18 +1989,33 @@ def _write_never_ran_hsi_event_counts_to_log_and_reset(self):
         self._never_ran_hsi_event_counts_cumulative += self._never_ran_hsi_event_counts_log_period
         self._never_ran_hsi_event_counts_log_period.clear()
 
+    def _write_hsi_event_no_blank_counts_to_log_and_reset(self):
```
thewati, line 1992: Need to add a doctring explaining what function `_write_hsi_event_no_blank_counts_to_log_and_reset()` does

src/tlo/methods/healthsystem.py
```diff
@@ -2697,6 +2824,22 @@ def record_never_ran_hsi_event(self,
             self._never_ran_appts[appt_type] += number
             self._never_ran_appts_by_level[level][appt_type] += number
 
+    def record_hsi_event_no_blank_appt_footprints(self,
+                                                  treatment_id: str,
+                                                  hsi_event_name: str,
+                                                  appt_footprint: Counter,
+                                                  level: str
+                                                  ) -> None:
+        """Add information about a`HSI_Event` that has a non-blank appointment footprint to the running summaries."""
```
thewati, line 2833: Similar to my comment above. This docstring also needs to include descriptions for each of the parameters
```suggestion
        """Add information about a`HSI_Event` that has a non-blank appointment footprint to the running summaries.
        :param treatment_id: .....
        :param hsi_event_name: .....
        :param appt_footprint: .....
        :param level: ....."""
```

src/tlo/methods/healthsystem.py
```diff
@@ -2737,6 +2880,17 @@ def write_to_log_and_reset_counters(self):
                 "Number_By_Appt_Type_Code_And_Level": self._never_ran_appts_by_level,
             },
         )
+            # Log summary of HSI_Events that do not have a blank appointment footprint
```
thewati, line 2883: Re-align comment to be inline with logger_summary.info(
```suggestion
            # Log summary of HSI_Events that do not have a blank appointment footprint
            logger_summary.info(
```

thewati (changes requested):
I have added comments to `healthsystem.py`. The other file `test_blank_footprint.py` looks good

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,218 @@
...
+                # is not at all available.
+                self.did_run = True
+        def make_blank_appt_footprint(self):
+            return {}
+
+    hsi_dummy = HSI_Dummy_with_blank(module=module, person_id=0)
+    hsi_dummy.initialise()
+    hsi_dummy.facility_info = module.sim.modules['HealthSystem']._facility_by_facility_id[facility_id]
+    return hsi_dummy
+
```
thewati, line 141: We can introduce a new function here to set the person and district since this seems to be called repeatedly at least in 2 spots. More on this in the comment below. Please note that you may change the function name as you see fit. However, the function can be as follows:
```suggestion
def set_person_district_id(sim):
      # sets facility and person to 0
      sim.population.props.at[0, 'district_of_residence'] = mfl.set_index('Facility_ID').loc[0].District
      return sim
```

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,218 @@
...
+            facility_ids=[0]),
+        tmpdir=tmpdir,
+        run=False
+    )
+
+    # Edit the `initialise_simulation` method of DummyModule so that, during the simulation, an HSI is run.
+    def schedule_hsi(sim):
+        """Drop-in replacement for `initialise_simulation` in the DummyModule module."""
+        # Make the district for person_id=0 such that the HSI will be served by facility_id=0
+        sim.population.props.at[0, 'district_of_residence'] = mfl.set_index('Facility_ID').loc[0].District
```
thewati, line 160: This line has also been repeated on line 200 in the function called `test_outputs_to_log_blank()`. I would suggest that we bundle this repetition in one function as seen in the comment above and then just call it when needed. We can simply call it as follows.

 ```suggestion
        sim = set_person_district_id(sim)
```

File: tests/test_blank_footprint.py [hunk 9]
```diff
@@ -0,0 +1,218 @@
...
```
thewati, line 200: This line can also be replaced with the function call:
```suggestion
        sim = set_person_district_id(sim)
```

thewati (changes requested):
Suggested some changes to the test file

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+    sim = get_sim_with_dummy_module_registered(
+        data=create_dummy_data_for_cons_availability(
+            intrinsic_availability=intrinsic_availability,
+            months=[1],
+            facility_ids=[0]),
+        tmpdir=tmpdir,
+        run=False
+    )
+
+        # Edit the `initialise_simulation` method of DummyModule so that, during the simulation, an HSI is run.
```
thewati, line 159: Need to indent this comment. It must be the way you indented it in the `test_outputs_to_log_blank()` function. Must be inline with `def schedule_hsi(sim):`
```suggestion
   # Edit the `initialise_simulation` method of DummyModule so that, during the simulation, an HSI is run.
```

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+            hsi_event=get_dummy_hsi_event_instance(module=sim.modules['DummyModule'], facility_id=0, blank = blank_footprint),
+            topen=sim.start_date,
+            tclose=None,
+            priority=0
+        )
+
+
+    sim.modules['DummyModule'].initialise_simulation = schedule_hsi
+
+        # Simulate for one day
```
thewati, line 176: The same applies here... Must be inline with `sim.sumulate()`
```suggestion
     # Simulate for one day
```

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+    sim.simulate(end_date=sim.start_date + pd.DateOffset(days=1))
+
+        # Check that log is created and the content is as expected.
```
thewati, line 179: Here too. Must be inline with `hsi_log`
```suggestion
     # Check that log is created and the content is as expected.
```

thewati (changes requested):
I believe these should be last changes to make

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+
+    sim.make_initial_population(n=100)
+
+    if run:
+        sim.simulate(end_date=start_date)
+
+    return sim
+
+
+def get_dummy_hsi_event_instance(module, facility_id=None, blank = False):
```
tbhallett, line 82: note 'blank' kwarg not being used

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+                self.did_run = True
+
+
+    hsi_dummy = HSI_Dummy_no_blank(module=module, person_id=0)
+    hsi_dummy.initialise()
+    hsi_dummy.facility_info = module.sim.modules['HealthSystem']._facility_by_facility_id[facility_id]
+    return hsi_dummy
+
+
+def get_dummy_hsi_event_instance_blank_footprint(module, facility_id=None):
```
tbhallett, line 115: did you mean to collapse these two functions (get_dummy_hsi_event_instance_blank_footprint and get_dummy_hsi_event_instance_no_blank) and use the 'blank' kwarg to swtich between them?

tests/test_blank_footprint.py
```diff
@@ -0,0 +1,221 @@
...
+            tclose=None,
+            priority=0
+        )
+
+    sim.modules['DummyModule'].initialise_simulation = schedule_hsi
+
+    # Simulate for one day
+    sim.simulate(end_date=sim.start_date + pd.DateOffset(days=1))
+
+    # Check that log is created and the content is as expected.
```
tbhallett, line 218: I think we want to check that an HSI that has a blank footprint:
* does appear in the log key `hsi_event_details`
* does NOT appear in the log key `hsi_event_no_blank_footprint`

src/tlo/methods/healthsystem.py
```diff
@@ -1772,6 +1798,52 @@ def call_and_record_never_ran_hsi_event(self, hsi_event, priority=None):
...
+               :param TREATMENT_ID: the treatment ID of the HSI that ran
+               :param Number_By_Appt_Type_Code: dictionary with number of HSIs types
+               :param Person_ID:  A sequence of person ID index values to use as the targets of the HSI events being scheduled
+               :param priority: The priority for the HSI events: 0 (highest), 1 or 2 (lowest). Either a single value for all events or an iterable of per-target values.
+               :param did_run: Boolean asserting if the HSI event did happen
+               :param Facility_Level: The level of the facility where the HSI ran. Can be 0, 1a, 1b, 2, 3, 4, or 5.
+               :param Facility_id: Facility where the HSI ran
+               :param Equipment: The set of equipment that is used in the HSI.
+               """
+        logger.debug(
```
tbhallett, line 1820: why do we need the debug logger?

src/tlo/methods/healthsystem.py
```diff
@@ -1772,6 +1798,52 @@ def call_and_record_never_ran_hsi_event(self, hsi_event, priority=None):
...
+                'Person_ID': person_id,
+                'priority': priority,
+                'did_run': did_run,
+                'Facility_Level': event_details.facility_level if event_details.facility_level is not None else -99,
+                'Facility_ID': facility_id if facility_id is not None else -99,
+                'Equipment': sorted(event_details.equipment),
+            },
+            description="record of each HSI event"
+        )
+        if did_run and len(event_details.appt_footprint) != 0 and not all(x == 0 for x in event_details.appt_footprint):
```
tbhallett, line 1835: this logic could go inside `write_to_hsi_log` couldn't it?
It would avoid writing the above section, which are repetitious.

src/tlo/methods/healthsystem.py
```diff
@@ -787,6 +795,7 @@ def on_simulation_end(self):
         if self._hsi_event_count_log_period == "simulation":
             self._write_hsi_event_counts_to_log_and_reset()
             self._write_never_ran_hsi_event_counts_to_log_and_reset()
+            self._write_hsi_event_no_blank_counts_to_log_and_reset()
```
tbhallett, line 798: no needed (I don't think) as these are to do with the full logger, that captrues the exact footprint anyway.

src/tlo/methods/healthsystem.py
```diff
@@ -2599,6 +2715,22 @@ def apply(self, population):
                     priority=-1,
                     did_run=True,
                 )
+                if len(_inpatient_appts) != 0:
+                    self.module.write_to_hsi_no_blank_footprint_log(
```
tbhallett, line 2719: to avoid this long (and mostly repetitive call), I think it would be neater to let the logic be inside the existing function (write_to_hsi_log), That way, we have one place where we're doing the split between blank and non-blank footprints.

src/tlo/methods/healthsystem.py
```diff
@@ -1958,6 +2050,7 @@ def on_end_of_year(self) -> None:
         if self._hsi_event_count_log_period == "year":
             self._write_hsi_event_counts_to_log_and_reset()
             self._write_never_ran_hsi_event_counts_to_log_and_reset()
+            self._write_hsi_event_no_blank_counts_to_log_and_reset()
```
tbhallett, line 2053: not needed

src/tlo/methods/healthsystem.py
```diff
@@ -2150,6 +2243,29 @@ def never_ran_hsi_event_counts(self) -> Counter:
                 }
             )
 
+    def hsi_no_blank_footprint_event_counts(self) -> Counter:
```
tbhallett, line 2246: _write_hsi_event_no_blank_counts_to_log_and_reset

tbhallett:
Thanks for the work on this @RachelMurray-Watson  and the super through reviewing @thewati 

It's nice to see this working nicely. I have some minor comments about the niceties of structuring the test, and making it test more concretely that for a HSI that has a blank footprint, it's logged in one place in the summary logger and not another.

But, main comment is that I think the changes do too much.

It's quite confusing but there are two loggers in this module: `tlo.methods.healthsystem` and `tlo.methods.healthsystem.summary`. The former has a key "HSI_Event" that logs every HSI that occurs. It has a compacted version of it too, whereby we count the occurrences of unique HSI_Events. We don't need to worry about those at all though, for this change-- because they're already recording the footprint that is associated with the HSI Event.

So, this change is exclusively concerned with the summary logger. This is managed by `HealthSystemSummaryCounter` and  has a single entry point `record_hsi_event`.

I think, therefore, that we can radically simplify this:

* Remove all the changes that are not concerned with the `HealthSystemSummaryCounter`
* Do not add another entry point to `HealthSystemSummaryCounter`. Instead, just making the changes inside `record_hsi_event`. (This way, we limit our logical changes to that class and not the wider module.)
* Do this by making `record_hsi_event_no_blank_appt_footprints` a private function that is called from `record_hsi` _if_ the footprint is blank. (You might wish to factor out the check of blankness of the footprint into a static member function of `HealthSystemSummaryCounter`, as that is the critical piece.)

src/tlo/methods/healthsystem.py
RachelMurray-Watson: I included it as it was also included in the "write_to_never_ran_hsi_log", but I can remove it

src/tlo/methods/healthsystem.py
```diff
@@ -2150,6 +2163,29 @@ def never_ran_hsi_event_counts(self) -> Counter:
...
```
tbhallett, line 2161: this function is never called!?

tbhallett:
@thewati  --- resolved some of your comments on this as @RachelMurray-Watson 's done them now, or the function was removed (so no doctoring needed).

tbhallett (approved):

tbhallett:
Thanks @RachelMurray-Watson  - I'm happy with this.

@thewati  - please could you take another look and approve if you're happy too?

tests/test_healthsystem.py
```diff
@@ -2570,3 +2570,67 @@ def get_capabilities(yearly_scaling: bool, scaling_by_level: bool, rescaling: bo
...
+
+        return parse_log_file(sim.log_filepath)['tlo.methods.healthsystem.summary']
+    # When the footprint is blank:
+    log = run_simulation_and_return_healthsystem_summary_log(tmpdir, blank_footprint=True)
+    assert log['HSI_Event']['TREATMENT_ID'].iloc[0] == {'Dummy': 1}  # recorded in usual logger
+    assert log['HSI_Event_non_blank_appt_footprint']['TREATMENT_ID'].iloc[0] == {}  # not recorded in 'non-blank' logger
+
+    # When the footprint is non-blank:
+    log = run_simulation_and_return_healthsystem_summary_log(tmpdir, blank_footprint=False)
+    assert (
```
thewati, line 2631: I am just wondering if its worth adding checks on the structure of the dataframe and the presence of HSIs like below:
```suggestion
    assert not log['HSI_Event'].empty
    assert 'TREATMENT_ID' in log['HSI_Event'].columns
    assert 'TREATMENT_ID' in log['HSI_Event_non_blank_appt_footprint'].columns
```
Not sure about this. What do you think?

thewati (commented):
All looks good, except for the one comment I have added.

tests/test_healthsystem.py
RachelMurray-Watson: Yes, I'm happy to add it if you think it's needed!

tbhallett:
@thewati --- if you're happy with this, please re-review and approve. thanks

thewati (approved):

PR #1413, opened 2024-07-04
Updates to HTM Swtiching Method to Fix Issue #1411

### LINKED ISSUE #1411: HTM Scale-up Not Having The Effect Expected
Whereas, the test script and the initial scenario run showed that the HTM scale-up was working as expected, it appears that the in practice they do not (see [here](https://github.com/UCL/TLOmodel/pull/1288#issuecomment-2195159453)).

The [test script](https://github.com/UCL/TLOmodel/blob/master/tests/test_htm_scaleup.py) and the [scenario](https://github.com/UCL/TLOmodel/blob/master/src/scripts/htm_scenario_analyses/scenario_plots.py) both examine scale-up occurring instantly in year 2012, whereas the scale-up scenarios we want for the final analysis is for scale-up instantly in 2019.

Looking at this, I can see a few reasons why this might be occurring:

* We write to a parameter that does not exist: https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1131-L1132

* Updating of the rates of testing does not work as expected, as the values for years after 2020 are set to nan, and note that future years are never read due to [this](https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1864-L1868) logic.
https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1124-L1125

(I don't think [this section](https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1137) suffers from the same issue - but it looks a bit fragile, for similar reasons).

* We write new values for parameter values, but we do not rebuild the `LinearModels` that use them (the linear model stores the value of parameter when it is built, and does not update when those parameters are updated later)
https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1113C13-L1113C52
https://github.com/UCL/TLOmodel/blob/e9263803787825168955b0e24ed15c383a2ba0dd/src/tlo/methods/hiv.py#L1121

There may be other issues in the Tb and malaria codes of a similar nature.

I would propose doing the following:

- [ ] Update the scenario file so that it reflects a scale-up in 2019, running at scale, and confirming that this also shows there being no epidemiological impact. (If this doesn't confirm it, then I have had some mistakes in [my Scenario file](https://github.com/UCL/TLOmodel/blob/60c9db0a99d9d60c643c1f5493f56fdd16cd0814/src/scripts/comparison_of_horizontal_and_vertical_programs/scenario_vertical_programs_with_and_without_hss.py) and in these observations)
- [ ] Searching for possible issues in hiv.py, tb.py and malaria.py and fixing
- [ ] Re-running the scenario
- [ ] Raising PR with fixes

tbhallett:
This fixes a problem with the implementation of the scale-up of HIV, TB and Malaria programs whereby the LinearModels were not rebuilt following parameter updates during the simulation. Now we rebuild the models after updating the model parameters.

We hoped that would fix #1411 but it doesn't quite do that (see below).

But, we do want it merged in anyway, as these are important fixes.


src/tlo/methods/enhanced_lifestyle.py
```diff
@@ -660,7 +660,7 @@ def update_all_properties(self, df):
         :param df: The population dataframe """
         # get months since last poll
         now = self.module.sim.date
-        months_since_last_poll = round((now - self.date_last_run) / np.timedelta64(1, "M"))
+        months_since_last_poll = round((now - self.date_last_run) / np.timedelta64(1, "m"))
```
tbhallett, line 663: is that right?

src/tlo/methods/hiv.py
```diff
@@ -1117,25 +1117,27 @@ def update_parameters_for_program_scaleup(self):
...
-            # prob ART start if dx, this is already 95% at 2020
-            p["prob_start_art_after_hiv_test"] = scaled_params["prob_start_art_after_hiv_test"]
-
             # viral suppression rates
             # adults already at 95% by 2020
             # change all column values
             p["prob_start_art_or_vs"]["virally_suppressed_on_art"] = scaled_params["virally_suppressed_on_art"]
 
+            # update exising linear models to use new scaled-up paramters
+            self.pre_initialise_population()
```
tbhallett, line 1139: good logic. I think I'll factor it out into `self.build_lms()` (that is called from `self.pre_initialise_population` in case there are other things that need to go into `self.pre_initialise_population` that we don't want run at this point.

tbhallett:
`job_id` = `htm_with_and_without_hss-2024-07-04T164311Z`

tbhallett:
Results from running analysis script show much the same results

Additional_Deaths_Averted_vs_Baseline_2020-2030
All_programs
All_Scenarios:_DALYs_2020-2030
DALYs_Averted_vs_Baseline_2020-2030
Deaths_2020-2030
HIV_programs
Malaria_programs
[scenario_vertical_programs_with_and_without_hss_draws.json](https://github.com/user-attachments/files/16126938/scenario_vertical_programs_with_and_without_hss_draws.json)
TB_programs

tbhallett:
@tdm32 -- I've made some cleaning-up changes on here, and I think is ready to go in following your review.

src/tlo/methods/enhanced_lifestyle.py
tbhallett: rolled back. Coding for month is "M", which is what we want and what was there originally. ("m" is the the code for minutes".

https://numpy.org/doc/stable/reference/arrays.datetime.html

src/tlo/methods/hiv.py
tbhallett: I've refactored this

src/tlo/methods/enhanced_lifestyle.py
tdm32: I don't know why but this causes errors on my machine! It won't run with "M"... I'll check out why this is happening when I'm back.

tbhallett:
Next job is to run a simplified version of the analysis, but...
* In Mode 1 (so we can see if it is the "mode 2 constraints" that is hampering the scale-up
* With the logs from HIV, TB and Malaria modules turned on, so that we can see if the service coverages are increasing as expected.

tdm32 (approved):

PR #1414, opened 2024-07-05
The Impact of Past Improvements in HRH: The Historical Analysis

tbhallett:
This PR does the analysis to estimate the impact of health of the increases in Human Resources for Health (healthcare workers) have occurred already in Malawi.

Dominic Nkhoma has provided data on the number of staff (4th July 2024), saved in the team Dropbox.

Basic descriptions of these data are shown below (done in the script `src/scripts/data_file_processing/healthsystem/human_resources/analysis_historic_changes_in_hr.py`)

image

image

PR #1415, opened 2024-07-06
Analysis of expanding HCW by cadre

BinglingICL:
In this branch, we analyse the marginal and combined impact (in terms of DALYs, Deaths, Services by appointment types/treatment types) of expanding current HCW capabilities for five cadres - Clinical (C), DCSA (D), Nursing and Midwifery (N), Pharmacy (P), Other (O) - given an annual extra budget that is 4.2% of total cost of these cadres of the previous year. These main cadres are among the priorities cadres (HSSP III HRH Annex) and tlo model has fairly simulated their time usage (not sure about DCSA). The 4.2% growth rate is referred to the annual GDP growth rate in Malawi (Margherita 2024), which is 4.2%, and the assumption that proportion of GDP allocated for HRH costing is fixed.

The HRH expansion path in the simulation is as below:
full expansion path

Todo:
(1.1) Find out why some scenarios have less services by treatments and appointments but more DALYs averted compared to others -> Only tell by the difference of HCW expansion scenarios, but not yet clear how it exactly links to services delivered and DALYs averted. May need analyses of individual appts and treatments and causes of DALYs.  (**The reason should be multifactorial and possible trade-off between cadres, appointment types and treatments types, which might not directly indicate the benefits on DALYs averted. Possible factors: Clinical cadre, Pharmacy cadre, Pharm Dispensing appointment, etc**).

(1.2) Sensitivity analysis

(1.3) Random sampling of C/D/P/N/O cadres to test the regression model in the main analysis?

(2.0) Once the analysis on national level is completed, consider district-level analysis and collaborate with relevant colleagues from the team and Malawi


BinglingICL:
> 1. Confirm the scenarios before summit to Azure (C + P or C+NM+P, scale up ratios, time of change)

to do for scenario design: Use the GDP expenditure on HR and HR cost (salary) data to figure out given a % increase of HR expenditure, how to increase the staff of cadres Clinical, Nursing and Midwifery, Pharmacy and DCSA, both independently or in a combined way. The ways to increase staff will be the scenarios to be examined.

BinglingICL:
Hi @marghe-molaro and @tbhallett, I have created a hr expansion function in the healthsystem module, considering the fractions of extra budget to be used by the four cadres (clinical, dcsa, nursing_and_midwifery, pharmacy), updated the scenarios and scenario script, and added some relevant files as needed. I would be very grateful if you could help check if the function and logics  are looking fine. Any comment or advice would be greatly appreciated. Thanks.

BinglingICL:
Note: I have got two warnings when run the script locally. Not sure if it is related to changes in this branch. Will check.
C:\Users\jdbb1\PycharmProjects\TLOmodel\src\tlo\scenario.py:438: InconsistentLoggedColumnsWarning: Inconsistent columns in logged values for tlo.scenario logger with key override_parameter compared to header generated from initial log entry:
  Columns in header not in logged values are
  {'new_value': 'bool', 'old_value': 'bool'}
  Columns in logged values not in header are
  {'new_value': 'str', 'old_value': 'str'}
  logger.info(
C:\Users\jdbb1\PycharmProjects\TLOmodel\src\tlo\scenario.py:438: InconsistentLoggedColumnsWarning: Inconsistent columns in logged values for tlo.scenario logger with key override_parameter compared to header generated from initial log entry:
o logger with key override_parameter compared to header generated from initial log entry:
  Columns in header not in logged values are
  {'new_value': 'bool', 'old_value': 'bool'}
  Columns in logged values not in header are
  {'new_value': 'int', 'old_value': 'int'}
  logger.info(

src/tlo/methods/healthsystem.py
```diff
@@ -2930,6 +2972,72 @@ def apply(self, population):
...
+
+        # get proportional daily extra budget for each of the four cadres
+        daily_extra_budget_by_officer = [
+            daily_extra_budget * i for i in self.module.parameters['HR_expansion_by_officer_type']]
+
+        # get the scale up factor for each cadre, assumed to be the same for each facility id of that cadre
+        sf_clinical = (total_cost_clinical + daily_extra_budget_by_officer[0])/total_cost_clinical
+        sf_dcsa = (total_cost_dcsa + daily_extra_budget_by_officer[1]) / total_cost_dcsa
+        sf_nursing = (total_cost_nursing + daily_extra_budget_by_officer[2]) / total_cost_nursing
+        sf_pharmacy = (total_cost_pharmacy + daily_extra_budget_by_officer[3]) / total_cost_pharmacy
```
BinglingICL, line 3021: Might create a logger for these scale up factors for years 2020-2030; the internally calculated scale up factors could be compared with those calculate externally (will write an individual script for this calculation).

BinglingICL:
> Note: I have got warnings when run the script locally. Not sure if it is related to changes in this branch. Will check.
> 
> C:\Users\jdbb1\PycharmProjects\TLOmodel\src\tlo\scenario.py:438: InconsistentLoggedColumnsWarning: Inconsistent columns in logged values for tlo.scenario logger with key override_parameter compared to header generated from initial log entry: Columns in header not in logged values are {'new_value': 'bool', 'old_value': 'bool'} Columns in logged values not in header are {'new_value': 'list', 'old_value': 'list'} logger.info( C:\Users\jdbb1\PycharmProjects\TLOmodel\src\tlo\scenario.py:438: InconsistentLoggedColumnsWarning: Inconsistent columns in logged values for tlo.scenario logger with key override_parameter compared to header generated from initial log entry: Columns in header not in logged values are {'new_value': 'bool', 'old_value': 'bool'} Columns in logged values not in header are {'new_value': 'str', 'old_value': 'str'} logger.info( C:\Users\jdbb1\PycharmProjects\TLOmodel\src\tlo\scenario.py:438: InconsistentLoggedColumnsWarning: Inconsistent columns in logged values for tlo.scenario logger with key override_parameter compared to header generated from initial log entry: Columns in header not in logged values are {'new_value': 'bool', 'old_value': 'bool'} Columns in logged values not in header are {'new_value': 'int', 'old_value': 'int'} logger.info(

As crossed-checked by locally running another scenario in repo, I have got similar warnings. I understand these warnings are due to that in the log, first parameter's data type (such as bool type for 'spurious_symptoms ') and subsequent parameters' data types (such as list type of 'HR_expansion_by_officer_type') are different. Therefore, I think this is not an issue. (A copy of one tlo.scenario log is attached below.)

{"uuid": "eeda90f84c", "type": "header", "module": "tlo.scenario", "key": "override_parameter", "level": "INFO", "columns": {"module": "str", "name": "str", "new_value": "bool", "old_value": "bool"}, "description": null}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["SymptomManager", "spurious_symptoms", true, false]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "Service_Availability", ["*"], ["*"]]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "use_funded_or_actual_staffing", "actual", "funded_plus"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "mode_appt_constraints", 1, 1]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "cons_availability", "default", "default"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "beds_availability", "default", "default"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "equip_availability", "all", "default"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "use_funded_or_actual_staffing_postSwitch", "funded_plus", "funded_plus"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "year_use_funded_or_actual_staffing_switch", 2020, 2100]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "mode_appt_constraints_postSwitch", 2, 1]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "year_mode_switch", 2020, 2100]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "cons_availability_postSwitch", "all", "default"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "year_cons_availability_switch", 2020, 2100]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "yearly_HR_scaling_mode", "no_scaling", "no_scaling"]}
{"uuid": "eeda90f84c", "date": "2010-01-01T00:00:00", "values": ["HealthSystem", "HR_expansion_by_officer_type", [0, 0, 0, 0], [0, 0, 0, 0]]}

src/tlo/methods/healthsystem.py
BinglingICL: Logger created. Below is the comparison of scale up factors from healthsystem logger and from manual calculation, which seems generally consistent. The minor difference should be due to the former is calculated using daily minutes (where total annual minutes are evenly distributed to 365.25 days and scaled down by 'capabilities_coefficient') formatted within healthsystem module and minute-salary, whereas the latter is calculated using total staff counts and annual salary.

image

BinglingICL:
> Hi @marghe-molaro and @tbhallett, I have created a hr expansion function in the healthsystem module, considering the fractions of extra budget to be used by the four cadres (clinical, dcsa, nursing_and_midwifery, pharmacy), updated the scenarios and scenario script, and added some relevant files as needed. I would be very grateful if you could help check if the function and logics are looking fine. Any comment or advice would be greatly appreciated. Thanks.

Hi all, I have done some checks myself and it seems the coding will work. Would still be grateful if you could help check and confirm the scenarios and settings to be run. Many thanks. 

I just wonder how long it would take to run the scenario script for only first two scenarios (there are 17 scenarios in total to explore). If it is fine, I would like to submit such a job to Azure first to see the results and design analysis steps.

src/tlo/methods/healthsystem.py
BinglingICL: The difference of the above two dataframes may be mainly due to extra budget fractions saved and read as two-decimal floats, which are not exactly the same to the fractions. For example, for this scenario S_2 here, the four fractions add up to 0.99, not 1.0.

BinglingICL: Solution added: instead of save as csv and reload the extra budget fractions data, now directly import the fractions data from the script that calculate them, so that floats do not have lose decimals.

Will check again the differences as above once all scenarios are confirmed and job is re-run.

BinglingICL:
> The two-scenario run (2 draws * 5 runs) takes 12 hours. Job ID: scenario_run_for_hcw_expansion_analysis-2024-08-16T160132Z

Thanks Tim very much for your efforts in PR #1414, from which I have learned a lot to extract and plot results of DALYs and relevant!

The results of the small run above are presented below. It seems the branch is overall working, although many details and confirmation of scenarios still need discussions.
(0) Extra staff and extra budget needed (to add)

(1) Absolute numbers of services, DALYs and deaths.
Services_by_appointment_type_2020-2029
DALYs_2020-2029
DALYs_by_cause_2020-2029
Deaths_2020-2029

(2) Relative numbers of results, compared to no expansion scenario
Services_increased_by_appointment_type_2020-2029
DALYs_averted_2020-2029
DALYs_averted_by_cause_2020-2029
Deaths_averted_2020-2029

(3) Return on Investment and Cost Effectiveness
Cost_per_DALY_averted_2020-2029
DALYs_averted_per_extra_USD_dollar_invested_2020-2029

Looking forward to discuss all these results and remaining issues (as listed in "todo" in the files) with you all!

marghe-molaro:
@BinglingICL a minor point, but also remember that I expand capabilities from 2019 and not from 2020. I think ideally we should make this consistent.

BinglingICL:
> @BinglingICL a minor point, but also remember that I expand capabilities from 2019 and not from 2020. I think ideally we should make this consistent.
> 
> **Added**: The Berman +2022 paper from which the data on capabilities was taken states "Staff data from districts, central hospitals, and headquarters, and enrollment data from all health training institutions were collected between October 2017 and February 2018", so my understanding was that this data would be "true" for 2018, hence why we'd expand from 2019 and not 2020.

@marghe-molaro Thanks for reminding me of this. I think that choice of 2018 is very sensible. But from my latest communication with CHAI, they said the data source should be HSSP III not Berman 2022 and I have asked them the exact year of the "current staff" of the version we have received, for which they said it is 2019, just before pandemic. And that is the reason for my choice of 2019. For next run, I will set the current year to be 2018 and change year to be 2019 to see if our results match. For the final run and presentation, I might prefer to set change year to be 2020, considering the information I have received; but I am not really sure, because your work will soon published.

BinglingICL:
> > @BinglingICL a minor point, but also remember that I expand capabilities from 2019 and not from 2020. I think ideally we should make this consistent.
> > **Added**: The Berman +2022 paper from which the data on capabilities was taken states "Staff data from districts, central hospitals, and headquarters, and enrollment data from all health training institutions were collected between October 2017 and February 2018", so my understanding was that this data would be "true" for 2018, hence why we'd expand from 2019 and not 2020.
> 
> @marghe-molaro Thanks for reminding me of this. I think that choice of 2018 is very sensible. But from my latest communication with CHAI, they said the data source should be HSSP III not Berman 2022 and I have asked them the exact year of the "current staff" of the version we have received, for which they said it is 2019, just before pandemic. And that is the reason for my choice of 2019. For next run, I will set the current year to be 2018 and change year to be 2019 to see if our results match. For the final run and presentation, I might prefer to set change year to be 2020, considering the information I have received; but I am not really sure, because your work will soon published.

Hi @marghe-molaro, could you please also have a quick look in that scenario file and suggest the settings (such as popsize, number of runs) to be consistent with your runs? Thanks.

BinglingICL:
> > Hi @BinglingICL sorry for the delayed response.
> > Wrt the year of increase, I don't think this should make a huge difference esp. given the uncertainty around other assumptions, but it would be good to have consistent approaches. So was the Berman +22 data never used? If available, would it be possible to compare it to estimates from CHAI to see what the difference was over the two years? @tbhallett, what's your view on this?
> > Wrt the scenario files, the following parameters are missing from your scenarios to make it consistent with mine. "scale_to_effective_capabilities": True, "policy_name": "Naive", "tclose_overwrite": 1, "tclose_days_offset_overwrite": 7,
> 
> Thanks @marghe-molaro. Will update scenario run settings accordingly.
> 
> Just to explain that the data source of HSSP III (and the one that generated our HR capabilities) is provided by CHAI, just an updated version of that used by Berman paper (which is also provided by CHAI and has the same officer categories, appointment types, districts, etcs, but updated staff counts).

Regarding the difference of CHAI's two versions, I am not sure it actually reflects the growth of HR or differences between year 2018 and 2019, as I am not sure how they update their data...

marghe-molaro:
Let me clarify here for the record that, when including the "rescale to effective capabilities" option, the data on HRH is actually completely superfluous, because we take the HRH capabilities to be whatever was necessary to deliver *all* requested appts in the year before the switch to mode 2, and hence never rely on HRH data. 

I think a more accurate way of describing what we are doing, rather than saying we use HSSP-III data, would be to say "We assume that all requested appointments are always delivered until 2018 (or 2019). In that final year, given that we know the time requirement from each officer type for each appt delivered, we calculate the average effective capabilities that would have been required to deliver all those appts in that year. In subsequent years, we enforce constraints on the delivery of appts based on those estimated capabilities, which will be expanded at different rates.

BinglingICL:
> Let me clarify here for the record that, when including the "rescale to effective capabilities" option, the data on HRH is actually completely superfluous, because we take the HRH capabilities to be whatever was necessary to deliver _all_ requested appts in the year before the switch to mode 2, and hence never rely on HRH data.
> 
> I think a more accurate way of describing what we are doing, rather than saying we use HSSP-III data, would be to say "We assume that all requested appointments are always delivered until 2018 (or 2019). In that final year, given that we know the time requirement from each officer type for each appt delivered, we calculate the average effective capabilities that would have been required to deliver all those appts in that year. In subsequent years, we enforce constraints on the delivery of appts based on those estimated capabilities, which will be expanded at different rates.
> 
> **Added** This means that the rescaling year is completely arbitrary, and not based on when the data on HRH capabilities was collected

Thanks @marghe-molaro for this explanation. This is super helpful. So, may I understand that we are actually scaling up the such estimated capabilities of the previous year (let us say 2018), not the capabilities stored in our resource file?

src/tlo/methods/healthsystem.py
```diff
@@ -1016,6 +1065,11 @@ def _rescale_capabilities_to_capture_effective_capability(self):
         # Note: Currently relying on module variable rather than parameter for
         # scale_to_effective_capabilities, in order to facilitate testing. However
         # this may eventually come into conflict with the Switcher functions.
+
+        # In addition, for Class HRExpansionByOfficerType,
+        # for the purpose of keep cost not scaled, need to scale down minute salary when capabilities are scaled up
+
+        minute_salary = self.parameters['minute_salary']
```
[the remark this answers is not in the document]
marghe-molaro: I don't think so, but I might be wrong! You can quickly check this by printing the self.parameters['minute_salary'] in the HRHExpansionByOfficerType function with and without the scaling_to_effective_capabilities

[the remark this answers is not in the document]
BinglingICL: I had checked the update last night, but I might have missed something, as it was midnight. Happy to check again and let you know. :)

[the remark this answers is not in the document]
BinglingICL: Hi @marghe-molaro, I have debugged the coding and believe that minute_salary and self.parameter['minute_salary'] are updated at the same time, as they are the same object and the latter is not a copy of the former.

BinglingICL:
Dear all, the job of all 33 scenarios has been submitted. Hope it will run well and we could see the results next week.

By the way, Margherita @marghe-molaro, happy to let you know that the DALYs of year 2029 from my last run is around 9 million, looks consistent with yours!

BinglingICL:
Dear all, I would like to share some results for all scenarios that have been designed. Looking forward to meeting you next week and would very much appreciate your advice and comments.
image
image
image
image
image
image
image

RachelMurray-Watson:
> Dear all, I would like to share some results for all scenarios that have been designed. Looking forward to meeting you next week and would very much appreciate your advice and comments. image image image image image image image image

Just for my own understanding, why is that some of the middle scenarios (e.g. CD_equal, CNO_equal) in the "Services increased by treatment type" graph seem to deliver fewer services than the "no expansion" scenario?

BinglingICL:
> Just for my own understanding, why is that some of the middle scenarios (e.g. CD_equal, CNO_equal) in the "Services increased by treatment type" graph seem to deliver fewer services than the "no expansion" scenario?

Thanks for this question Rachel. Those scenarios do not have significantly less services than no expansion scenario. I have attached the percentages of service increased below. In scenarios where there is not much services increased but with more DALYs averted (as compared to no expansion scenario), there should be some trade-off between some treatments/appointments, leading to trade-off of dalys of some causes/injuries.

The contrary results (of services and dalys) of middle groups of scenarios and the right group, might also have such reasons behind, which need to be examined. The fundamental HRH allocation difference is that middle group have Clinical cadre expanded but no Pharmacy, whereas the right group have Pharmacy cadre expanded but not Clinical.

RachelMurray-Watson:
> Thanks for this question Rachel. Those scenarios do not have significantly less services than no expansion scenario. I have attached the percentages of service increased below. In scenarios where there is not much services increased but with more DALYs averted (as compared to no expansion scenario), there should be some trade-off between some treatments/appointments, leading to trade-off of dalys of some causes/injuries.
> 
> <style> </style>
> ## Percentage service increased by treatment type against no expansion scenario
> scenario | lower | mean | upper CNO_equal | 0.50% | 0.55% | 0.64% CD_equal | -0.03% | 0.13% | 0.26% Clinical | -0.54% | -0.26% | -0.03% CO_equal | -0.39% | -0.16% | -0.04%
> 
> The contrary results (of services and dalys) of middle groups of scenarios and the right group, might also have such reasons behind, which need to be examined. The fundamental HRH allocation difference is that middle group have Clinical cadre expanded but no Pharmacy, whereas the right group have Pharmacy cadre expanded but not Clinical.

Ah okay, so it's very small - do you think it's just stochasticty? I understand what you're saying in terms of DALYs etc., but I still would have thought that expansion of healthcare cadres would at least maintain the number of services provided, even if it doesn't actually increase it. (Irrespective of the effect on DALYs.)

BinglingICL:
> > Thanks for this question Rachel. Those scenarios do not have significantly less services than no expansion scenario. I have attached the percentages of service increased below. In scenarios where there is not much services increased but with more DALYs averted (as compared to no expansion scenario), there should be some trade-off between some treatments/appointments, leading to trade-off of dalys of some causes/injuries.
> > <style> </style>
> > ## Percentage service increased by treatment type against no expansion scenario
> > scenario | lower | mean | upper CNO_equal | 0.50% | 0.55% | 0.64% CD_equal | -0.03% | 0.13% | 0.26% Clinical | -0.54% | -0.26% | -0.03% CO_equal | -0.39% | -0.16% | -0.04%
> > The contrary results (of services and dalys) of middle groups of scenarios and the right group, might also have such reasons behind, which need to be examined. The fundamental HRH allocation difference is that middle group have Clinical cadre expanded but no Pharmacy, whereas the right group have Pharmacy cadre expanded but not Clinical.
> 
> Ah okay, so it's very small - do you think it's just stochasticty? I understand what you're saying in terms of DALYs etc., but I still would have thought that expansion of healthcare cadres would at least maintain the number of services provided, even if it doesn't actually increase it. (Irrespective of the effect on DALYs.)

It might be for some scenarios, among the middle groups and right groups, which have <= 6% DALYs averted, and the last three scenarios, which have 0% DALYs averted. But I am not very sure, because looking at the results of the middle and right groups, the patterns are very consistent. If it is due to stochastic variation, some in middle group should have more services increased and/or some in right group should have less services increased.

RachelMurray-Watson:
> > Ah okay, so it's very small - do you think it's just stochasticty? I understand what you're saying in terms of DALYs etc., but I still would have thought that expansion of healthcare cadres would at least maintain the number of services provided, even if it doesn't actually increase it. (Irrespective of the effect on DALYs.)
> 
> It might be for some scenarios, among the middle groups and right groups, which have <= 6% DALYs averted, and the last three scenarios, which have 0% DALYs averted. But I am not very sure, because looking at the results of the middle and right groups, the patterns are very consistent. If it is due to stochastic variation, some in middle group should have more services increased and/or some in right group should have less services increased.

True, there is a pattern, so is there another reason why more healthcare workers would result in fewer services?

BinglingICL:
> > > Ah okay, so it's very small - do you think it's just stochasticty? I understand what you're saying in terms of DALYs etc., but I still would have thought that expansion of healthcare cadres would at least maintain the number of services provided, even if it doesn't actually increase it. (Irrespective of the effect on DALYs.)
> > 
> > 
> > It might be for some scenarios, among the middle groups and right groups, which have <= 6% DALYs averted, and the last three scenarios, which have 0% DALYs averted. But I am not very sure, because looking at the results of the middle and right groups, the patterns are very consistent. If it is due to stochastic variation, some in middle group should have more services increased and/or some in right group should have less services increased.
> 
> True, there is a pattern, so is there another reason why more healthcare workers would result in fewer services?

I think the question is why more Clinical staff and less Pharmacy staff (as in Middle group scenarios) leading to less services and treatments (in total at least) but more DALYs averted, compared to less Clinical staff and more Pharmacy staff (as in Right group scenarios).

In an absolutely terms, there should no scenario that have more HRH but fewer services, which would indicate something not right in the simulation, except stochastic variations. Did you spot any such scenario, Rachel? I should take a closer look into that.

BinglingICL:
> I think the question is why more Clinical staff and less Pharmacy staff (as in Middle group scenarios) leading to less services and treatments (in total at least) but more DALYs averted, compared to less Clinical staff and more Pharmacy staff (as in Right group scenarios).

Oh, dear all, I think the reason might be relevant to the "Pharm Dispensing" appointment (and the "Contraception" treatment), which is a dummy one; with more clinical staff, some services could be delivered rather than diverted to Pharm Dispensing. But I am not sure. Need to dig into this.

BinglingICL:
> > > > It might be for some scenarios, among the middle groups and right groups, which have <= 6% DALYs averted, and the last three scenarios, which have 0% DALYs averted. But I am not very sure, because looking at the results of the middle and right groups, the patterns are very consistent. If it is due to stochastic variation, some in middle group should have more services increased and/or some in right group should have less services increased.
> > > 
> > > 
> > > True, there is a pattern, so is there another reason why more healthcare workers would result in fewer services?
> > 
> > 
> > I think the question is why more Clinical staff and less Pharmacy staff (as in Middle group scenarios) leading to less services and treatments (in total at least) but more DALYs averted, compared to less Clinical staff and more Pharmacy staff (as in Right group scenarios).
> > In an absolutely terms, there should no scenario that have more HRH but fewer services, which would indicate something not right in the simulation, except stochastic variations. Did you spot any such scenario, Rachel? I should take a closer look into that.
> 
> Do these ones I've squared in black here here not have more HRH (because they're expansion scenarios) but fewer services (negative on the axis)?
> 
> image

Thanks Rachel. Those scenarios have some more services like Contraception but less EPI and Firstattence, and the total services increased are close to 0 or positive. But you are right, we need to examine why some services are reduced in these expansion scenarios compared to no expansion scenario.

src/tlo/methods/healthsystem.py
```diff
@@ -2976,11 +3157,75 @@ def apply(self, population):
             matches = re.match(pattern, officer)
             # Extract ID and officer type from
             facility_id = int(matches.group(1))
-            district = self.module._facility_by_facility_id[facility_id].district
+            district = self.module._facility_by_facility_id[facility_id].district  # todo: check if district callable
```
BinglingICL, line 3160: Hi Margheria @marghe-molaro, just raise a small todo task here: to check if _facility_by_facility_id[facility_id] has the attribute district and this line can run successfully.

If not, then a small fix might be:  district = self.module._facility_by_facility_id[facility_id].name.split('_')[-1], as the facility name info has district indicated.

BinglingICL:
Hi Tim @tbhallett, the branch has now merged in the latest master and all errors fixed. Happy for you to review it and see if any change needed before we merge it to master.

src/tlo/methods/healthsystem.py
```diff
@@ -2833,17 +2960,45 @@ def write_to_log_and_reset_counters(self):
...
+            # year.
+            logger_summary.info(
+                key="Capacity_By_OfficerType_And_District",
+                description="The fraction of healthcare worker time that is used each day, averaged over this "
+                            "calendar year, for each officer type in each district.",
+                data=flatten_multi_index_series_into_dict_for_logging(
+                    self.frac_time_used_by_officer_district(clinic)),
+            )
+
+            # Log mean of 'fraction time used by officer type and facility level and district' from daily entries
```
marghe-molaro: Not sure if the logging should happen here, ideally it would be logged at the same time as all other counters. This is to avoid a issues related to the frequency of rescaling being non-synchronous with the desired frequency of the logging

src/tlo/analysis/utils.py
```diff
@@ -570,6 +570,100 @@ def compute_summary_statistics(
         return summary
 
 
+def compute_summary_statistics_by_cause(
```
tbhallett: this might be useful for others, but the code and logic is a bit duplicative with the basic `compute_summary_statistics`. Here's a proposal for a reworking. Does it still do what you want it to do?

```suggestion```

File: src/tlo/methods/healthsystem.py [hunk 6]
```diff
@@ -351,11 +351,40 @@ class HealthSystem(Module):
         "tclose_overwrite": Parameter(
             Types.INT, "Decide whether to overwrite tclose variables assigned by disease modules"
         ),
-        "tclose_days_offset_overwrite": Parameter(
+                "tclose_days_offset_overwrite": Parameter(
             Types.INT,
             "Offset in days from topen at which tclose will be set by the healthsystem for all HSIs"
             "if tclose_overwrite is set to True.",
         ),
+        "HR_expansion_by_officer_type": Parameter(
```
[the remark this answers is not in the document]
BinglingICL: Hi Tim, fully agree that the other way is more re-usable . This branch was designed for the purpose of the paper in the first place. If we do not need the full branch in the master but only the re-usable HRH expansion by cadre and year function, I can write a HRH expansion class like others already in the module, where the input parameter is a data frame of HRH scale-up factor by cadre and year. The 'prepare_minute_salary_and_extra_budget_frac_data.py' file will generate the input values of 134 scenarios, which are not necessarily to be stored in .csv files in the resource folder as that is quite a lot. In the health system module, we then will not need to read in salaries and relevant. Please let me know if this is the prefered the way and I would be happy to do that,

BinglingICL:
This branch will be put in personal TLO.

marghe-molaro:
Hi @BinglingICL, I think similarly the decision here was that you don't want to merge this branch into master, so I think we can close this PR too?

BinglingICL:
> Hi @BinglingICL, I think similarly the decision here was that you don't want to merge this branch into master, so I think we can close this PR too?

Yes @marghe-molaro. These branches can be closed. Just wonder if they can be stores locally and will not be overwritten by the new version of TLO, so that we can look for some existing coding in case needed in the future. Thanks.

PR #1417, opened 2024-07-09
Lookup Consumables in RTI

thewati:
Function to lookup consumables in RTI in a more efficient way. Updates consumable dictionaries for fracture, laceration, and burn treatment from the same lookup function


tbhallett:
Thanks very much for this @thewati 
I'm noticing a conflict. Please could you resolve it?

src/tlo/methods/rti.py
```diff
@@ -41,7 +41,8 @@ def __init__(self, name=None, resourcefilepath=None):
         super().__init__(name)
         self.resourcefilepath = resourcefilepath
         self.ASSIGN_INJURIES_AND_DALY_CHANGES = None
-        self.item_codes_for_consumables_required = dict()
+        # self.item_codes_for_consumables_required = dict()
```
tbhallett, line 44:
```suggestion
```

src/tlo/methods/rti.py
```diff
@@ -3961,14 +4048,11 @@ def apply(self, person_id, squeeze_factor):
...
-            get_item_code('Bandage, crepe 7.5cm x 1.4m long , when stretched'): 200,
-        }
-        is_cons_available = self.get_consumables(
-            self.module.item_codes_for_consumables_required['fracture_treatment']
-        )
+
+        # If they have a fracture that needs a cast, ask for plaster of paris, update POP count
+        self.module.cons_item_codes['fracture_treatment'].update({
+            get_item_code('Plaster of Paris (POP) 10cm x 7.5cm slab_12_CMST'): fracturecastcounts})
+        is_cons_available = self.get_consumables(self.module.cons_item_codes['fracture_treatment'])
```
tbhallett, line 4055: I feel that we could have looked up this where we do everything else and that this would complete the pattern.

src/tlo/methods/rti.py
```diff
@@ -3961,14 +4048,11 @@ def apply(self, person_id, squeeze_factor):
...
-        self.module.item_codes_for_consumables_required['fracture_treatment'] = {
-            get_item_code('Plaster of Paris (POP) 10cm x 7.5cm slab_12_CMST'): fracturecastcounts,
-            get_item_code('Bandage, crepe 7.5cm x 1.4m long , when stretched'): 200,
-        }
-        is_cons_available = self.get_consumables(
-            self.module.item_codes_for_consumables_required['fracture_treatment']
-        )
+
+        # If they have a fracture that needs a cast, ask for plaster of paris, update POP count
+        self.module.cons_item_codes['fracture_treatment'].update({
```
tbhallett, line 4053: This updated the module's store of the consumables, right? So would have consequences for all subsequent individuals. I think the interned behaviour originally was that these are additional
Items needed only for this person.

src/tlo/methods/rti.py
```diff
@@ -4094,25 +4178,19 @@ def apply(self, person_id, squeeze_factor):
...
-            }
             # If wound is "grossly contaminated" administer Metronidazole
             # todo: parameterise the probability of wound contamination
             p = self.module.parameters
             prob_open_fracture_contaminated = p['prob_open_fracture_contaminated']
             rand_for_contamination = self.module.rng.random_sample(size=1)
             # NB: Dose used below from BNF is for surgical prophylaxsis
             if rand_for_contamination < prob_open_fracture_contaminated:
-                self.module.item_codes_for_consumables_required['open_fracture_treatment'].update(
+                self.module.cons_item_codes['open_fracture_treatment'].update(
```
tbhallett, line 4188: Same comment as above about us probably not wanting to edit the module's store of the consumables

src/tlo/methods/rti.py
```diff
@@ -4316,11 +4390,11 @@ def apply(self, person_id, squeeze_factor):
         assert df.loc[person_id, 'rt_med_int'], 'this person has not been treated'
         if burncounts > 0:
             # Request materials for burn treatment
-            self.module.item_codes_for_consumables_required['burn_treatment'] = {
+            self.module.cons_item_codes['burn_treatment'].update({
```
tbhallett, line 4393: Here too

tbhallett (changes requested):
Fantastic job with this @thewati 

It's a very nice improvement.  

See one comment where I'm questions whether we really are reproducing the original logic, and also where we can factor-out those last few calls to "additional" consumables items in the same manner you've done for everything else. 

Let me know what you think / if you think I've misread it! 

Thanks again for this.

tbhallett (approved):

PR #1418, opened 2024-07-09
Calculating probability of premature death 

### LINKED ISSUE #1374: Summarise 'Probability of Premature Death' (PPD) as additional useful health metric
Prematature deaths are defined as deaths occurring before the age of 70 (See [Norheim et al (2015)](https://pubmed.ncbi.nlm.nih.gov/25242039/))

This could be in the logger of either the HealthBurden or Demography Module, or as an extension to the life-expectancy calculation.

RachelMurray-Watson:
Fixes #1374 

Initial code for calculating the probability of premature death (defined as before 70 years old) and a test for that code. Based off of life_expectancy.py, using the cumulative probabilities of death before the age of 70 years. Calculated separately for M and F.  Seems to roughly match figures based on https://apps.who.int/gho/data/view.searo.60980?lang=en.


src/tlo/analysis/probability_premature_death.py
```diff
@@ -0,0 +1,107 @@
...
+from pathlib import Path
+from typing import Dict, Tuple
+
+import pandas as pd
+
+from tlo.analysis.life_expectancy import _aggregate_person_years_by_age, _num_deaths_by_age_group
+from tlo.analysis.utils import get_scenario_info, summarize
+
+
+def _calculate_probability_of_dying_before_70(
```
tbhallett, line 18: I would rename to `_calculate_probability_of_premature_death` and throughout 'soft-code' the "magic number of 70".

I would declare at the top of the script

`AGE_BEFORE_WHICH_DEATH_IS_DEFINED_AS_PREMATURE = 70`


and provide the reference to the paper that is cited in the issue.

That way we're super-clear about the origin of the magic number. And, if, we change our mind about the definition, it's a simple change.

tests/test_probability_premature_death.py
```diff
@@ -0,0 +1,36 @@
...
+import datetime
+import os
+from pathlib import Path
+
+import pandas as pd
+
+from tlo.analysis.probability_premature_death import get_probability_of_dying_before_70
+
+
+def test_get_probability_premature_death():
```
tbhallett, line 11: given that we're running this test on a known data set, would it make sense to check that the actual numerical answer obtained is exactly equal to what we want?
(i realise the test you modelled this on, didn't do this, but arguably it should have!)

src/tlo/analysis/probability_premature_death.py
```diff
@@ -0,0 +1,107 @@
...
+    _person_years_at_risk: pd.Series,
+    _number_of_deaths_in_interval: pd.Series
+) -> Dict[str, float]:
+    """
+    For a single run, estimate the probability of dying before age 70 for males and females.
+    Returns: Dict (keys by "M" and "F" for the sex, values the estimated probability of dying before age 70).
+    """
+    probability_of_dying_before_70 = dict()
+
+    age_group_labels = _person_years_at_risk.index.get_level_values('age_group').unique()
```
tbhallett, line 28: as this logic has so much in common with the life-expectancy calculation, I think it would make sense to factor it out, in order that the same calculation can serve both purposes (life-expectnacy and probability of premature death).

I would also recommend putting these functions into the same file as the life-expectnacy calculations

tbhallett (changes requested):
Very nice!

src/tlo/analysis/probability_premature_death.py
RachelMurray-Watson: So do you think this file should go entirely, and just include it all in the life expectancy file?

tbhallett: yes, I think that would be fine. 

Main point is about the refactoring to avoid repeating the logic about computing death rates in age-groups and building the cumulative probability of death. 

(Location of the function in which file is secondary, but I think it being in one file would be fine).

src/tlo/analysis/life_expectancy.py
```diff
@@ -98,6 +100,47 @@ def _aggregate_person_years_by_age(results_folder, target_period) -> pd.DataFram
...
+            number_alive_at_start_of_interval[i] = (1 - probability_of_dying_in_interval[i - 1]) * \
+                                                   number_alive_at_start_of_interval[i - 1]
+
+        mortality_metrics[sex] = {
+            'death_rate_in_interval': death_rate_in_interval,
+            'probability_of_dying_in_interval': probability_of_dying_in_interval,
+            'number_alive_at_start_of_interval': number_alive_at_start_of_interval
+        }
+
+    return mortality_metrics
```
tbhallett, line 143: this newly-added function isn't used!?

tests/test_probability_premature_death.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+from pathlib import Path
+
+from tlo.analysis.life_expectancy import get_probability_of_premature_death
+from tlo.analysis.utils import extract_results
+
+target_period=(datetime.date(2010, 1, 1), datetime.date(2080, 12, 31))
+AGE_BEFORE_WHICH_DEATH_IS_DEFINED_AS_PREMATURE = 70
+
+def test_probability_premature_death(tmpdir):
```
tbhallett, line 11: I would add to the `test_life_expectancy.py` file as that file already exists and it seems reasonable for to 'house' test for the two function in the `life_expectnancy.py` file

tests/test_probability_premature_death.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    total_male = extract_results(
+        results_folder_dummy_results,
+        module="tlo.methods.demography",
+        key="population",
+        column='male',
+        do_scaling=False
+    )
+
+    probability_premature_death_sim_F = len(deaths[deaths[0] == 'F'])/len(total_female)
```
tbhallett, line 39: it's not clear that these desths will be 'preamture' and occur within the taget_period You're extracting all deaths. I assume this is because all the deaths are premature in the pickle file and all are within the target_period? 
But, I would make a clauses so that it's explicit that this is what we're doing.

tests/test_probability_premature_death.py
```diff
@@ -0,0 +1,55 @@
+import datetime
+import os
+from pathlib import Path
+
+from tlo.analysis.life_expectancy import get_probability_of_premature_death
+from tlo.analysis.utils import extract_results
+
+target_period=(datetime.date(2010, 1, 1), datetime.date(2080, 12, 31))
```
tbhallett, line 8: this can go inside the function

tests/test_probability_premature_death.py
```diff
@@ -0,0 +1,55 @@
...
+AGE_BEFORE_WHICH_DEATH_IS_DEFINED_AS_PREMATURE = 70
```
tbhallett, line 9: I've suggested that this becomes an argument to the function `get_probability_of_premature_death` that defaults to 70, and that would obviate the need to define this here.

tbhallett (changes requested):

src/tlo/analysis/life_expectancy.py
```diff
@@ -248,3 +291,93 @@ def get_life_expectancy_estimates(
...
+            person_years_by_sex / interval_width / fraction_of_last_age_survived)
+        probability_of_dying_in_interval = pd.Series(index=number_of_deaths_by_sex.index, dtype=float)
+        probability_of_dying_in_interval[condition] = 1
+        probability_of_dying_in_interval[~condition] = interval_width * death_rate_in_interval / (
+            1 + interval_width * (1 - fraction_of_last_age_survived) * death_rate_in_interval)
+        probability_of_dying_in_interval.at['90'] = 1
+
+        # Calculate cumulative probability of dying before the defined premature age
+        cumulative_probability_of_dying = 0
+        number_alive_at_start_of_interval = 1.0
```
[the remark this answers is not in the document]
RachelMurray-Watson: Yes; have corrected now, thanks!

src/tlo/analysis/life_expectancy.py
RachelMurray-Watson: Sorry; have removed now!

tests/test_probability_premature_death.py
RachelMurray-Watson: Grand! Have done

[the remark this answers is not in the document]
RachelMurray-Watson: Done!

src/tlo/analysis/probability_premature_death.py
RachelMurray-Watson: Done!

src/tlo/analysis/probability_premature_death.py
RachelMurray-Watson: Done! REfactored all logic around probability of dying

tdm32 (approved):

tests/test_life_expectancy.py
```diff
@@ -33,3 +39,53 @@ def test_get_life_expectancy():
...
+        results_folder=results_folder_dummy_results,
+        target_period=target_period,
+        summary=True,
+    )
+
+    # Confirm both methods gives the same answer
+    assert np.isclose(
+        probability_premature_death_summary.loc["F"].loc[(0, 'mean')],
+        prob_premature_death['F'],
+        atol=0.01
```
tbhallett, line 85: @RachelMurray-Watson  I've proposed updating this test to be more direct check that the utility function gives the answer we expect. I was surprised that the tolerance allowance on this had to be 0.01 and not something smaller. Do you know why the two don't come closer?

RachelMurray-Watson: I'm not sure - looking at the life expectancy estimates, too, can see that there's sometimes +- 3% difference. It may be here that there's a very small population size, so the estimates from probabilities leads to inaccurate results?

tbhallett: I don't think there's any stochasticity in the procedure generating these estimates though....? To my mind, the two methods in the test (manual and using the new function) should give same result as each is operating on exactly the same data?

tbhallett: I was thinking that it might be something to do with the PY being used in the calc in the function, versus the just the initial cohort size in the manual calc.

eg. is the probability of premature death rightly calculated as proportion of deaths that occur at an age less than aged 70 (per manual calculation); or, is expected cumulative risk of death by age 70.

I think these two quantities are subtly different....!?

RachelMurray-Watson: Ah sorry, I see what you mean when applied to the same synthetic cohort...  I guess what I meant with the small cohort was that some assumptions e.g. about fractions of age groups survived may not hold as well, but I think that that does lead in to your point about PY vs the "manual" calculation as you said. 

I think that I understand the probability of premature death as a metric as the cumulative risk (like a life-table method), as you said. So do you think for a "correct" test, I should instead calculate the PY, etc. at risk etc for this cohort? 

I guess then I'm wondering what this test does tell us then; that this is an imperfect estimate?

tbhallett: Ok, that's the answer then. That "small assumptions" are made about the PY spent in age-group in our calculation that mean that our manual or 'utility fn' approaches legitimately can be a little bit apart.

If that's the case, then I think we're done! I'll add in a little comment on the back of the assert line to say that.

tbhallett (approved):

tests/test_life_expectancy.py
RachelMurray-Watson: Amazing! Thanks so much like

tbhallett:
No idea why `test_determinism.py` is failing on this branch now. Could this be something to do with the runners @matt-graham ???

matt-graham:
> No idea why `test_determinism.py` is failing on this branch now. Could this be something to do with the runners @matt-graham ???
> 
> If it's only that, we can merge this now.

It's failing on `master` post merging #1470  (more details in #1472). I was hoping #1473 would fix but it doesn't appear to and currently trying to investigate alternatives (annoyingly this is only happening on Actions runner so bit difficult to figure out what is going on!). As unrelated to changes here I think we can safely merge this in if that's the only test failing.

PR #1422, opened 2024-07-12
Adding script for generating table of parameters

### LINKED ISSUE #1419: Parameter Table
For our overview paper (and also generally, I think), we need to have in one place an assembly of all the parameters that are being used in the model*. This would be a "table" (but see below) of the form:

| Module | Parameter Name | Parameter Description | Value |
|--------|--------|--------|--------|

... where 'Parameter Description' is the description that is given when the Parameter is declared by each module.

Of course, the 'Value' for each parameter can be a something simple to put into a table (`Types.BOOL`, `Types.REAL`, and even `Types.LIST` and `Types.DICT`), but several of these parameters are of type `pandas.DataFrame` and `pandas.Series` and I think these would need to be put into another little table.

For the immediate use, ideally this would table would be rendered as one single pdf file, wherein sub-tables are linked-to by within-document hyperlinks. If that's hard, then any other format single file format  (e.g. html, excel etc) would work.

I was thinking this might be quite a similar job to what @matt-graham has done already to render the [Resource Files](https://www.tlomodel.org/resources/index.html) on the tlomodel.org website. 

(As an aside - and just to link the issues for information -- this seems to me _slightly_ related to https://github.com/UCL/TLOmodel/issues/1337 --- as this would avoid us having parameters of data frames which some associated detritus (columns not used, arbitrary annotations etc.)).

*The parameter values actually being used in a specific run of the model -- specifically, those that exist at the start of the simulation in [this](https://github.com/UCL/TLOmodel/blob/master/src/scripts/calibration_analyses/scenarios/long_run_all_diseases.py) specification of the model.

matt-graham:
Work in progress but should resolve #1419 

Script can be run from root of respository with

```
python docs/tlo_parameters.py resources parameters.md
```

with this generating a Markdown output `parameters.md` which can be converted to HTML with

```
pandoc -s parameters.md -o parameters.html --css pandoc.css --self-contained
```

using CSS from https://gist.github.com/killercup/5917178

I'll hopefully be able to get this to also output reStructuredText to allow running as part of the documentation build and including on the website.


src/tlo/methods/healthsystem.py
```diff
@@ -165,7 +165,7 @@ class HealthSystem(Module):
         'use_funded_or_actual_staffing': Parameter(
             Types.STRING, "If `actual`, then use the numbers and distribution of staff estimated to be available"
                           " currently; If `funded`, then use the numbers and distribution of staff that are "
-                          "potentially available. If 'funded_plus`, then use a dataset in which the allocation of "
+                          "potentially available. If `funded_plus`, then use a dataset in which the allocation of "
```
matt-graham, line 168: Change here is to prevent Markdown rendering in parameter table getting disrupted by non-matching quotation marks

tbhallett:
I've made an edit in [this](8b315ded68f4bd4bcdaa8f1c6a152f1cfb696886) commit (floating loose in the repo) to make the pagination more manageable (by excluding two parameters, which are really just datasets). Also, excluded the `Wasting` module that has no parameters.

matt-graham:
Thanks @tbhallett - I've applied the patch you suggested and also now updated to allow outputting a reStructuredText file as well as Markdown to allow inclusion in our docs, and updated the docs build to automatically generate this page.

tbhallett:
Thanks @matt-graham  -- that's brilliant and very timely. It will be great to have this on the website.

matt-graham:
@tbhallett just to check are you happy for this to be merged in? Know that you already approved but as I made a few subsequent changes after that wanted to make sure!

tbhallett:
Yes please do merge in. Thanks again. And, if you can, please could you also regenerate the website so that it appears there?

PR #1423, opened 2024-07-16
Use `IndividualDetails` context for `SymptomManager.has_what`

willGraham01:
We are currently spending a lot of time within the `SymptomManager.has_what` method (called inside the `do_at_generic_first_appt` methods) - approx 8% of simulation time according to the [most recent profiling](http://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_1009_a83f624e56817fe0c37d7b0e1db55cf593ccb63e.html) result.

SnakeViz places the blame for this on calls to pandas methods within `SymptomManager`, because it does not make use of the `IndividualDetails` context that is setup during the `generic_first_appt` method.

master_snakeviz

This PR introduces the following changes:

- `SymptomManager.has_what` can now be passed an `IndividualProperties` instance, which it will use instead of looking up values directly from the population DataFrame.
- This same method now only takes keyword arguments, and must be supplied either the `person_id` or `individual_details` inputs. The `individual_details` argument takes priority over the `person_id` method if both are supplied (since it is at worst the same speed as looking up from the DF each time).
- All calls to `SymptomManager.has_what` in the codebase now respect the keyword arguments the function takes.
- `test_has_what` has been updated to check that the new method of retrieving symptoms via the `individual_details` keyword performs as expected.

With these changes, the time spent within `has_what` has almost halved:

exp_snakeviz

with us now spending 174s inside `has_what` and all subfunctions, rather than `331s` that we were spending before. Experiments were run using the `scale_run` simulation over 2 years and using a relatively small initial population of 10,000, so have triggered a profiling run here to get some more accurate figures.

### DIFF
--- a/src/tlo/methods/symptommanager.py
+++ b/src/tlo/methods/symptommanager.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Optional, Sequence, Union
+from typing import TYPE_CHECKING, List, Optional, Sequence, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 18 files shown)


willGraham01:
Profiled run succeeded but we hit the expired token problem. Rerunning now

matt-graham:
Just to say in case it's helpful for checking things in the interim, the artifact from the previous profiling run is still available to download even though it wasn't uploaded to profiling website https://github.com/UCL/TLOmodel/actions/runs/9953255205/artifacts/1707268804

willGraham01:
> Just to say in case it's helpful for checking things in the interim, the artifact from the previous profiling run is still available to download even though it wasn't uploaded to profiling website https://github.com/UCL/TLOmodel/actions/runs/9953255205/artifacts/1707268804

Thanks - it looks like the `SymptomManager` footprint inside the generic first appts method has indeed shrunk considerably (at least according to the artifact):

image

src/tlo/methods/symptommanager.py
```diff
@@ -460,33 +465,81 @@ def who_not_have(self, symptom_string: str) -> pd.Index:
...
+        assert (
+            disease_module.name in ([self.name] + self.recognised_module_names)
+            if disease_module is not None
+            else True
+        ), "Disease Module Name is not recognised"
+
+        if individual_details is not None:
+            # We are working in an IndividualDetails context, avoid lookups to the
+            # population DataFrame as we have this context stored already.
+            assert individual_details["is_alive"], "The person is not alive"
```
tamuri, line 499: I know this replicates the existing behaviour but this check probably belongs somewhere else. It's conceivable to want to know the symptoms of deceased individual. @tbhallett ?

tamuri (approved):
Looks good - with nice improvement in performance!

src/tlo/methods/symptommanager.py
tbhallett: I'm not sure if we would need to know the symptoms of someone who has already died. We may wish to log symptoms of someone at the point of death (like we log other characteristics) but that's different.
I think throughout the code we assume that the information from deceased individuals is deleted (even though its not)

PR #1424, opened 2024-07-17
Update run-profiling.yaml with note about new token

willGraham01:
Our token for pushing to the profiling repository expired 10/7/2024 which is the cause of failure for the last few profiling runs. 

I've regenerated the token to last another year, with the same minimal permissions. Adding a note to the workflow so if this happens again we can find the cause.


.github/workflows/run-profiling.yaml
```diff
@@ -164,6 +164,7 @@ jobs:
       ## Token can be (re)generated by a member of the UCL organisation, 
       ## the current member is the rc-softdev-admin.
       ## [10-07-2023] The current token will expire 10-07-2024
+      ## [17-07-2024] New token generated, will expire 10-07-2025
```
matt-graham, line 167: I'd personally say better to just replace the current comment than add a new line, as only the latest date is relevant (and the previous comment will still be in commit history if we wanted to check previous expiration dates for some reason).

matt-graham (approved):
Minor comment about updating existing commment rather than adding but otherwise good. Thanks for dealing with renewing the token!

PR #1425, opened 2024-07-17
Convert excel ResourceFiles to csv - Add a method to read converted files

mnjowe:
This PR aims at adding a method that will help reading the newly converted csv files in a similar manner as their excel equivalent were read. It also addresses situations where multiple sheets are required i.e `pd.read_excel(resourcefile_path, sheet_name=[sheet1, sheet2, ...., sheetN])`.


src/tlo/util.py
```diff
@@ -436,3 +437,30 @@ def get_person_id_to_inherit_from(child_id, mother_id, population_dataframe, rng
...
+    elif file_name is not None:
+        for f_name in file_name:
+            df = pd.read_csv(f'{resource_filepath}/{f_name}.csv')
+            all_data[f_name] = df.drop(df.filter(like='Unnamed'), axis=1)
+
+    else:
+        for f_name in resource_filepath.rglob("*.csv"):
+            file_base_name = str(f_name).split('/')[-1].split('.')[0]  # Extract base name without extension
+            df = pd.read_csv(f_name)
+            all_data[file_base_name] = df.drop(df.filter(like='Unnamed'), axis=1)
```
tbhallett, line 464: What is the df.filter thing doing here? I'm sure it's going a good job but I think it would be worth commenting on it and also noting in the docstring if this introduces any behaviour the user should be aware of? Eg. Sheets with certain names will get dropped silently?

src/tlo/util.py
```diff
@@ -436,3 +437,30 @@ def get_person_id_to_inherit_from(child_id, mother_id, population_dataframe, rng
...
     elif mother_id >= 0:
         return mother_id
+
+
+def read_csv_files(resource_filepath: Path, file_name: list[str] = None) -> DataFrame | dict[str, DataFrame]:
+    """
+    A function to read csv files in a similar way pandas read Excel files(pd.read_excel).
+
+    :param resource_filepath:  path to resource file folder
+    :param file_name: preferred csv file name(s). This is the same as sheet names in Excel file
```
tbhallett, line 447: Note that if None then all the files there will be loaded

src/tlo/util.py
```diff
@@ -436,3 +437,30 @@ def get_person_id_to_inherit_from(child_id, mother_id, population_dataframe, rng
         return abs(mother_id)
     elif mother_id >= 0:
         return mother_id
+
+
+def read_csv_files(resource_filepath: Path, file_name: list[str] = None) -> DataFrame | dict[str, DataFrame]:
```
tbhallett, line 442: I would recommend not using the phrase "resource file path" for this as we use that for the destination of the overall resources folder. Instead how about the kwargs `folder: Path, files: list[str]`

tbhallett (commented):
Thanks @mnjowe 

This looks firmly on the right tracks to me.

src/tlo/util.py
matt-graham, line 464: In addition to @tbhallett's comment regarding documenting this behaviour, ideally we would avoid the reperition of the `df.drop(df.filter(like='Unnamed'), axis=1)` snippet - either by refactoring the conditional structure here or factoring out into a helper function, so that its easier to update and keep consistent.

matt-graham (commented):
Thanks @mnjowe - agree with @tbhallett this is looking good, have added some additional comments below.

mnjowe:
Thanks @tbhallett and @matt-graham for your helpful comments above. I have pushed a couple of commits that seeks to address them all. Let's keep the discussion going. Thanks

tests/test_utils.py
```diff
@@ -317,3 +319,170 @@ def check_hash_is_valid(dfh):
...
+            # the new folder name should be the same as the Excel file name
+            assert excel_file_path.name.split('.')[0] == path_to_new_directory.name
+            for sheet_name in xl.sheet_names:
+                path_to_new_file = Path(path_to_new_directory / f'{sheet_name}.csv')
+                # new csv file(s) should be created with name(s) resembling sheet name(s) in excel file
+                assert path_to_new_file.exists() and path_to_new_file.is_file()
+                assert sheet_name == path_to_new_file.name.split('.')[0]
+
+            # remove the newly created folders and files to keep the resources folder clean
+            shutil.rmtree(Path(str(excel_file_path).split('.')[0]))
```
matt-graham, line 472: I'd be very wary of using `shutil.rmtree` in tests, particularly with a programmatically constructed path, as it's difficult to predict where tests will be run from and whether this will actually delete the files you want to or something else unintended. For example if this line was run for a value of `excel_file_path` which contains a directory along the path with a name prefixed with `.` (a common pattern for hidden directories in Unix systems), then this line will attempt to remove the parent of that directory and all of its sub files / directories!

A safer pattern here would be to use [the `tmp_path` fixture ](https://docs.pytest.org/en/stable/how-to/tmp_path.html) and copy the relevant files required for conversion there and then rely the automatic clean up of the files in the `tmp_path` fixture directory at the end of the test function.

mnjowe:
Thanks @matt-graham for all the helpful comments above. I will address them soon

tests/test_utils.py
```diff
@@ -317,3 +319,170 @@ def check_hash_is_valid(dfh):
...
+            assert excel_param_dict == csv_param_dict
+
+        def initialise_population(self, population):
+            pass
+
+        def initialise_simulation(self, sim):
+            pass
+
+    dummy_module = DummyModule()
+    dummy_module.read_parameters(Path(resourcefilepath / 'ResourceFile_test_convert_to_csv'))
```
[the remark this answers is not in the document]
mnjowe: Hi @matt-graham . The only challenge with testing for matching dataframes  outside a disease module is that at this stage `read_csv_files` is outputting datframes with all data types as strings. At the other end `pandas.read_excel` is doing all the necessary data type formatting as such the two dataframes though having same values are appearing to differ in datatypes. That;s why I opted for a Dummy module to show that even though this is the case with input, it will be fine in the real modules because all parameters will be formatted accordingly using parameter datatypes defined in their respective disease modules. I don't know if you get what I mean?

tests/test_utils.py
mnjowe: Thanks. I have changed to now using temporal directory.

tests/test_utils.py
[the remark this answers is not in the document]
matt-graham: Hi @mnjowe. Ah good point around the data types, I hadn't realised the pandas `read_excel` and `read_csv` functions differed in how they handled data types. I can see then why it would make sense to check that we get equivalent parameter values read in whether we use an Excel or CSV resource file. As this is more about testing the functionality of the `load_parameters_from_dataframe` method rather than `read_csv_files`it might make more sense to have a test of this form in `tests/test_core.py` given the `Module.load_parameters_from_dataframe` method is defined in `src/tlo/core.py`. 

Just to note we can at least check that the loaded dataframes match in value when converting all columns to `str` dtype by adding something like 
```Python
    # values in dataframes when cast to strings should match
    for key in df_excel:
        assert df_excel[key].astype(str).equals(df_csv[key].astype(str))
```
to the end of the current `test_read_csv_files_method` test function.[](url)

matt-graham (approved):
Thanks @mnjowe for your updates and addressing my comments. Not sure if you are still making changes to this or not, but from my perspective it looks good to merge.

tests/test_utils.py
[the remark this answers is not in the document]
mnjowe: Thanks Matt and thanks for your suggestion on how the dataframes from these two methods can be tested. I'm okay replacing dummy module with this check. I feel that if we can confirm two dataframes from these two methods are equal, then we can be assured of equal output when these are passed to load_parameters function.

mnjowe:
Thanks @matt-graham. Just one more commit to address another of your comment on making tests small and simple.

matt-graham (approved):
Thanks @mnjowe for the updates to split up the test functions - this looks great!

PR #1426, opened 2024-07-17
Refactor cancer modules to reduce code duplication

willGraham01:


willGraham01:
Going to close this as a lot of the reported savings would also be achieved through a wider refactoring like that described in #1428

Issue #1428, opened 2024-07-18
Lots of `Modules` perform the same initial setting of DataFrame columns in `pre_initialise_simulation`


willGraham01:
Each `Module` subclass defines a list of `PROPERTIES` that it "owns" in the population DataFrame. During the `initialise_population` method, each `Module` subclass then sets the initial values for said columns.

However, the majority of `Module` subclasses seem to perform the same actions in this method - which is to just set the columns (defined in `self.PROPERTIES`) to the default value for the variable type of the property in question. As an example, consider the chronicsyndrome module - [the `PROPERTIES`](https://github.com/UCL/TLOmodel/blob/2dff6751ab1294ef42917880dc1ddc46b167e522/src/tlo/methods/chronicsyndrome.py#L83-L91) match up identically to [the columns set](https://github.com/UCL/TLOmodel/blob/2dff6751ab1294ef42917880dc1ddc46b167e522/src/tlo/methods/chronicsyndrome.py#L142-L147) in the population DataFrame during `initialise_population`.

Other modules do this and a bit more; `Alri` for example has [10 properties](https://github.com/UCL/TLOmodel/blob/2dff6751ab1294ef42917880dc1ddc46b167e522/src/tlo/methods/alri.py#L749-L799) which sets [an additional group of columns](https://github.com/UCL/TLOmodel/blob/2dff6751ab1294ef42917880dc1ddc46b167e522/src/tlo/methods/alri.py#L887) on top of those defined in it's `PROPERTIES`. However, it still needs to set the `PROPERTIES` columns to their defaults too.

We can likely automate this process for the vast majority of modules by implementing `initialise_population` generally in the `Module` class itself:

```python
def initialise_population(self, population: Population) -> None:
  df = population.props

  for property_name, property in self.PROPERTIES.items():
      df.loc[df.is_alive, property_name] = (
          property._default_value
          # If the property is a CATEGORICAL, we might have to lookup the correct default value to assign here
      )
```

Disease `Module`s that then need to do something _different_ to this can still overwrite `initialise_population` as usual. `Module`s that need to do something in addition to this can invoke `super().initialise_population` to copy these steps, before running the custom instructions they need in advance. `Module`s that just need to do these steps then don't even need to implement `initialise_population` explicitly.

This also means that if the names of the `PROPERTIES` for a given module are ever updated, they don't need to be changed in two places (within `PROPERTIES` and again in `initialise_population`).

PR #1429, opened 2024-07-19
Refactor setting of `PROPERTIES` columns across `Module`s

willGraham01:
Concerns #1428 |

Large portion of the additions in the diff are the `PROPERTIES` attributes of the modules being linted by `tox check` and split over multiple lines, so apologies for that to any reviewers.

Key changes introduced are within the `core` module though - specifically the new [`default_value`](https://github.com/UCL/TLOmodel/pull/1429/files#diff-61a3a14c3a4e4cd2a525077fbe1f370e706eaa458d9471819ceb27436b3304abR141) that `Property`s now take and the abstractly implemented [`initialise_population`](https://github.com/UCL/TLOmodel/pull/1429/files#diff-61a3a14c3a4e4cd2a525077fbe1f370e706eaa458d9471819ceb27436b3304abR366) method. Changes to disease modules should mostly be deletions of the explicit setting of columns and addition of default values for categories where necessary.

### DIFF
--- a/src/tlo/core.py
+++ b/src/tlo/core.py
@@ -8,12 +8,14 @@ from __future__ import annotations
 
 import json
 from enum import Enum, auto
-from typing import TYPE_CHECKING, Dict, FrozenSet
+from typing import TYPE_CHECKING, Any, Dict, FrozenSet, Optional, Set, Type
 
 import numpy as np
 import pandas as pd
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from typing import Optional
 
     from tlo.simulation import Simulation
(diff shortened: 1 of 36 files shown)


tamuri:
I suggest breaking this PR into several PRs. One adding the new functionality, the others refactoring specific modules to use the new functionality. This does mean that we end up supporting both old and new approaches for a time, but it is easier to work with modellers to update and check their modules and also identify any regressions easily.

willGraham01:
>I suggest breaking this PR into several PRs. One adding the new functionality, the others refactoring specific modules to use the new functionality. This does mean that we end up supporting both old and new approaches for a time, but it is easier to work with modellers to update and check their modules and also identify any regressions easily.

No problem, it should just be a case of cherry-picking the various commits from this branch (they're on a per-module basis). Will get started on this.

willGraham01:
Closing in favour of (the first of many) starting with #1436. Leaving the branch here for cherry-picking purposes.

PR #1433, opened 2024-07-25
Fix CMD-HIV Interaction

tbhallett:
This PR adds one line to the file ResourceFile_Improved_Healthsystem_And_Healthcare_Seeking.xlsx::sheet='hypertension_testing'

This is the line:
`rr_hiv	1`

The reason this is needed is because there is an element of the parameter set CardioMetabolicDisorders.parameters["hypertension_testing"], which is originally imported from ResourceFile_cmd_condition_testing.xlsx:::sheet='hypertension'.

Without this line, when using the ScenarioSwitcher, the parameters CardioMetabolicDisorders.parameters["hypertension_testing"] are over-written with the content of  ResourceFile_Improved_Healthsystem_And_Healthcare_Seeking.xlsx::sheet='hypertension_testing', and therefore the parameter `rr_hiv` is not available for use (in particular at `cardio_metabolic_disorders.py::67`)


RachelMurray-Watson (approved):
This has fixed the issue with the ScenarioSwitcher.

tdm32 (approved):

PR #1436, opened 2024-07-29
Allow `PROPERTIES` to set a default value, and Modules to auto-initialise their DF columns to these values

willGraham01:
Concerns #1428 | Cherry-pick of the global functionality that was proposed in #1429

Introduces the functionality needed for us to remove a lot of repeated code in the `initialise_population` method for `Module`s. `Property`s can now [have a default value](https://github.com/UCL/TLOmodel/blob/d93cb321007233a0e47a70107dff99eff4a3afb9/src/tlo/core.py#L153) set when they are created (falling back on the currently-provided defaults in `PANDAS_TYPE_DEFAULT_TYPE_MAP` if not provided) which will be used to fill the corresponding column in the population dataframe [during `initialise_population`](https://github.com/UCL/TLOmodel/blob/d93cb321007233a0e47a70107dff99eff4a3afb9/src/tlo/core.py#L400-L405).

`Module`s that just want to set their own property columns in `initialise_population` can then directly inherit this implementation without needing to explicitly implement it themselves. If `Module`s want to run additional steps in this method, they can call `super().initialise_population` first, and then provide their specific instructions afterwards. And finally, if a `Module` wants to do something completely different at this stage, they can just overwrite the method as normal and avoid calling `super()`.

As [@tamuri suggested](https://github.com/UCL/TLOmodel/pull/1429#issuecomment-2244828422), we'll introduce the changes proposed in #1429 incrementally. This does mean we'll be supporting both methods of setting the property columns in the dataframe for a while, but gives us some granularity for bug-fixing/tracing.


src/tlo/core.py
```diff
@@ -76,7 +80,7 @@ class Specifiable:
         Types.BITSET: int,
     }
 
-    def __init__(self, type_, description, categories=None):
+    def __init__(self, type_: Types, description: str, categories: Set[Any] = None):
```
matt-graham, line 83: Not sure the type hint for `categories` is appropriate here - it definitely needs to at least allow for `None` types (the default) but also the docstring implies a list of strings is required rather than a set. While I think the types allowed for the `categories` argument to a pandas `Categorical` are wider than that in practice, from a brief review it looks like most (potentially) all uses in `tlo` use a list of strings so I would say `list[str] | None` would be better type hint here.

matt-graham, line 83: Not sure the type hint for `categories` is appropriate here - it definitely needs to at least allow for `None` types (the default) but also the docstring implies a list of strings is required rather than a set. While I think the types allowed for the `categories` argument to a pandas `Categorical` are wider than that in practice, from a brief review it looks like most (potentially) all uses in `tlo` use a list of strings so I would say `list[str] | None` would be better type hint here.

src/tlo/core.py
```diff
@@ -94,16 +98,16 @@ def __init__(self, type_, description, categories=None):
             self.categories = categories
 
     @property
-    def python_type(self):
+    def python_type(self) -> Type[Any]:
```
matt-graham, line 101:
```suggestion
    def python_type(self) -> type:
```

[`Type[Any]` (and `type[Any]`) is equivalent to `type`](https://docs.python.org/3/library/typing.html#the-type-of-class-objects) and [`Type` is deprecated in from Python 3.9 onwards in favour of using `type` directly.](https://docs.python.org/3/library/typing.html#typing.Type)

src/tlo/core.py
matt-graham:
```suggestion
    def __init__(self, type_: Types, description: str, categories: list[str] | None = None):
```

src/tlo/core.py
```diff
@@ -201,48 +249,47 @@ class attribute on a subclass.
...
-    PROPERTIES = {}
+    PROPERTIES: Dict[str, Property] = {}
 
     # The explicit attributes of the module. We list these to distinguish dynamic
     # parameters created from the PARAMETERS specification.
     __slots__ = ('name', 'parameters', 'rng', 'sim')
 
-
-    def __init__(self, name=None):
+    def __init__(self, name: str = None) -> None:
```
matt-graham, line 292:
```suggestion
    def __init__(self, name: str | None = None) -> None:
```

Or `Optional[str]`

willGraham01:
Checks are failing due to some syntax in files that are not affected by this PR.

Review comments are addressed so happy to either wait for this to be fixed, or merge in now.

matt-graham:
> Checks are failing due to some syntax in files that are not affected by this PR.

Yeah this should be fixed by #1459

> Review comments are addressed so happy to either wait for this to be fixed, or merge in now.

Shouldn't matter either way as we're already failing on `master` 🙈 - ideally I guess we merge in #1459 first though.

matt-graham (approved):

PR #1437, opened 2024-07-29
Fix over-allocation of Bed Days in #1399

### LINKED ISSUE #1399: `BedDays` can over-allocate facility beds when combining footprints
Whilst working on speedups & unit tests in #1392, I noticed the following behaviour in the model. Not sure if this is intended/accepted, but it struck me as odd so mentioning it here.

At [present](https://github.com/UCL/TLOmodel/commit/6af737a820aa43f4d1bda9ce6edb058e52eeae30), when a person is already an inpatient and they are scheduled additional bed days at a facility, their remaining footprint [is combined with the incoming footprint](https://github.com/UCL/TLOmodel/blob/6af737a820aa43f4d1bda9ce6edb058e52eeae30/src/tlo/methods/bed_days.py#L309-L316) when allocating the total bed days they should be provided.

This can result in a facility being over-allocated in a particular type of bed. For example, consider the situation where (at some facility):
- A person with ID `p` is currently scheduled to occupy `bed_type1` on `2010-01-01` and `2010-01-02`.
- On `2010-01-01`, `p` is scheduled another appointment that imposes a footprint of `bed_type1` for 1 day, and `bed_type2` for 6 days.

"Combining" the existing footprint with the incoming one, as is what is currently done, results in:
- `p` occupies `bed_type1` on `2010-01-01` and `2010-01-02`,
- `p` occupies `bed_type2` on `2010-01-03` through to `2010-01-08`.

However, there is no guarantee that `bed_type2` is available on `2010-01-08` or `2010-01-07`. The incoming footprint was allocated assuming the footprint would start on `2010-01-01`, thus only reserving `bed_type2` space until `2010-01-06`. The existing footprint doesn't reserve any `bed_type2`s. As a result, the facility can be over-allocated for `bed_type2` on `2010-01-08` and `2010-01-07`. Since this logic applies to all HSI events, multiple patients can encounter the same issue, resulting in larger over-allocations on particular days.

Running the scale_run script for 1 year with an initial population of 10,000 will produce an example of this behaviour for patient 10107 on `2010-07-04`. They occupy a `general_bed` on `2010-07-04` for the rest of that day. However on `2010-07-04` they are also given a new allocation for 7 `non_bed_space` days, to start immediately. The resulting combined footprint runs until `2010-07-11`, however there is no guarantee that on `2010-07-11` there will be a bed for them.

## Possible Solutions

If there is meant to be a "consistency check" for cases such as this, the [`impose_beddays_footprint`](https://github.com/UCL/TLOmodel/blob/6af737a820aa43f4d1bda9ce6edb058e52eeae30/src/tlo/methods/bed_days.py#L287) function is too late to run it since bed days footprints are imposed [_after_ the HSI event has already run](https://github.com/UCL/TLOmodel/blob/6af737a820aa43f4d1bda9ce6edb058e52eeae30/src/tlo/methods/hsi_event.py#L177), so we cannot "undo" the HSI event that is in conflict. This check could be moved to happen earlier though, although I'm not sure on the exact location in the codebase as to where it should go.

Alternatively, it would be possible (at this stage in the code) to combine the two footprints by allocating the highest priority bed on any given day that they overlap, rather than extending the existing one. So in the case above;
- `p` would occupy `bed_type1` on `2010-01-01` and `2010-01-02`.
- `p` occupies `bed_type2` from `2010-01-03` through to `2010-01-07`.

Both the current and incoming footprints need `bed_type1` on `2010-01-01`. On `2010-01-02`, the `bed_type2` requirement from the incoming footprint can be satisfied by assigning the "higher-priority" occupation of `bed_type1` from the current footprint.
Then from `2010-01-03` we allocate the remaining 5 days of `bed_type2` from the incoming footprint.
This is guaranteed not to result in beds being over-allocated, since both footprints have checked for the availability of their respective beds, and we select the greatest of the two, rather than requesting both and allocating more bed-time without checking for it. However, it is **a distinct behaviour from that which is currently implemented**.

willGraham01:
Closes #1399 |

Fixes how bed days footprints are combined when a person is allocated another set of bed-days, but is already an inpatient. Implements the "2nd solution" in #1399; the higher-priority bed on each day is assigned. As an explicit example:

- Person `p` current has a footprint of `bed1` for 2 days and `bed2` for 0 days.
- `p` is to be assigned a second footprint of `bed1` for 1 days and `bed2` for 6 days.
- The resulting footprint lasts for 7 days:
  - On the first day, both footprints want `bed1`, which is allocated.
  - On the second day, the `bed1` need from the current footprint is higher priority than the `bed2` need from the second footprint. Thus, `bed1` is assigned.
  - For days 3 through 7, `bed2` is assigned.
  - This gives a "combined" or "resolved" footprint of `bed1` for 2 days, `bed2` for 5 days.

The `combine_footprints_for_same_patient` now handles combining the two footprints. `test_bed_days_in_isolation` now has a check that covers the case used as an example in #1399. 

NB: **This is a change in model behaviour**. Long simulations may start to provide different results if this goes in. This is because this change should allow more people - in the short term - to receive treatment than before the bug was fixed (since the bug caused a person to "block" their own treatment, and to occupy beds for longer than they should, depriving others of resources). However, the knock-on effects of the event-based simulation could mean that people who "benefited" from this misallocation and received treatment no longer do.


tbhallett (approved):
Thanks @willGraham01

tests/test_beddays.py
```diff
@@ -82,6 +82,28 @@ def test_beddays_in_isolation(tmpdir, seed):
...
     assert ([cap_bedtype1] * days_sim == tracker.values).all()
 
+    # 4) check that combining footprints for a person returns the expected output
+    # Test uses example fail case given in https://github.com/UCL/TLOmodel/issues/1399
+    # Person p has: bedtyp1 for 2 days, bedtype2 for 0 days.
+    # Person p then assigned: bedtype1 for 1 days, bedtype2 for 6 days.
+    # EXPECT: p's footprints are combined into bedtype1 for 2 days, bedtype2 for 5 days.
+    existing_footprint = {"bedtype1": 2, "bedtype2": 0}
+    incoming_footprint = {"bedtype1": 1, "bedtype2": 6}
+    expected_resolution = {"bedtype1": 2, "bedtype2": 5}
```
mnjowe, line 92: Are we suggesting that we will have a max of two bed type in a footprint? if not, can we also look at how this function will behave when a footprint containing three bed types is issued? Its a rare event of-course but I also understand we are allowed to request bed days from all bed types expect non-bed space.

mnjowe (approved):
Thanks @willGraham01 for the wonderful work on this. I have left one suggestion which isn't necessarily important(just thinking of all possibilities).

tests/test_beddays.py
willGraham01: The `BedDays.combine_footprints_for_same_patient` function should be general enough to handle any number of `bed_types`; it [loops through the `self.bed_types`](https://github.com/UCL/TLOmodel/blob/dfbc24c5cfda04f3657fbab517492e908c8dfcca/src/tlo/methods/bed_days.py#L357-L363) attribute so it will consider every bed type defined in the simulation (and thus, in the footprints).

Though you're correct that we should consider a more general case than just 2 beds - I've added another test case that involves 3 bed types.

willGraham01:
@mnjowe I've added another test as you suggested - I think you're right that it would be useful to have at least one test that involves more than 2 bed types being overwritter/reallocated.

Have tagged you for re-review again. If you're happy, then we can merge this by the looks of things :smile:

mnjowe (approved):
Thanks @willGraham01 for adding this test. Thanks also for fixing that bug in the last commit, that's exactly what prompted me to ask for testing this function with three bed types. I'm happy with all the changes, you may proceed to merging this in

PR #1439, opened 2024-07-29
Updates to the Horizontal and Vertical Programs Analysis

tbhallett:
* Remove absence scenario: Given discussion about the quantification of the impact of avoidable absence on availability of minutes of patient-facing time, we are removing an attempt to represent this in our scenarios

* Refactor so that we have definition of baseline, each hss element, and the hss package in one place; and that the hss package doesn't involve duplicates of the definition of each element

* Add loggers for HIV, Tb and Malaria


src/scripts/comparison_of_horizontal_and_vertical_programs/scenario_definitions.py
```diff
@@ -31,12 +31,6 @@ def _baseline(self) -> Dict:
...
 
-                    # Baseline scenario is with absence of HCW
-                    'year_HR_scaling_by_level_and_officer_type': self.YEAR_OF_CHANGE_FOR_HSS,
-                    'HR_scaling_by_level_and_officer_type_mode': 'with_absence',
-                    # todo <-- Do we want the first part of the run be with_abscence too...? (Although that will mean
-                    #          that there is actually greater capacity if we do the rescaling)
-
                     # Normalize the behaviour of Mode 2
                     "policy_name": "Naive",
                     "tclose_overwrite": 1,
```
tdm32, line 36: do we definitely want tclose to be so rigid?

tdm32 (approved):

src/scripts/comparison_of_horizontal_and_vertical_programs/scenario_definitions.py
tbhallett: which bit do you mean?

tdm32: setting tclose to 1

tbhallett: ah - I see. the '1' here just means "True" (it was from the time when the read-in from Excel of bools was unreliable). The actual value that does over-write tclose (in the line below)

PR #1441, opened 2024-07-29
Let the Scale-up of HIV, TB, Malaria services go to "MAXIMAL" levels (as well as target levels)

tdm32:
This PR partly addresses issue #1411 where programme scale-up should be 'target' or 'max'
The parameter for choosing programme scale-up has changed to 'type_of_scaleup' and requires a string, default value is 'none'
in def update_parameters_for_program_scaleup, the values for scale-up parameters are chosen based on this parameter

test_htm_scaleup is also updated 

script: comparison_of_vertical_programs/analysis_maxHTM_scenario.py runs scenarios with program scale-up. This only changes scale-up in hiv, tb and malaria and uses the default values for all other modules.


src/tlo/methods/enhanced_lifestyle.py
tbhallett, line 1: these changes from lifestyle coming in again!? Very weird! I'll have to roll back for the version to merge

src/scripts/comparison_of_horizontal_and_vertical_programs/analysis_maxHTM_scenario.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+if __name__ == '__main__':
+    from tlo.cli import scenario_run
+
+    scenario_run([__file__])
+
+
+
+# %% Produce some figures and summary info
```
tbhallett, line 119: is this the code that generates the figures you showed me this morning?

src/tlo/methods/tb.py
```diff
@@ -377,16 +377,16 @@ def __init__(self, name=None, resourcefilepath=None, run_with_checks=False):
             "length of inpatient stay for end-of-life TB patients",
         ),
         # ------------------ scale-up parameters for scenario analysis ------------------ #
-        "do_scaleup": Parameter(
-            Types.BOOL,
-            "argument to determine whether scale-up of program will be implemented"
+        "type_of_scaleup": Parameter(
+            Types.STRING, "argument to determine type scale-up of program which will be implemented, "
+                          "can be 'none', 'target' or 'max'",
         ),
```
tbhallett, line 383: to help us with checking etc, I think the `type_of_scaleup` should be defined as a categorical:

```suggestion
        "type_of_scaleup": Parameter(
            Types.CATEGORICAL, 
            "argument to determine type scale-up of program which will be implemented",
            categories=['none', 'target', 'max']
        ),
```

resources/ResourceFile_HIV.xlsx
tbhallett, line 1: these sheets seem fine. The only thing that I worry about is the column 'default' value' as this isn't linked to what is actually the default value on the other sheet, which could lead to confusion. I would suggest removing the columns 'default value'. 

And, while we're at it, also removing the column `module`.

tbhallett (changes requested):
Thanks Tara -- looks like it will do the job nicely!

src/tlo/methods/enhanced_lifestyle.py
tdm32: it won't run locally with np.timedelta64 value="M" for me:
enhanced_lifestly.py np.timedelta64 as ValueError: Unit M is not supported. Only unambiguous timedelta values durations are supported. Allowed units are 'W', 'D', 'h', 'm', 's', 'ms', 'us', 'ns'

src/scripts/comparison_of_horizontal_and_vertical_programs/analysis_maxHTM_scenario.py
tdm32: almost the same, edited to show just the 5 runs in the script (rather than mode1 vs mode2)

resources/ResourceFile_HIV.xlsx
tdm32: sure

src/tlo/methods/tb.py
tbhallett: That seems familiar actually. Okay leave it as string then. Please could you raise an issue so that we fix this for the future?

src/tlo/methods/enhanced_lifestyle.py
tbhallett: This is very weird. Matt/Asif would know better but I guess it's to do with perhaps your setup not being quite the same the specification? (In which case delete it and start again with the conda environment setup...?)

src/tlo/methods/tb.py
tdm32: Issue #1442 now raised

src/tlo/methods/enhanced_lifestyle.py
tdm32: oh no...

tbhallett (approved):

PR #1444, opened 2024-07-29
Update Consumable availability (New OpenLMIS data)

sakshimohan:
The previous version of consumable availability estimates were based on 2018 OpenLMIS data. This PR brings in 2021-2023 OpenLMIS data to update these estimates.


sakshimohan:
The final result of this analysis was that we did not observe a significant rising or falling trend in consumable availability during the 5-year period.

I have produced an estimate of **average annual % change in consumable availability** between 2018 and 2023 based on this model through the following steps:

1. Estimate a **logistic regression model of consumable availability** based on the data from 2018, 2021, 2022, and 2023 with random effects at the level of item. I had to drop facility-level random effects because of the time this was taking but again just as with the HHFA analysis, the variance between items is far greater than that between facilities so this is justifiable.
2. Based on the model, **predict availability** for each item by year (all 6 years between 2018-2023), facility level, consumable category, is_vital.
3. Calculate the % change in availability for each combination of characteristics between subsequent years.
4. Take an average of these Annual % change values to get the Average annual % change in availability during the period 2018-2023.
5. The attached .csv contains these estimates. See column - "**Average annual % change (2018 to 2023**)". Note that the values represent the proportional change, not %, i.e. they need to be multiplied by 100 to get percentage change.

[predicted_availability_mixedeffects_model.csv](https://github.com/user-attachments/files/16625239/predicted_availability_mixedeffects_model.csv)

The figures below show the interaction between the facility/consumable characteristics  and time. There is evidently a lot of noise. We need someone else to clean and analyse the data in more detail to understand what's really going on. 

predicted_values_by_programmatic_category
predicted_values_by_eml_category
predicted_values_by_fac_level

sakshimohan:
These are descriptive figures of consumable availability over time - 

mean_time_trend_by_category_FullData
mean_time_trend_by_fac_level_FullData
mean_time_trend_by_fac_level_FullData

Issue #1316, opened 2024-04-12
Avoiding usages of `object` dtype properties / columns in `RTI` module


matt-graham:
Using the `fullmodel` set of modules, the population dataframe currently shows the current distribution of column datatypes (reported using `DataFrame.info` method)
```
dtypes: bool(176), category(68), datetime64[ns](98), float64(11), int64(85), object(6)
```
All six of the columns with `object` datatype correspond to properties defined in the `RTI` module which are used to contain lists of injury codes

https://github.com/UCL/TLOmodel/blob/bd0a7632e1b0add9ae0ae8d3afc2a151bdd15f7d/src/tlo/methods/rti.py#L1051-L1058
https://github.com/UCL/TLOmodel/blob/bd0a7632e1b0add9ae0ae8d3afc2a151bdd15f7d/src/tlo/methods/rti.py#L1079-L1080

For a population dataframe with 51000 rows, the memory usage of each of these columns is around 2.3MiB compared to 0.4MiB for a corresponding column with 64-bit integers or floating point values.

While these columns currently use a list type, in reality I believe they are used as sets, with the lists used to store unique injury codes without duplicates and the order in the list abritrary. This suggests a bitset representation would be more parsimonious memory wise while also potentially reducing the cost of operations to update or check membership of the sets. However, the number of injury codes currently defined in

https://github.com/UCL/TLOmodel/blob/bd0a7632e1b0add9ae0ae8d3afc2a151bdd15f7d/src/tlo/methods/rti.py#L66-L73

is 97 which is too many to use with a bitset with 64-bit integer representation, and there aren't any 128-bit integer types available in Pandas (or the additional types available via PyArrow). 

One option would be to extend `BitsetHandler` to allow using columns with variable bit-width using the fixed-length string / bytes data dtype. For example

```Python 
series = pd.Series([""] * 10, dtype="S13")
```

creates a series with 10 entries with with 13 bytes = 104 bits per value (which would be the minimum number of bytes required to represent 97 injury codes). This can be viewed as an array of unsigned integers / bytes using

```Python
series.array.view("13B")
```
with output
```
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[...]

PR #1448, opened 2024-07-31
Make `BitsetHandler` variable width and a pandas extension

willGraham01:
Concerns #1316 |

Introduces the functionality for us to store bitsets as an extension to pandas, rather than having to use the `BitSetHandler` as an intermediary class.

`bitset` series (and thus dataframe columns) are stored as fixed-width bytestrings, where each character is represented by a `np.uint8`. The size of the bytestring (number of "characters") is chosen when the bitset is created to use the minimum number of bytes for each entry (IE we use 1 byte per 8 possible entries in a bitset).

The `BitsetArray` provides the necessary `ExtensionArray` instance for pandas series and dataframes. The operators like `<=, ==, <`, etc have been overloaded to perform their entry-wise `set` equivalents, so users should be able to interact with the series in the following manner:

```python
bdt = BitsetDtype(0, 1, 2, 3, 4, 5, 6, 7, 8)
a_series = pd.Series([{"1"}, {"4", "8"}, {"2", "7", "3"}], dtype=big_bdt)

# Return a boolean series indicating equality
big_s == {"4", "8"}
# Return a boolean series indicating entry-wise superset-ness
# Note this is equivalent to the 'in' functionality requested in
# https://github.com/UCL/TLOmodel/issues/1316#issuecomment-2052071639,
# but the 'in' operator has a particular meaning for series so we can't use it for this context.
big_s >= {"3"}
# Add an item to an entry
big_s.loc[0] += {"0"}
```

There are some more usage examples in the `bitset_extension.py` file, which is currently my quick-and-easy live testing ground.

### Tests

The basic pandas suite for testing `DtypeExtension`s passes, however we need some more tests for the functionality we are including on top. Primarily tests for how we handle different inputs when assigning and comparing (since allowing sets to be input values upsets pandas, since they are also iterables and so can be interpreted in a wonky way).


willGraham01:
@matt-graham I won't tag you for review just yet since there's still those bugs mentioned in the PR, but any comments you have (particularly if this was something you were [expecting in here](https://github.com/UCL/TLOmodel/issues/1316#issuecomment-2052071639)) would be appreciated :grin:

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+        return arr.view(f"{fixed_width}S").squeeze()
+
+    @classmethod
+    def _concat_same_type(cls, to_concat: Sequence[BitsetArray]) -> BitsetArray:
+        concat_data = np.concatenate(bsa._data for bsa in to_concat)
+        return cls(concat_data, to_concat[0].dtype)
+
+    @classmethod
+    def _from_sequence(
+        cls, scalars: Iterable[Set[NodeType]], *, dtype: BitsetDtype | None = None, copy: bool = False
```
matt-graham, line 275:
```suggestion
        cls, scalars: Iterable[Set[NodeType] | NodeType], *, dtype: BitsetDtype | None = None, copy: bool = False
```
For consistency with docstring

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+
+        Parameters
+        ----------
+        scalars : Sequence[Set[NodeType] | NodeType]
+            Sequence of sets of elements (or single-values to be interpreted as single-element sets)
+        dtype : dtype, optional
+            Cast to this datatype, only BitsetDtype is supported if not None. If None, an attempt will be made to
+            construct an appropriate BitsetDtype using the scalar values provided.
+        copy : bool, default False
+            If True, copy the underlying data.
```
matt-graham, line 288: I think this is a NumPy style docstring parameter list - ideally we should use normal Sphinx / rST type with `:param name:` directives. Also think type hints should make type descriptions for parameters redundant

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+        """
+        # Check that we have only been passed sets as scalars. Implicitly convert single-items to sets.
+        for s in scalars:
+            if not isinstance(s, set):
+                if isinstance(s, ALLOWABLE_ELEMENT_TYPES):
+                    s = set(s)
+                else:
+                    raise ValueError(f"{s} cannot be cast to an element of a bitset.")
```
matt-graham, line 296: I don't think this will do what it looks like you're intending. Setting `s = set(s)` will create a new `set` object which contains the object previously bound to name `s` and then bind to name `s` but won't affect original object bound to `s` from `scalars` iterable. Something like 

```suggestion
        for i, s in enumerate(scalars):
            if not isinstance(s, set):
                if isinstance(s, ALLOWABLE_ELEMENT_TYPES):
                    scalars[i] = set(s)
                else:
                    raise ValueError(f"{s} cannot be cast to an element of a bitset.")
```

would work though would rely on `scalars` being mutable (so not a tuple for example).

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+        assert all(isinstance(s, set) for s in scalars), "Not all scalars have been cast correctly."
```
matt-graham, line 297: I think this check should be redundant with suggested fix above

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+
+        # If no dtype has been provided, attempt to construct an appropriate BitsetDtype.
+        if dtype is None:
+            # Determine the elements in the bitset by looking through the scalars
+            all_elements = set()
+            for s in scalars:
+                # Take union of sets to form list of all possible Bitset elements
+                all_elements |= s
```
matt-graham, line 305:
```suggestion
            # Take union of sets to form list of all possible Bitset elements
            all_elements = set().union(scalars)
```

matt-graham (commented):
Wow this is looking great @willGraham01. I've added some initial comments but no major issues sprang out at me from a quick once over.

src/tlo/bitset_handler/bitset_extension.py
```diff
@@ -0,0 +1,730 @@
...
+    def construct_from_string(cls, string: str) -> BitsetDtype:
+        """
+        Construct an instance of this class by passing in a string of the form
+        that str(<instance of this class>) produces.
+        
+        That is, given a string of the form
+        bitset(#elements): e1, e2, e3, ...
+
+        this method will return a BitsetDtype with elements e1, e2, e3, ... etc.
+        """
```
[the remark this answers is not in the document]
willGraham01: > would be a better alternative anyway? Is there a strong reason to allow both formats?

Pandas is the reason. The `dtype` argument to `pd.Series` needs to be able to take a string as an acceptable value, and then the default test suite seems to imply (quite reasonably) that

```python
pd.Series(some_data, dtype=str(BitsetDtype(e1, e2, ..., en)))
```

should work. IE that the output of `BitsetDtype.__str__` should be an acceptable argument to `dtype` as well as an explicit instance of `BitsetDtype` itself.

My intention with `BitsetDtype.__str__` was to display

```
bitset(#elements): e1, e2, ..., en
```

so that `print` and `pd.Series` display correctly, whilst also providing the user with all the at-a-glance information they need (this thing is a bitset of n elements, here's an explicit list). Granted, we potentially don't need the element count, but if we have a lot of elements (or _long_ names for the elements, I think it's useful!).

As such, this means that whatever we choose to display here is what `pandas` expects to be passed to `dtype`. My implementation of `construct_from_string` actually allows us to leave off the `bitset(#elements)` part because I originally wrote it assuming the input could be identical to `__init__`, not knowing about this caveat until after getting the test suite to work!

TBH I'd like to continue being able to `construct_from_string` by just passing the elements, though I'm not attached to this. I will add the check that the number of elements is correct though! :sweat_smile:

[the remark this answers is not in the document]
matt-graham: Thanks for the detailed explanation 😁 - that all seems reasonable then!

tests/bitset_handler/test_bitset_pandas_dtype.py
```diff
@@ -0,0 +1,28 @@
+import re
+
+import pytest
+from pandas.tests.extension.base import BaseDtypeTests
+
+from tlo.bitset_handler.bitset_extension import BitsetDtype
+
+
+class TestMyDtype(BaseDtypeTests):
```
matt-graham, line 9:
```suggestion
class TestBitsetDtype(BaseDtypeTests):
```

Presumably name was taken from pandas example code, but probably something more specific here would make sense!

tests/bitset_handler/conftest.py
matt-graham, line 1: Might be worth adding a comment somewhere pointing to https://github.com/pandas-dev/pandas/blob/main/pandas/tests/extension/conftest.py or the permalink https://github.com/pandas-dev/pandas/blob/bdb509f95a8c0ff16530cedb01c2efc822c0d314/pandas/core/dtypes/dtypes.py indicating that this is implementing the fixtures specified in that file to allow use of standard test suite as I was initially a bit confused by things like we needed the `data_for_twos` fixture.

matt-graham (commented):
Thanks @willGraham01 for the updates in response to my comments and adding in the tests. This looks pretty much ready to merge to me? I think the checks fail is just due to a `isort` complaining about a missing new line.

tests/bitset_handler/test_bitset_pandas_dtype.py
willGraham01: No idea what you mean, totally didn't need to borrow code from any examples here... :innocent:

matt-graham (approved):
Looks all good to me

matt-graham:
With regards to the failing Pylint check - I did a bit of reading at it seems there are some on-going issues with handling definitions in `.pyi` files in NumPy specifically within Pylint (https://github.com/pylint-dev/astroid/pull/2375), so I would say as the error does appear to be a false positive here, the simplest solution would be to just add a comment to locally disable checking for E0611 in the relevant line, that is something like

```diff
-from numpy.dtypes import BytesDType
+from numpy.dtypes import BytesDType  # pylint: disable=E0611
```

willGraham01:
pylint update fixes everything - adding a reminder note here for me to press merge once CI passes.

PR #1452, opened 2024-08-02
Allow declaring and retriveing names associated with draws in the `Scenario` class

### LINKED ISSUE #1451: Save a label for the draw when using `Scenario` class
The `Scenario` class works using draw-numbers, but analysis scripts want to label results with "friendly names". The work-around being used currently is for these friendly names to be generated in the instance of the `Scenario` class, which is then used in the analysis class. (e.g. [this](https://github.com/UCL/TLOmodel/blob/master/src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/scenario_impact_of_healthsystem.py) scenario and the corresponding part of the analysis file: https://github.com/UCL/TLOmodel/blob/d9d3f62dd37c4a4cb278497e41207c091ffab699/src/scripts/overview_paper/C_impact_of_healthsystem_assumptions/analysis_impact_of_healthsystem.py#L41-L47)

This a bit cumbersome:
- analysis scripts need to import and use scenario classes, which limits how much we can recycle the scripts
- in a worse-case-scenario, the scenario class code and the results being analysed, become out of sync, and so the wrong labels are applied.

A good solution would enable retrieving such a friendly name from the results themselves.

One solution: the friendly name could be logged in `tlo.scenario` log, in the same manner of the parameters that are being over-ridden. A helper function (like this one for the parameters) could then retrieve it:

https://github.com/UCL/TLOmodel/blob/d9d3f62dd37c4a4cb278497e41207c091ffab699/src/tlo/analysis/utils.py#L199-L230

tbhallett:
Fixes #1451

This line can be added to the `Scenario` class:

```python
    def draw_names(self, draw_number) -> str:
        return list(self._scenarios.keys())[draw_number]
```

and in the analysis script, the draw_names can be retrieved using
```python
extract_draw_names(results_path)
```

Optionally, the existing `extract_parameters` function can index the parameters being changes by these same draw names.


src/tlo/scenario.py
```diff
@@ -234,6 +234,11 @@ def draw_parameters(self, draw_number, rng):
         """
         return None
 
+    def draw_name(self, draw_number) -> str:
```
tamuri, line 237: Do draw names need to come from a function i.e. need to be calculated in some interesting way? Instead, you could add a line in the scenario's constructor, such as:

```
self.draw_names = ['one', 'two', 'three']
```

or

```
self.draw_names = [f'draw_{i}' for i in range(0,10)]
```

?

tbhallett: They don't _need_ to come from a function, I don't think..... But, I thought it was good to follow the pattern established by `draw_parameters(self, draw_number, rng)`, and also this gives most flexibility.

Most often we define a dictionary of the form:
```python
self._scenarios = {'my_scenario_name': {'ModuleName': {'parameter_name': new_parameter_value}}} 
```
and so it's true that we could have instead in the `__init__`:
```python
self.draw_names = list(self._scenarios.keys()
```

src/tlo/scenario.py
```diff
@@ -234,6 +234,11 @@ def draw_parameters(self, draw_number, rng):
...
+        """Returns the name of the draw corresponding to the given draw number. This is offered for convenience so that
+        the logfile contain a 'user-friendly' label for the draw."""
+        return ""
```
tamuri, line 240: Can the default be this instead?

```suggestion
        return str(draw_number)
```

tamuri (approved):

PR #1455, opened 2024-08-15
PR to record the prevalence of disease/condition, still births,  neonatal death, maternal mortality

RachelMurray-Watson:
The point prevalence is recorded for a number of modules and conditions within modules (Alri, BladderCancer, BreastCancer, CardioMetabolicDisorders ( chronic_ischemic_hd, chronic_kidney_disease, chronic_lower_back_pain, diabetes, hypertension), COPD, Depression, Diarrhoea, Epilepsy, Hiv, Labor (Intrapartum stillbirth), Malaria, Measles, NewbornOutcomes, OesophagealCancer, OtherAdultCancer, PostnatalSupervisor, PregnancySupervisor (Antenatal stillbirth), ProstateCancer, RTI, Schisto, TB, Demography (maternal_deaths, newborn_deaths). 

This is still a draft as I had two major causes for questions: 
	- For some modules (RTI, ARLI, or heart attack and stroke in the CardioMetabolic module), may be more accurately described by calculating incidence. 
    - I have had some issues deriving a test. 
      
              - For the test, some logs don’t record prevalence (measles records incidence), and I can’t access the information I need from the logs to calculate a prevalence. 
              - For some modules (RTI, ARLI), the result is more accurately an incidence, not prevalence
              - In the malaria module, only clinical prevalence is recorded
          - With this in mind, is there an easily-accessible outside data source that has been used for calibrations etc with which I could compare?

Additional questions: 
    - Okay to calculate the prevalence of diarrhoea? It is a not really a disease in its own right, more of a symptom
    - Is there a better way to record maternal deaths, neonatal deaths, and stillbirths?


src/tlo/methods/tb.py
```diff
@@ -1009,6 +1009,16 @@ def report_daly_values(self):
 
         return health_values.loc[df.is_alive]
 
+    def report_prevalence(self):
```
tdm32, line 1012: I think here we should log only active cases, this way we can compare with WHO reports / GBD etc. Also the way that we assign latent cases is not identical to other models, we don't have infections -> latent -> active so we would under-estimate the latent infections. Best to stick to symptomatic active cases only.

tests/test_record_prevalence_healthburden_class.py
```diff
@@ -0,0 +1,365 @@
...
+popsize = 1000
+seed = 42
+do_sim = True
+def check_dtypes(simulation):
+    df = simulation.population.props
+    orig = simulation.population.new_row
+    assert (df.dtypes == orig.dtypes).all()
+
+
+def find_closest_recording(prevalence, target_date, log_value, column_name, multiply_by_pop):
```
tdm32, line 23: I'm not sure of the usefulness of finding the closest reported prevalence value. Would a useful test perhaps be to check all registered modules are logging prevalence every month and the inverse of this (no prevalence reported if module not registered), or set the incidence to 0 for one disease and check logger not reporting anything above 0, assert prevalence values for 2010 within reasonable range, e.g. test for extreme or unlikely values.

src/tlo/methods/healthburden.py
```diff
@@ -116,12 +119,16 @@ def initialise_simulation(self, sim):
...
-        # 4) Launch the DALY Logger to run every month, starting with the end of the first month of simulation
-        sim.schedule_event(Get_Current_DALYS(self), sim.date + DateOffset(months=1))
-
+        # 4) Launch the DALY and Prevalence Logger to run every month, starting with the end of the first month of simulation
         # 5) Schedule `Healthburden_WriteToLog` that will write to log annually
-        last_day_of_the_year = Date(sim.date.year, 12, 31)
-        sim.schedule_event(Healthburden_WriteToLog(self), last_day_of_the_year)
+        sim.schedule_event(Get_Current_DALYS(self), sim.date + DateOffset(months=1))
+        if self.parameters['test']:
+            sim.schedule_event(Get_Current_Prevalence(self), sim.date + DateOffset(months=1))
```
tdm32, line 126: could we make a choice here for the frequency of logging - if we are interested in very rapidly developing/resolving conditions we could set to daily logger, for broader analyses we could set to annual!?

src/tlo/methods/demography.py
```diff
@@ -383,6 +383,54 @@ def on_birth(self, mother_id, child_id):
...
+                                                 (df['date_of_death'] >= (self.sim.date - DateOffset(months=1)))
+                                                 ])
+                indirect_deaths_non_hiv = len(df.loc[
+                                                  (df['is_pregnant'] | df['la_is_postpartum']) &
+                                                  (df['cause_of_death'].str.contains(
+                                                      'Malaria|Suicide|ever_stroke|diabetes|chronic_ischemic_hd|ever_heart_attack|chronic_kidney_disease') |
+                                                   (df['cause_of_death'] == 'TB')) &
+                                                  (df['date_of_death'] >= (self.sim.date - DateOffset(months=1)))
+                                                  ])
+                direct_deaths_non_hiv = len(df.loc[
```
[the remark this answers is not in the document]
RachelMurray-Watson: Sorry! Yes well spotted, have changed that now

src/tlo/methods/healthburden.py
```diff
@@ -643,21 +682,80 @@ def apply(self, population):
...
+        if not self.module.recognised_modules_names or not self.module.causes_of_disability:
+            return
+
+        # Calculate the population size
+        population_size = len(self.sim.population.props[self.sim.population.props['is_alive']])
+
+        # Create a DataFrame with one row and assign the population size
+        prevalence_from_each_disease_module = pd.DataFrame({'population': [population_size]})
+        for disease_module_name in self.module.recognised_modules_names:
+            if disease_module_name in ['NewbornOutcomes', 'PostnatalSupervisor', 'Mockitis', 'DiseaseThatCausesA',
```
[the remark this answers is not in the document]
RachelMurray-Watson: Mockitis is now removed form the list, as it is being included in the test as a dummy check of prevalences. For the others, however, they do not return any prevalences/conditions, and to avoid a log of lots of columns of zeroes, I have it so that they're skipped over here.

src/tlo/methods/rti.py
```diff
@@ -2314,6 +2314,15 @@ def report_daly_values(self):
         disability_series_for_alive_persons = df.loc[df.is_alive, "rt_disability"]
         return disability_series_for_alive_persons
 
+    def report_prevalence(self):
```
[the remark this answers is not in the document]
RachelMurray-Watson: Okay, I think I had 
         total_prev = len(
             df[(df['is_alive']) & (df['rt_inj_severity'] != 'none')]
         ) / len(df[df['is_alive']])

         return total_prev
     Originally, but I think Margherita implied that this would  include a consideration of recovery time. Do you think it's okay to use, still?

src/tlo/methods/schisto.py
```diff
@@ -214,6 +214,14 @@ def get_total_disability_weight(list_of_symptoms: list) -> float:
         return pd.Series(index=df.index[df.is_alive], data=0.0).add(disability_weights_for_each_person_with_symptoms,
                                                                     fill_value=0.0)
 
+    def report_prevalence(self):
+        # This returns dataframe that reports on the prevalence of schisto for all individuals
+        df = self.sim.population.props
+        is_infected = (df[self.cols_of_infection_status] == 'Non-infected').any()
```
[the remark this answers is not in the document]
RachelMurray-Watson: Ah sorry I see now it should have been !=, but I have changed to your way for clarity, thank you!

src/tlo/methods/tb.py
RachelMurray-Watson: Okay, that's great to know, thank you!

tests/test_record_prevalence_healthburden_class.py
RachelMurray-Watson: We had a long discussion about the test yesterday on our call Part II. We decided to include a dummy disease with which to compare prevalences (this will be in the new test file), and to see if the prevalence of what it reports matches with what has been reported it its own logging file. 

I suppose by doing it the way that I was doing it, I was trying to see if the calculations themselves were working, as well as the general mechanics of logging. But do you think such a test is unnecessary? And that by showing e.g. with a dummy module and/or what you have suggested above, it would suffice?

src/tlo/methods/healthburden.py
```diff
@@ -529,8 +536,40 @@ def log_df_line_by_line(key, description, df, force_cols=None) -> None:
...
+                logger.info(
+                    key=key,
+                    data=row.to_dict(),
+                    description=description,
+                )
+
+        # Check that the format of the internal storage is as expected.
+        self.check_multi_index()
+
+        log_df_line_by_line(
```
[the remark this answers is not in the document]
RachelMurray-Watson: Okay, I have added that!

RachelMurray-Watson:
> so that it's formally part of the definition of a disease module

Grand! Have that done

src/tlo/methods/healthburden.py
RachelMurray-Watson: I have included a parameter now that allows us to set the time of logging as either daily, monthly, or yearly

src/tlo/methods/healthburden.py
```diff
@@ -529,8 +536,40 @@ def log_df_line_by_line(key, description, df, force_cols=None) -> None:
...
-
+        # 5) Log the prevalence of each disease
+        log_df_line_by_line(
+            key='prevalence_of_diseases',
+            description='Prevalence of each disease.',
+            df=self.prevalence_of_diseases,
+            force_cols=self.recognised_modules_names,
+        )
         self._years_written_to_log += [year]
+    def write_to_log_prevalence_monthly(self):
```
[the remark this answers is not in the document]
RachelMurray-Watson: I have made this change below

src/tlo/methods/healthburden.py
```diff
@@ -99,6 +101,7 @@ def initialise_simulation(self, sim):
         self.years_life_lost_stacked_time = pd.DataFrame(index=self.multi_index_for_age_and_wealth_and_time)
         self.years_life_lost_stacked_age_and_time = pd.DataFrame(index=self.multi_index_for_age_and_wealth_and_time)
         self.years_lived_with_disability = pd.DataFrame(index=self.multi_index_for_age_and_wealth_and_time)
+        self.prevalence_of_diseases = pd.DataFrame(index=year_index)
```
tbhallett, line 104: if we want it to take different frequencies (e.g month, year etc) then we'd need a difference index.

src/tlo/methods/healthburden.py
```diff
@@ -116,10 +119,21 @@ def initialise_simulation(self, sim):
         self.process_causes_of_disability()
         self.process_causes_of_dalys()
 
-        # 4) Launch the DALY Logger to run every month, starting with the end of the first month of simulation
+        # 4) Launch the DALY and Prevalence Logger to run every month, starting with the end of the first month of simulation
+        # 5) Schedule `Healthburden_WriteToLog` that will write to log annually
         sim.schedule_event(Get_Current_DALYS(self), sim.date + DateOffset(months=1))
+        if self.parameters['logging_frequency_prevalence'] == 'day':
+            sim.schedule_event(Get_Current_Prevalence(self), sim.date + DateOffset(days=0))
+            sim.schedule_event(Healthburden_WriteToLog_Prevalences(self), sim.date + DateOffset(days=0))
```
tbhallett, line 127: if these two events happen at the same frequency and we want to guarantee the order they happen in, I think they should be ONE event.

src/tlo/methods/healthburden.py
```diff
@@ -529,8 +543,61 @@ def log_df_line_by_line(key, description, df, force_cols=None) -> None:
...
             force_cols=self._causes_of_dalys,
         )
-
+        # 5) Log the prevalence of each disease
+        log_df_line_by_line(
+            key='prevalence_of_diseases',
+            description='Prevalence of each disease.',
+            df=self.prevalence_of_diseases,
+            force_cols=self.recognised_modules_names,
+        )
```
tbhallett, line 552: is this  unintentionally left-in? The function is defined below (embedded in write-to-log-prevalence)

src/tlo/methods/healthburden.py
```diff
@@ -529,8 +543,61 @@ def log_df_line_by_line(key, description, df, force_cols=None) -> None:
...
         self._years_written_to_log += [year]
+    def write_to_log_prevalence(self):
+        """Write to the log the prevalence of conditions .
+        N.B. This is called at the end of the simulation as well as at the end of each month, so we need to check that
+        the year is not being written to the log more than once."""
+        def log_df_line_by_line(key, description, df, force_cols=None) -> None:
```
tbhallett, line 558: this is a copy-and-paste of the function of the same defined as above. It's better to refactor than duplicate: so I would propose refactoring the original to be a function on this module and then calling it from here, rather than redefining it.

src/tlo/methods/healthburden.py
```diff
@@ -661,3 +782,17 @@ def __init__(self, module):
 
     def apply(self, population):
         self.module.write_to_log(year=self.sim.date.year)
+
+class Healthburden_WriteToLog_Prevalences(RegularEvent, PopulationScopeEventMixin):
+    """ This event with a specified frequency to record the prevalence logger
+    Added test to log daily if it is a test"""
+    def __init__(self, module):
```
tbhallett, line 789: this would be easier i..e pass in th frequency (as a DateOffset) as an argument when initiasiing the event.
```suggestion
def __init__(self, module, frequency: pd.DateOffset):
    super().__init__(
            module, frequency=frequency
        ) 
````

File: src/tlo/methods/healthburden.py [hunk 15]
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+        prevalence_from_each_disease_module.drop(
+            prevalence_from_each_disease_module.index.intersection(
+                ['NewbornOutcomes', 'PostnatalSupervisor', 'DiseaseThatCausesA', 'ChronicSyndrome']
+            ),
+            axis=0, inplace=True
+        )
+        self.module.prevalence_of_diseases = prevalence_from_each_disease_module
+
+
 class Healthburden_WriteToLog(RegularEvent, PopulationScopeEventMixin):
```
tbhallett, line 776: As mentioned above, I would combine these two events

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+        if not self.module.recognised_modules_names or not self.module.causes_of_disability:
+            return
+
+        # Calculate the population size
+        population_size = len(self.sim.population.props[self.sim.population.props['is_alive']])
+
+        # Create a DataFrame with one row and assign the population size
+        prevalence_from_each_disease_module = pd.DataFrame({'population': [population_size]})
+        for disease_module_name in self.module.recognised_modules_names:
+            if disease_module_name in ['NewbornOutcomes', 'PostnatalSupervisor', 'DiseaseThatCausesA',
```
tbhallett, line 741: why do we skip these? Can these modules instead just not have such a member function redefined from the base class, and the base class returning None? That would be easier. In most cases, we want framework stuff like this to be agnostic the modules that are defined and not have to worry about 'special cases'

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+                                       'ChronicSyndrome']:
+                continue
+
+            disease_module = self.sim.modules[disease_module_name]
+            prevalence_from_disease_module = disease_module.report_prevalence()
+
+            if disease_module_name == "CardioMetabolicDisorders":
```
tbhallett, line 748: as above, rather than be a special case, can we insetad detect if the thing being returned in a pd.DataFrame or a pd.Series and take it from there (e.g. if instance(prevalence_from_disease_module, pd.DataFrame))...)

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+                for i, column_name in enumerate(prevalence_from_disease_module.columns):
+                    prevalence_from_each_disease_module[column_name] = prevalence_from_disease_module.iloc[:, i]
+            else:
+                prevalence_from_disease_module = pd.DataFrame([[prevalence_from_disease_module]])
+
+                column_name = ("Intrapartum stillbirth" if disease_module_name == "Labour" else
```
tbhallett, line 754: this kind of relabelling is better done in the module that is doing the reporting

src/tlo/methods/skeleton.py
```diff
@@ -148,6 +149,16 @@ def report_daly_values(self):
         """
         raise NotImplementedError
 
+    def report_prevalence(self):
+        """
+        This must send back a pd.Series or pd.DataFrame that reports on the prevalence of a diseas/condition or the monthly
+        rate of certain events.
+        If multiple conditions in a module are defined, a pd.DataFrame must be returned with a column
```
tbhallett, line 156: this part of the docstring doesn't seem right (left-over from one of the other docstrings?). Also, does it need to be a pd.Series that we're returning? I wonder if instead a simple and more flexible type (e.g. dict) would be suitable.

src/tlo/methods/skeleton.py
```diff
@@ -148,6 +149,16 @@ def report_daly_values(self):
         """
         raise NotImplementedError
 
+    def report_prevalence(self):
```
tbhallett, line 152: btw - its' great you've put this here --- but this is not the base class of Module. That' found in core.py. That's where we can define the 'default' forms of the functions for the disease module. So I think for report_prevalence we can have a function there which just return None. That way, if a disease module doesn't define its own version of that function (over-writing the default) there is still something to call. We just need to add some logic so make sure that when we get a None returned, we know what it means and handle accordingly.

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+                               "Antenatal stillbirth" if disease_module_name == "PregnancySupervisor" else
+                               disease_module_name)
+
+                # Add the prevalence data as a new column to the DataFrame
+                prevalence_from_each_disease_module[column_name] = prevalence_from_disease_module.iloc[:, 0]
```
tbhallett, line 759: in most cases, there module is returning a single number; sometimes a set of numbers. Could this be defined by a dict instead, for simplicity?

tests/test_record_prevalence_healthburden_class.py
```diff
@@ -0,0 +1,161 @@
...
+from tlo.methods.fullmodel import fullmodel
+
+resourcefilepath = Path(os.path.dirname(__file__)) / '../resources'
+outputpath = Path("./outputs/")
+
+start_date = Date(2010, 1, 1)
+end_date = Date(2012, 1, 1)
+
+popsize = 1000
+seed = 42
```
tbhallett, line 23: seed can be a 'magical' kwarg to a function beginning with test and pyyest will populate it for you. (same as tmpdir)

tests/test_record_prevalence_healthburden_class.py
```diff
@@ -0,0 +1,161 @@
...
+
+    max_date_in_prevalence = max(prevalence['date'])
+
+    for j in range(len(prevalence_mockitis_log)):
+         target_date = prevalence_mockitis_log['date'][j]
+         regular_log_value = (
+             prevalence_mockitis_log["TotalInf"][j])
+
+         if target_date <= max_date_in_prevalence:
+             find_closest_recording(prevalence, target_date, regular_log_value, 'Mockitis', True)
```
tbhallett, line 159: do we need this extra step of finding the closest recording. For this dummy we can set the logging frquency to be the same, so that we can do a straight-forward comparison, can't we?

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
+
+    def apply(self, population):
+        if not self.module.recognised_modules_names or not self.module.causes_of_disability:
+            return
+
+        # Calculate the population size
+        population_size = len(self.sim.population.props[self.sim.population.props['is_alive']])
+
+        # Create a DataFrame with one row and assign the population size
+        prevalence_from_each_disease_module = pd.DataFrame({'population': [population_size]})
```
tbhallett, line 739: I'm wondering why this is dataframe

src/tlo/methods/healthburden.py
```diff
@@ -643,15 +710,69 @@ def apply(self, population):
...
 
 
+class Get_Current_Prevalence(RegularEvent, PopulationScopeEventMixin):
+    """
+    This event runs every month and asks each disease module to report the prevalence of each disease
+    during the previous month.
+    """
+
+    def __init__(self, module):
+            super().__init__(module, frequency=DateOffset(months=1))
```
tbhallett, line 729: I think its very murky if we allow this frquency to be differnt to the logging frequncy, That's why I think combine these two events into one.

tbhallett (changes requested):

src/tlo/methods/healthburden.py
```diff
@@ -58,7 +59,8 @@ def __init__(self, name=None, resourcefilepath=None):
         'Age_Limit_For_YLL': Parameter(
             Types.REAL, 'The age up to which deaths are recorded as having induced a lost of life years'),
         'gbd_causes_of_disability': Parameter(
-            Types.LIST, 'List of the strings of causes of disability defined in the GBD data')
+            Types.LIST, 'List of the strings of causes of disability defined in the GBD data'),
+        'logging_frequency_prevalence': Parameter(Types.BOOL, 'Set to the frequency at which we want to make calculations of the prevalence logger')
```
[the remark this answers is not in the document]
RachelMurray-Watson: Sorry, yes, that was a hangover from an earlier iteration. Corrected now.

src/tlo/methods/healthburden.py
RachelMurray-Watson: Ah okay, I understand. I thought this may prevent unwieldy log files with lots of Nones but I understand why this is preferable.

src/tlo/methods/skeleton.py
RachelMurray-Watson: Ah I see - I got confused, because when I looked at the one in core.py, there was no report_dalys function, so I supposed that I misinterpreted what you said (as it was to be like the report_daly function). I could add report_dalys whilst I'm at it, too, then?

src/tlo/methods/healthburden.py
RachelMurray-Watson: Leaving here for reference - if I don't skip DiseaseThatCausesA, the test for the test_healthburden.py fails (module created in that test)

tests/test_record_prevalence_healthburden_class.py
RachelMurray-Watson: I think mockitis logs every 6 months, so rather than setting a more detailed frequency for logging, I have kept the "closest match" date here.

src/tlo/methods/healthburden.py
RachelMurray-Watson: I think to use the log_df_line_by_line, it needs to be a dataframe, but I may have misinterpreted

src/tlo/methods/healthburden.py
RachelMurray-Watson: I initiated it a sa dataframe that I could then populate. I think for the the logging line-by-line function it needs to be a dataframe (like the DALYs), but I may have misinterpreted.

RachelMurray-Watson:
> just adding here the comment made in-person: I like the way this is being done overall.... and we should add this new method to the `Module` base class so that it's formally part of the definition of a disease module (in the same way that `report_dalys` is.)

(Based on conversation yesterday) - changed so that it is no longer in base class (as not all modules are disease modules), but there is an assertion checking to see that if something uses the healthburden module, it must have the report_prevalence function

src/tlo/methods/healthburden.py
RachelMurray-Watson: EDIT: Have now changed so that the prevalence is now collected as dictionaries

src/tlo/methods/healthburden.py
RachelMurray-Watson: EDIT: are all now dictionaries

tests/test_record_prevalence_healthburden_class.py
RachelMurray-Watson: Edited: new dummydisease in mockitis file. Set the logging frequency to be the same, so no need for this function

src/tlo/methods/malaria.py
```diff
@@ -755,6 +755,16 @@ def report_daly_values(self):
 
         return health_values.loc[df.is_alive]  # returns the series
 
+
+    def report_prevalence(self):
```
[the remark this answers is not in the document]
tdm32: Thanks for changing this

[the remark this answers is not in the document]
tdm32: Just to note also, if reporting point prevalence each month, this could potentially miss some cases of malaria which occur and then resolve within the month. One other option could be to use the property ma_date_symptoms to find all malaria cases who have had onset of symptoms within the last time period. Whichever way you prefer is ok.

src/tlo/methods/rti.py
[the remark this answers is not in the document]
tdm32: Could possibly select cases using property rt_date_inj to make sure injury occurred in most recent time period!?

src/tlo/methods/tb.py
tdm32: Thanks for making the change

src/tlo/methods/malaria.py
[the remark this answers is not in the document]
RachelMurray-Watson: Oh interesting! I suppose most of the other modules are also point prevalence, so maybe for consistency with that, we keep it this way. But if having more of a period prevalence is more useful to you/in general, happy to change it!

src/tlo/methods/rti.py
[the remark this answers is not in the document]
RachelMurray-Watson: Oh good idea! I've made that change now

PR #1456, opened 2024-08-20
scale-up VMMC from 2019 onwards to match data

tdm32:
Recently acquired data shows a rapid scale-up of VMMC from 2019 onwards. Currently in the HIV model we have falling VMMC rates due to the drop in child VMMC from 2020. Therefore a new parameter increase_in_prob_circ_2019 applies a scaling factor to the probability of VMMC following negative HIV test to bring the rate closer to the reported data. This is fixed for 2019 onwards. 

In future, this can be more nuanced and match the data more closely, but for now this will suffice. 

image

image
image
image


src/tlo/methods/hiv.py
```diff
@@ -597,6 +602,11 @@ def _build_linear_models(self):
             p["prob_circ_after_hiv_test"],
             Predictor("hv_inf").when(False, 1.0).otherwise(0.0),
             Predictor("sex").when("M", 1.0).otherwise(0.0),
+            Predictor("year",
+                      external=True,
+                      conditions_are_mutually_exclusive=True,
+                      conditions_are_exhaustive=True).when("<2019", 1)
```
tbhallett, line 608: as we're using `otherwise`, I don't think the we can say that the (`when`) conditions are exhaustive
```suggestion
                      ).when("<2019", 1)
```

tbhallett (changes requested):

tbhallett:
good catch!

tdm32:
Closed as changes incorporated into PR #1520

PR #1459, opened 2024-09-09
Use `isinstance` instead of checking type equality

matt-graham:
[Scheduled checks run failed on `master`](https://github.com/UCL/TLOmodel/actions/runs/10764131778/job/29846489383) due to an explicit comparison between `type` evaluated on an object an a type rather than using `isinstance` (or `is` with type).


matt-graham:
@willGraham01 tagging you for a review if that's okay as this should hopefully be quick as it's a couple of line change!

willGraham01 (approved):

PR #1460, opened 2024-09-09
Use `pytest-results-actions` to generate test result summary

matt-graham:
Experiment with using https://github.com/pmeier/pytest-results-action to try to generate summaries of test results in workflow summary to allow easier parsing of failed tests


matt-graham:
Example of generated summary https://github.com/UCL/TLOmodel/actions/runs/10769229850?pr=1460#summary-29860010293

For case of all passing tests not all that useful, but more interestingly with failing tests we should get expandable details of failing tests as in https://github.com/pmeier/pytest-results-action/actions/runs/10193348678/attempts/1#summary-28197796517

PR #1463, opened 2024-09-23
Enable all pycodestyle and isort rules in Ruff and fix violations

matt-graham:
Updates Ruff configuration to enable all [pycodestyle error (`E` prefix)](https://docs.astral.sh/ruff/rules/#error-e), [pycodestyle warning (`W` prefix)](https://docs.astral.sh/ruff/rules/#warning-w) and [isort (`I` prefix)](https://docs.astral.sh/ruff/rules/#isort-i) rules. Previously we used the [default Ruff configuration](https://docs.astral.sh/ruff/configuration/) which only selects a subset of the E rules (E4, E7 and E9), no W rules and doesn't enable any isort rules (but we separately used isort package to check import ordering).

The only violations of the additional pycodestyle error rules now enabled were of the [E501 `lines-too-long`](https://docs.astral.sh/ruff/rules/line-too-long/) rule, and this PR also fixes all instances of this rule being violated in `src/tlo` and `tests`, and adds a global directory level ignore for `src/scripts` (under the assumption we probably don't care too much about formatting violations in the scripts).

The only violations of the additional pycodestyle warning rules enabled were of [W291 `trailing-whitespace`](https://docs.astral.sh/ruff/rules/trailing-whitespace/) and [W293 `blank-line-with-whitespace`](https://docs.astral.sh/ruff/rules/blank-line-with-whitespace/) rules, with the this PR also fixing all instances of this rule being violated.

[Ruff's isort rules are very close to the rules checked by isort itself](https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort) and the only violation after enabling isort rules was due to presence of a commented out import in [src/scripts/hiv/projections_jan2023/analysis_full_model.py](https://github.com/UCL/TLOmodel/compare/mmg/stricter-ruff-rule-selection?expand=1#diff-95e0a9f90d823484b67a2f957a5ece87fd69fc910610d5964d3dc0e323e4cf26). The main advantages of using Ruff rules rather than separate isort are a slight improvement to run time of checks by not requiring separate installation and running of isort, and more usefully, for those who use a Ruff integration in their IDE, unsorted imports will now be flagged directly in IDE. I am still commonly caught out by the isort check failing on pushing regularly, so this should hopefully make this less likely by making it easier to catch violations earlier!

While this PR touches a lot of files, all of the changes should have no effect on actual code as they are just changes to whitespace / line reflowing (hiding whitespace changes in diff should make it a bit easier to review!).


src/tlo/methods/hiv.py
matt-graham, line 1: @tdm32 Have you got any active branches that update `src/tlo/methods/hiv.py` that the below changes (reformatting lines to make them conform with 120 character line length limit) might conflict with? If so we can hold off merging this until those changes are in to `master` and then update here.

PR #1466, opened 2024-09-26
Show total time in flat profile output

matt-graham:
Now that joerick/pyinstrument#285 has been merged and made it into recent `pyinstrument` releases (from version 4.7.0 onwards) , we can switch to showing total time spent in each function (included nested function calls) in the flat profile output added in #1236 rather than time just within the function itself (with the total time being useful for getting a sense of which function calls are dominating runtime, irrespective of call path).


matt-graham:
@tamuri this should be quite quick to review as it's just a few lines of changes and is just changing the output options for the profiling script so not affecting any of the model code.

tamuri (approved):

PR #1471, opened 2024-10-03
Add script for visualizing module dependencies as graph

matt-graham:
Wrote this a while back but forgot to add to repository!

Uses `pydot` to create a basic visualization of module dependency graph.

@RachelMurray-Watson - this is the script we discussed in chat today.


tamuri (approved):
Nice!

PR #1475, opened 2024-10-04
Updated declared dependencies in disease modules

RachelMurray-Watson:


tbhallett:
Hi @RachelMurray-Watson 
Thanks very much for this. I _**really**_ like these graphs.
I can see there is some little issue about the dependencies. I'll ask @matt-graham to review this.
I'm wondering whether it's possible to make the graphs without actually touching the modules in this PR (?), or whether the process of making these graphs has revealed some oversights in the declaration of dependencies that had previously gone unnoticed (?)

RachelMurray-Watson:
> Hi @RachelMurray-Watson Thanks very much for this. I _**really**_ like these graphs. I can see there is some little issue about the dependencies. I'll ask @matt-graham to review this. I'm wondering whether it's possible to make the graphs without actually touching the modules in this PR (?), or whether the process of making these graphs has revealed some oversights in the declaration of dependencies that had previously gone unnoticed (?)

Yes! I actually meant to come back to this, but for the property-level graphs, I don't actually need to know the module dependencies at all, so I can remove that from this PR, and perhaps add it as an issue that not all modules declare all of their dependencies?

tbhallett:
> Yes! I actually meant to come back to this, but for the property-level graphs, I don't actually need to know the module dependencies at all, so I can remove that from this PR, and perhaps add it as an issue that not all modules declare all of their dependencies?

That would be perfect!

RachelMurray-Watson:
> > Yes! I actually meant to come back to this, but for the property-level graphs, I don't actually need to know the module dependencies at all, so I can remove that from this PR, and perhaps add it as an issue that not all modules declare all of their dependencies?
> 
> That would be perfect!

Done!

src/tlo/analysis/module_dependencies_graph.py
tbhallett, line 1: I would move this file into a folder inside src/scripts

(Stuf in src/tlo is "core framework" stuff, whereas this is more akin to an analysis of the model)

tbhallett (changes requested):

src/tlo/analysis/module_dependencies_graph.py
RachelMurray-Watson: Sorry! Done that

tbhallett (approved):

src/tlo/analysis/module_dependencies_graph.py
tbhallett: Thanks!

PR #1477, opened 2024-10-07
Improve MNH logging 

joehcollins:
Replace current system of incidence/outcome logging for MNH modules in which all heavy lifting (i.e. calculation of incidence/mortality rates) is done in the analysis script. Now yearly rates will be outputted from logging events to make analysis scripts more readable and to reduce excess repeated coding within such scripts.


tbhallett (approved):
These seem fine to me

PR #1482, opened 2024-10-09
Logger in HealthSystem to report Number of Staff and PFT at start of each year

tbhallett:


marghe-molaro:
@tbhallett this is now implemented as follows:
- _daily_capabilities_per_staff is initialised in format_daily_capabilities alongside _daily_capabilities, by using the 'Total_Mins_Per_Day' and 'Staff_Count' information provided by the Resource file. (Note: this initialisation could probably be done more concisely). 
- If scale_to_effective_capabilities=True, both _daily_capabilities *and* _daily_capabilities_per_staff are rescaled by the "productivity" factor. This reflects the fact that we assume daily_capabilities are extended as a result of the same staff being more productive, not a higher staff count. 
- When the HealthSystemLogger is scheduled, this calculates and logs the current staff counts (rescaled by the capabilities_coefficient_parameter) in a way that is independent of the assumed productivity level.

tbhallett:
Thanks so much @marghe-molaro --- logic and implementation seem right to me.

I am 'Approving' this (but can't, because I opened the PR).

marghe-molaro (approved):

PR #1485, opened 2024-10-09
Climate change and health - precipitation and healthcare provision

RachelMurray-Watson:
Branch that investigates the effect of changes in precipitation patterns - caused by climate change - on the provision of healthcare services. 

Steps: 

- [x] Using SHAP files, grid Malawi into 1degree-by-1degree squares 
- [x] Associate facilities reported in ResourceFile_Master_Facilities_List with a grid (which will be used to provide location semi-specific climate projections)
- [ ] Download projected weather patterns from each available model for 2015 - 2100 (daily: historical data can be monthly). Initial plan: three SSPs (SSP1 - 1.9, SSP2-4.5 (current path), and SSP5 - 8.5) https://cds.climate.copernicus.eu/datasets/projections-cmip6?tab=download

          - [X] ssp1-1.9
          - [X] ssp2-4.5
          - [ ] ssp5-8.5

- [x] Aggregate daily precipitation across all models (median is usual metric)

- [x] Download historical monthly precipitation data for Malawi from https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-monthly-means?tab=overview (probably 2011 - 2024, as this period has good reporting from healthcare facilities)
- [x] Download historical reporting patterns of facilities from DHIS2 (initially just central facilities)
- [x] Associate reporting facilities with grid on map 
- [x] Association between historical precipitation events and reporting of facilities 

Issues so far... 

1. ~No specific data for location of facilities.~
                   - ~Could use the largest intersecting area, but for some the intersections could be evenly divided between >2 grid squares.~
                - ~Could randomly assign a grid square that is known to overlap with the region. WILL DO THIS but then leads to issue of small overlaps are as likely to lead to grid square being chosen as larger ones.~ Now DO have locations for most facilities, but may still need random assignation if no direct match can be made.
     

Very weak association between historical precipitation and reporting... Could improve by curtailing data but loose many data points. 

HOWEVER, have not accounted for the fact that there can be serious knock-on events. Precipitation may temporarily remove access, but also destroy buildings (see https://www.manaonline.gov.mw/index.php/the-star/item/4099-phalombe-health-centre-re-opened-for-operation). In DHIS2, these are returning "0" for % reporting, so even though there may be no rain, they are still not operating as usual.
################# 
So, instead of looking at total monthly rainfall, could instead look at daily maximums... ERA5 reanalysis data is now available (published 14/10/2024). 

- [x] Download daily maximum precipitation from 2014 - 2024 - _in progress, some issues with server_
- [x] Re-examine relationship between historical reporting and rainfall
- [x] Download smaller-level facility data
     - have done (clinics, district/rural/central hospitals, health centres). 90 have no expectation of reporting
- [x] Associate smaller facilities with grid squares based on new file

Issues part 2: 

1. Can get reporting data for 760 facilities, but only location <400 on a grid. Still may be fine, that's a lot of data

################# 
Reporting metric is slightly obtuse - don't have a good sense of e.g. what a drop in reporting means for a facility level. Excellent suggestion to look at ANC data, as is more interpretable and perhaps more stable than other metrics. 

- 

- [x]  Download HMIS Total Antenatal Visits from DHIS2
- [x] Run regression again with ANC and daily maximum (and/or monthly data)
- [x] Start projecting forward - perhaps days with projected disruptions?

################## 21.11.24

- [x] Based on feedback from Sally, re-evaluating ERA5 data for historical trend. Was in part based on [this](http://www.indiaenvironmentportal.org.in/files/file/climate_change_rainfall_madagascar.pdf) paper, which showed good agreement for the Mozambique/Malawi data and more "observational" datasets (occasionally agreeing better than with the observational with themselves). However, should do own more specific checks. 
- [ ] Build ML model of association 
- [x] Download DHIS2 data for other service provisions (e.g. HIV care?)
- [x] KD-Tree and Ball algorithm - potential means of avoiding regridding etc. CMIP6 data. Instead of assigning each facility to a grid point, go directly to projections and make "individual ensembles" for each facility? 
- [x] Mask COVID19 months
- [x] Add distance to closest clinic

Downloaded CHIRPS and TAMSAT data. There is no very strong agreement between the three (some outliers for all), but ERA5 and CHIRPS have better agreement than TAMSAT and the others. 

################## December 2024

- [x] Download downscaled CMIP6 data from https://planetarycomputer.microsoft.com/dataset/cil-gdpcir-cc0#Ensemble-example
- [x] Find min, median, and max model for all time points across all grid points
- [x] Calculate indices of interest (5 day monthly max and monthly cumulative)
- [x] Associate facilities with gridsquares using KDBall algorithm 
- [x] Use these as new bases for CMIP6 projections
- [x] Build demographic model in Excel? Or way of using TLO model - e.g. could get an estimate of the births from the Longterm projections model? 
- [x] Estimate number of Antenatal disruptions from the models (how does this translate?  Is one entry one appointment? How does ANC4 relate?)


RachelMurray-Watson:
> Quick thought on issue of individuals completely defaulting from care if HSI is cancelled due to climate disruption in the case where disease modules do not already account for possibility of individuals defaulting/re-entering care:
> 
> a simple/centralised way to address this could be to schedule a dummy symptom onset after a given delay at the time when the HSI is cancelled, so that the individual can be picked up again by the HealthSeekingBehaviourPoll. The dummy symptom would ensure that this doesn't interfere with the standard disease progression, i.e. if the individual no longer has real symptoms at the time of healthcare seeking the triaging event would simply not refer them for any actual appointment

I wonder though how this interacts with the “disruption” part of it. I think we assumed that the appointment deficit was indicative of absolute cancellation, rather than delay, as delays would have “appeared” in the data for another month. Unless this missed behavior was also somehow probabilistic?

PR #1486, opened 2024-10-10
Remove hardcoded reference to simulation years preventing simulations being ran with a start date after 2010

joehcollins:
This PR edits two modules which have hard coded reference to years in the simulation prevent running of a simulation with a start date after a given year. Therefore running the simulation with a start date later that these years causes the model to crash. This is mainly relevant to attempts to use the model in a cohort mode. 

@tdm32 have requested your review as its two of your modules. please let me know if you'd like me to change what i've got here.

### DIFF
--- a/src/tlo/methods/epi.py
+++ b/src/tlo/methods/epi.py
@@ -182,7 +182,10 @@ class Epi(Module):
         sim.schedule_event(EpiLoggingEvent(self), sim.date + DateOffset(years=1))
 
         # HPV vaccine given from 2018 onwards
-        sim.schedule_event(HpvScheduleEvent(self), Date(2018, 1, 1))
+        if self.sim.date.year < 2018:
+            sim.schedule_event(HpvScheduleEvent(self), Date(2018, 1, 1))
+        else:
+            sim.schedule_event(HpvScheduleEvent(self), Date(self.sim.date.year, 1, 1))
 
         # Look up item codes for consumables
         self.get_item_codes()
--- a/src/tlo/methods/hiv.py
+++ b/src/tlo/methods/hiv.py
@@ -1503,20 +1503,20 @@ class Hiv(Module, GenericFirstAppointmentsMixin):
 
         df = self.sim.population.props
 
-        # get number of tests performed in last time period
-        if self.sim.date.year == (self.sim.start_date.year + 1):
-            number_tests_new = df.hv_number_tests.sum()
+        if not self.stored_test_numbers:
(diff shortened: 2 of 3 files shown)


src/tlo/methods/hiv.py
```diff
@@ -1504,7 +1504,7 @@ def per_capita_testing_rate(self):
         df = self.sim.population.props
 
         # get number of tests performed in last time period
-        if self.sim.date.year == 2011:
+        if self.sim.date.year == (self.sim.start_date.year + 1):
```
tdm32, line 1507: per_capita_testing_rate is called by HivLoggingEvent, which first occurs 01/01/2011, i.e. one year after sim start. The function uses the previous years number of tests performed to calculate the number of tests performed in this time period. So for the first year, the value of previous_test_numbers must equal 0. 

I think a workaround could be:

# Check if it's the first year of the simulation and no results yet stored in self.stored_test_numbers

if not self.stored_test_numbers:
    # If it's the first year, set previous_test_numbers to 0
    previous_test_numbers = 0
else:
    # For subsequent years, retrieve the last stored number
    previous_test_numbers = self.stored_test_numbers[-1]

# Calculate number of tests now performed - cumulative, include those who have died
number_tests_new = df.hv_number_tests.sum()

# Store the number of tests performed in this year for future reference
self.stored_test_numbers.append(number_tests_new)

# Number of tests performed in the last time period
number_tests_in_last_period = number_tests_new - previous_test_numbers

Please note - I haven't tested this!

src/tlo/methods/epi.py
```diff
@@ -182,7 +182,10 @@ def initialise_simulation(self, sim):
         sim.schedule_event(EpiLoggingEvent(self), sim.date + DateOffset(years=1))
 
         # HPV vaccine given from 2018 onwards
-        sim.schedule_event(HpvScheduleEvent(self), Date(2018, 1, 1))
+        if self.sim.date.year < 2018:
```
tdm32, line 185: This is fine

src/tlo/methods/hiv.py
joehcollins: ok no probs - let me try that

matt-graham (approved):
These look like sensible changes to me and as @tdm32 has approved and tests are passing I would say this is good to merge unless @tbhallett wants to do a final review first?

joehcollins:
@tbhallett are you happy for this PR to go in, sorry only just noticed @matt-graham's review

Issue #1260, opened 2024-01-22
Check consistency of population-wide HSI events assumptions


marghe-molaro:
Currently in mode 2 we do not enforce resource constraints on population-wide HSI events as no modules in the simulation at the time of mode 2 development employed population-wide HSI events. The current healthcare system set up however still allows for this possibility, meaning developers may introduce population-wide HSI events in the future. For consistency, we should either 1. remove population-wide HSI events completely from the simulation, or 2. Allow for the possibility of population-wide HSI events to be resource constrained (both by HR capabilities and equipment). 

@tbhallett @EvaJanouskova

tbhallett:
Good point. It would make life most easy if we got rid of population wide HSI.

EvaJanouskova:
Thank you, @marghe-molaro, this answers my question well. @tbhallett, does this mean that for now I can only check the essential equipment availability for individual-level HSI events and if it's not available prevent the event from running; and leave the checking of equip availability for population-wide HSI events to be addressed as part of this issue?

tbhallett:
> Thank you, @marghe-molaro, this answers my question well. @tbhallett, does this mean that for now I can only check the essential equipment availability for individual-level HSI events and if it's not available prevent the event from running; and leave the checking of equip availability for population-wide HSI events to be addressed as part of this issue?

Yes, I think so.

PR #1493, opened 2024-10-18
Remove population-wide HSI event in ChronicSyndrome module

marghe-molaro:
As raised in Issue #1260, our assumptions around how the healthsystem handles resource constraints does not allow for the use of Population-wide HSI events. Here we remove one such instance from the mock ChronicSyndrome module used by some of the HealthSystem tests.


src/tlo/methods/chronicsyndrome.py
```diff
@@ -214,11 +214,11 @@ def initialise_simulation(self, sim):
         self.sim.schedule_event(outreach_event, self.sim.date + DateOffset(months=6))
 
         # Schedule the occurance of a population wide change in risk that goes through the health system:
```
matt-graham, line 216:
```suggestion
        # Schedule the occurrence of a population wide change in risk:
```

Possibly remove the reference to health system in comment as no longer a HSI event?

matt-graham (commented):
With regards to #1260 - would the plan be to remove the logic in the health system scheduler for processing population scoped HSI events in a separate PR? It might also be good to add some logic to `HSI_Event` or `PopulationScopeEventMixin` in such a PR that raises an error when trying to subclass from both if so to ensure we fail loudly on trying to instantiate a population-scoped HSI event.

tbhallett:
> With regards to #1260 - would the plan be to remove the logic in the health system scheduler for processing population scoped HSI events in a separate PR? It might also be good to add some logic to `HSI_Event` or `PopulationScopeEventMixin` in such a PR that raises an error when trying to subclass from both if so to ensure we fail loudly on trying to instantiate a population-scoped HSI event.

Yes I think that's a good idea. @marghe-molaro do you want to do that, or shall I?

marghe-molaro:
Hi @tbhallett and @matt-graham, 
The reason I didn't include that 'loud error' is that I'm a bit concerned that this might lead to failure during runtime if there are rare population-wide HSI events which are missed by tests. I think the safer option would be to remove the option for a pop-wide HSI completely, to make sure they cannot be used anywhere.

What do you think?

tbhallett:
> Hi @tbhallett and @matt-graham, The reason I didn't include that 'loud error' is that I'm a bit concerned that this might lead to failure during runtime if there are rare population-wide HSI events which are missed by tests. I think the safer option would be to remove the option for a pop-wide HSI completely, to make sure they cannot be used anywhere.
> 
> What do you think?

I reckon we should take the plunge now (and in this PR) to remove the possibility of "Population-wide" HSI in the HealthSystem code entirely. (It clutters up the code, and going-forwards, it complicates things, and we don't want anyone to implement a new one!). We'll do tests to see if there is any other other population-HSI lurking. As @marghe-molaro proposes changes to accommodate the results of her emulation models etc, it seems a good time to make this change and set a cleaner baseline.

tbhallett:
> Just adding a note here that actually the original motivation for removing Pop-wide HSI events is that it would be difficult to enforce HCW time resource constraints on them. Working on it now!

Yes, that's quite right.

tbhallett:
Closing this as this is now done, together with wider changes to remove Population level HSI in #1557

PR #1494, opened 2024-10-18
Reduce quantity of requested consumables in RTI module

joehcollins:
This PR reduces the number of units of consumables requested in the RTI module which were incorrectly (by me!) coded with the wrong units (i.e. per mg/ml instead of per vial etc).

### DIFF
--- a/src/tlo/methods/rti.py
+++ b/src/tlo/methods/rti.py
@@ -2362,7 +2362,7 @@ class RTI(Module, GenericFirstAppointmentsMixin):
             # request a general anaesthetic
             get_item_codes("Halothane (fluothane)_250ml_CMST"): 100,
             # clean the site of the surgery
-            get_item_codes("Chlorhexidine 1.5% solution_5_CMST"): 100,
+            get_item_codes("Chlorhexidine 1.5% solution_5_CMST"): 600,
             # tools to begin surgery
             get_item_codes("Scalpel blade size 22 (individually wrapped)_100_CMST"): 1,
             # administer an IV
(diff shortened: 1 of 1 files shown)


src/tlo/methods/rti.py
```diff
@@ -2386,7 +2386,7 @@ def look_up_consumable_item_codes(self):
             # request a local anaesthetic
             get_item_codes("Halothane (fluothane)_250ml_CMST"): 100,
             # clean the site of the surgery
-            get_item_codes("Chlorhexidine 1.5% solution_5_CMST"): 500,
+            get_item_codes("Chlorhexidine 1.5% solution_5_CMST"): 100,
```
sakshimohan, line 2389: Hi @joehcollins Just to double check, this means that 100ml of `Chlorhexidine 1.5% solution_5_CMST` is needed for both minor and major surgery?

joehcollins: Hi @sakshimohan - i dont have a good sense of how much would be needed for either procedure to be honest so seemed sensible to just have them both the same. This 100ml may still be too high...

sakshimohan: I wonder if this reference might help. [UNICEF's Obstetric,surgical kit,suppl.1-drugs ](https://supply.unicef.org/s9908300.html) consists of 30 units of Chlorhexidine conc. sol. 5%/BOT-1000 ml for "average of 50 deliveries with complications, including caesarean sections, laparotomy and other minor surgery in emergency". This would imply 30 X 1000ml (5%) /50 per surgery on average ( = **600ml of 5% solution /surgery**).

joehcollins: That sounds logical to me. I can update the major surgery to 600ml based on this. What about for the minor surgeries? leave as less?

sakshimohan: I'll have to leave that decision to you, sorry! 😅 Either will be acceptable for the overall costing calibration with the updates we're making I think, given that we're reducing the unit cost of Chlorhexidine.

joehcollins: ok i've put 600 for major and 300 for minor!

sakshimohan (approved):
Thanks, Joe. All looks good to me.

PR #1503, opened 2024-11-04
Ensure `bitset_handler` tests run in GitHub Actions workflow and remove usages of `BytesDType`

### LINKED ISSUE #1501: Tests in `tests/bitset_handler` not currently being run in GitHub Actions workflow
The tests in the modules in `tests/bitset_handler` added in #1448 are not currently being run in the tests GitHub Action workflow as that is currently set up as a matrix job over the test files found by the following logic

https://github.com/UCL/TLOmodel/blob/8f22e78386cc5b34dc6c4876587b4a6e03a64b6a/.github/workflows/tests.yml#L48-L56

that is all files matching the pattern `tests/test_*.py`. As the files in `tests/bitset_handler` do not match this pattern they are not picked up and so the tests not run.

### LINKED ISSUE #1502: Tests in `tests/bitset_handler` fail with NumPy v1.24.4
@mnjowe found that the test modules in `tests/bitset_handler` added in #1448 fail when run with NumPy v1.24.4 (which is what we currently specify our requirements) with an error

```
tests/bitset_handler/conftest.py:14: in <module>
    from numpy.dtypes import BytesDType  # pylint: disable=E0611
ModuleNotFoundError: No module named 'numpy.dtypes'
```

but do run correctly in a local environment with NumPy v1.26.4 installed. 

This is currently not being picked up by tests workflow because of #1501.

We need to either remove the offending import (should be possibly as I think it's just used for type hints) or change our specified NumPy minimum version.

matt-graham:
Resolves #1501 and resolves #1502.

Avoids import of `BytesDType` from `numpy.dtypes` (which is not available in `numpy==1.24.4` which is what we currently pin at) and also updates construction of matrix of test files in tests workflow to also include test modules under subdirectories in `tests`. Also makes a couple of regex strings raw string literals to avoid warnings about unrecognised character escapes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,9 +46,10 @@ jobs:
         name: Set matrix
         run: |
           set -e
+          shopt -s globstar
           # Find all test files and generate their list in JSON format
           VAR_FILES="{\"include\":["
-          for file in tests/test_*.py; do
+          for file in tests/**/test_*.py; do
               VAR_FILES="${VAR_FILES}{\"file\":\"${file}\"},"
           done
           VAR_FILES="${VAR_FILES}]}"
--- a/src/tlo/bitset_handler/bitset_extension.py
+++ b/src/tlo/bitset_handler/bitset_extension.py
@@ -19,7 +19,6 @@ from typing import (
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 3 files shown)


tamuri (approved):

mnjowe:
Thanks @matt-graham. I think with your changes I should be okay reverting to Numpy v1.24.4 then.

PR #1508, opened 2024-11-07
Convert excel files to csv files. Joe's modules

mnjowe:
Converting all excel files to CSV files. Joe's modules


joehcollins:
@mnjowe - i've just deleted some of the old parameter files which are no longer needed and pushed the changes. (sorry didnt mean to use such an uninformative comit name).

Additionally- does this same process need to happen for `pregnancy_supervisor.py`?

joehcollins:
@EvaJanouskova - have requested your review of this as the updates for the contraception parameter files are also in this branch.

mnjowe:
> @mnjowe - i've just deleted some of the old parameter files which are no longer needed and pushed the changes. (sorry didnt mean to use such an uninformative comit name).
> 
> Additionally- does this same process need to happen for `pregnancy_supervisor.py`?

Absolutely right Joe. That Excel file needs to be converted using same method. Can you also remind me any of your modules you feel are using excel files but not included in this PR?

joehcollins (approved):
Great thanks @mnjowe - this looks all good to me!

EvaJanouskova:
@timcolbourn, I wonder, if we won't loose some useful information this way. Because in csv files you don't see where the numbers are coming from, shouldn't the csv files also contain the information on what is the origin of the numbers?

For example, values in the column `r_init1_age` (B2-B36) in the `Initiation_ByAge.csv` is based on the `proportion relative change compared to average` values (E8-E42) from `DELETE_Initiation1_by_age.csv`. Should be such information stored in the `Initiation_ByAge.csv`, or?

If so, I would also suggest to update the names of the files with calculations to smt like CALC_ instead of DELETE_.

EvaJanouskova:
> @timcolbourn, I wonder, if we won't loose some useful information this way. Because in csv files you don't see where the numbers are coming from, shouldn't the csv files also contain the information on what is the origin of the numbers?
> 
> For example, values in the column `r_init1_age` (B2-B36) in the `Initiation_ByAge.csv` is based on the `proportion relative change compared to average` values (E8-E42) from `DELETE_Initiation1_by_age.csv`. Should be such information stored in the `Initiation_ByAge.csv`, or?
> 
> If so, I would also suggest to update the names of the files with calculations to smt like CALC_ instead of DELETE_.

@timcolbourn, I went through all the CSVs (converted from the `ResourceFile_Contraception` sheets) and beside the sheets starting with DELETE_, or DEL_, we also don't use the `contraception_failure_discont` sheet. All these CSVs can be deleted. And if you like to have backup of the calculations of the data included in the CSVs we are using, the original spreadsheet could be saved on OneDrive, and that's it. What do you think?

@mnjowe, Are you happy to do all of that (delete the CSVs, and save the original `ResourceFile_Contraception` spreadsheet on OneDrive)? Or do you need me to do any of that?

tamuri:
@mnjowe - please action Eva's comments.

Once that's done, please request review from @EvaJanouskova - she need to approve before this can be merged.

mnjowe:
original contraception excel file moved to sharepoint [here](https://imperiallondon.sharepoint.com/:x:/r/sites/tlomodel-WP/Shared%20Documents/08%20-%20Archive/Excel%20files%20converted%20to%20CSV/[ResourceFile_Contraception.xlsx](https://imperiallondon.sharepoint.com/:x:/r/sites/tlomodel-WP/Shared%20Documents/08%20-%20Archive/Excel%20files%20converted%20to%20CSV/ResourceFile_Contraception.xlsx?d=w700fadc1715543f68f45f05598e7b780&csf=1&web=1&e=XvveUr)?d=w700fadc1715543f68f45f05598e7b780&csf=1&web=1&e=XvveUr)

EvaJanouskova (approved):
Seems good to me @mnjowe. Thanks!

PR #1510, opened 2024-11-11
Log Consumables That are Assumed to Be Used

### LINKED ISSUE #1505: Consumables consumption double-counting
Are we over-counting use of consumables in the following scenario (common in contraception module)....

1- HSI runs, requests consumables A and B. A is available, B is not. Consumption of B is logged.
2- HSI not successful so is rescheduled.
3- HSI runs again and request consumables A and B. Consumption of A and B is logged.

Thus, we have counted use of B twice, when in fact the delivery of the service requires only one item of A and one item of B.

We could overcome this by making the initial check without there being any logging, and only logging usage if all (essential) available (when the service is actually provided).

(cc @sakshimohan )

tdm32:
Fixes #1505

PR creates a new entry in the consumables logger which records only those consumables used. This uses the logic that if all requested item_codes are available, then the HSI will run and the items will be used. The availability check only requires essential item_codes, any available optional_item_codes are logged only if the essential item_codes are all available.

### DIFF
--- a/src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
+++ b/src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
@@ -34,8 +34,8 @@ resourcefilepath = Path("./resources")
 
 # %% Run the simulation
 start_date = Date(2010, 1, 1)
-end_date = Date(2015, 1, 1)
-popsize = 2000
+end_date = Date(2022, 1, 1)
+popsize = 5000
 
 # scenario = 1
 
@@ -46,12 +46,12 @@ log_config = {
     "custom_levels": {
         "*": logging.WARNING,
         # "tlo.methods.deviance_measure": logging.INFO,
-        "tlo.methods.epi": logging.INFO,
+        # "tlo.methods.epi": logging.INFO,
         "tlo.methods.hiv": logging.INFO,
         "tlo.methods.tb": logging.INFO,
         "tlo.methods.demography": logging.INFO,
         # "tlo.methods.demography.detail": logging.WARNING,
-        "tlo.methods.healthsystem.summary": logging.INFO,
+        # "tlo.methods.healthsystem.summary": logging.INFO,
         # "tlo.methods.healthsystem": logging.INFO,
         # "tlo.methods.healthburden": logging.INFO,
     },
--- a/src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
+++ b/src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
@@ -34,8 +34,8 @@ resourcefilepath = Path("./resources")
 
 # %% Run the simulation
 start_date = Date(2010, 1, 1)
-end_date = Date(2022, 1, 1)
-popsize = 5000
+end_date = Date(2015, 1, 1)
+popsize = 2000
 
 # scenario = 1
 
@@ -46,12 +46,12 @@ log_config = {
     "custom_levels": {
         "*": logging.WARNING,
         # "tlo.methods.deviance_measure": logging.INFO,
-        # "tlo.methods.epi": logging.INFO,
+        "tlo.methods.epi": logging.INFO,
         "tlo.methods.hiv": logging.INFO,
         "tlo.methods.tb": logging.INFO,
         "tlo.methods.demography": logging.INFO,
         # "tlo.methods.demography.detail": logging.WARNING,
-        # "tlo.methods.healthsystem.summary": logging.INFO,
+        "tlo.methods.healthsystem.summary": logging.INFO,
         # "tlo.methods.healthsystem": logging.INFO,
         # "tlo.methods.healthburden": logging.INFO,
     },
--- a/src/tlo/methods/consumables.py
+++ b/src/tlo/methods/consumables.py
@@ -240,7 +240,7 @@ class Consumables:
                     'TREATMENT_ID': treatment_id or "",
                     'Item_Available': str(items_available),
                     'Item_NotAvailable': str(items_not_available),
-                    'Items_Used': str(items_used),
+                    'Item_Used': str(items_used),
                 },
                 description="Record of requested and used consumable items."
             )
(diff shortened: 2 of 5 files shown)


src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
tbhallett, line 1: roll back this file!

src/tlo/methods/consumables.py
```diff
@@ -208,35 +209,44 @@ def _request_consumables(self,
...
+                key='Consumables',
+                data={
+                    'TREATMENT_ID': treatment_id or "",
+                    'Item_Available': str(items_available),
+                    'Item_NotAvailable': str(items_not_available),
+                    'Items_Used': str(items_used),
+                },
+                description="Record of requested and used consumable items."
+            )
+            self._summary_counter.record_availability(items_available, items_not_available, items_used)
```
tbhallett, line 249:
```suggestion
            self._summary_counter.record_availability(
            items_available=items_available, 
            items_not_available=items_not_available, 
            items_used=items_used,
            )
```

src/tlo/methods/consumables.py
```diff
@@ -208,35 +209,44 @@ def _request_consumables(self,
...
-            self._summary_counter.record_availability(items_available=items_available,
-                                                      items_not_available=items_not_available)
+            items_available = {k: v for k, v in _all_item_codes.items() if available[k]}
+            items_not_available = {k: v for k, v in _all_item_codes.items() if not available[k]}
+
+            # Log items used if all essential items are available
+            items_used = {
+                **{k: v for k, v in item_codes.items() if available[k]},
+                **{k: v for k, v in optional_item_codes.items() if available[k]}
+            } if all(available.get(k, False) for k in item_codes) else {}
```
tbhallett, line 237:
```suggestion
            items_used = items_available if all(available.get(k, False) for k in item_codes) else {}
```

src/tlo/methods/consumables.py
```diff
@@ -200,6 +200,7 @@ def _determine_default_return_value(cons_availability, default_return_value):
     def _request_consumables(self,
                              facility_info: 'FacilityInfo',  # noqa: F821
                              item_codes: dict,
```
tbhallett, line 202: to be explicit shall we rename this kwarg "essential_item_codes"

src/tlo/methods/consumables.py
```diff
@@ -200,6 +200,7 @@ def _determine_default_return_value(cons_availability, default_return_value):
...
+                             optional_item_codes: Optional[dict],
```
tbhallett, line 203: the type hint says Optional but no default is provided. I think it's good to make it optional though, and this avoids then having to make changes to the test.

```suggestion
                             optional_item_codes: Optional[dict] = None,
```

... and then inside the function

```optional_item_codes = dict() if optional_item_codes is None```

File: tests/test_consumables.py
Comment (tbhallett) [inline tests/test_consumables.py:1]: see comment in Consumables.py: if we have default for kwarg `optional_item_codes` we could avoid these changes.

**Review by tbhallett — CHANGES_REQUESTED** [review]
Nicely done @tdm32 --- this looks like it's doing the job to me!

See some small changes and one thing on the core part of the logic.

Have you done any trial runs / test to make sure it's all behaving?

Should we add a test formally?

File: src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py
Comment (tdm32) [reply src/scripts/hiv/projections_jan2023/analysis_logged_deviance.py:1]: done

File: src/tlo/methods/consumables.py (hunk 3 above)
Comment (tdm32) [reply src/tlo/methods/consumables.py:202]: done

File: tests/test_consumables.py
Comment (tdm32) [reply tests/test_consumables.py:1]: great, I've removed the optional_item_codes from the tests and updated the kwarg to essential_item_codes to make it explicit which items we include in this

**Comment by tdm32** [comment]
> Nicely done @tdm32 --- this looks like it's doing the job to me!
> 
> See some small changes and one thing on the core part of the logic.
> 
> Have you done any trial runs / test to make sure it's all behaving?
> 
> Should we add a test formally?

Thanks @tbhallett. I've made the requested changes and added a test

File: src/tlo/methods/consumables.py [hunk 5]
```diff
@@ -208,35 +209,45 @@ def _request_consumables(self,
...
-                        )
-
-            self._summary_counter.record_availability(items_available=items_available,
-                                                      items_not_available=items_not_available)
+            items_available = {k: v for k, v in _all_item_codes.items() if available[k]}
+            items_not_available = {k: v for k, v in _all_item_codes.items() if not available[k]}
+
+            # Log items used if all essential items are available
+            items_used = items_available if all(available.get(k, False) for k in item_codes) else {}
+
```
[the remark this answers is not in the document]
tdm32: There's now also a test which has zero availability for an optional item, but the essential items are still logged as 'Used' so I think it's working...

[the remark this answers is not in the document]
joehcollins: Ok great!! My mistake - i read this as only used if all were available

tbhallett:
Thanks very much @tdm32  -- all looks really good to me.

I see one failing test, but that looks like a simple one to fix: just add the name the new logging key in `test_two_loggers_in_healthsystem`.

EvaJanouskova:
@tdm32, And also add assert similar to what is for Item_Available and Item_NotAvailable, see below the comment line in the test_two_loggers_in_healthsystem:
#  - Consumables (total over entire period of log that are available / not available)  # add _Item_

EvaJanouskova:
It seems good, that is what I imagined as the simpler version, which is nice as ntg has to be changed for the consumables stock-outs graphs. My only concern was that it may add more lines in logging, than the solution I was suggesting. But I didn't test how much difference it would do, hence I think this smooth solution is fine. Good work @tdm32!

tdm32:
> Thanks very much @tdm32 -- all looks really good to me.
> 
> I see one failing test, but that looks like a simple one to fix: just add the name the new logging key in `test_two_loggers_in_healthsystem`.

This is fixed now

tdm32:
> @tdm32, And also add assert similar to what is for Item_Available and Item_NotAvailable, see below the comment line in the test_two_loggers_in_healthsystem: `# - Consumables (total over entire period of log that are available / not available) # add _Item_`

This is done now

tbhallett (approved):

PR #1511, opened 2024-11-11
Calibrate Use of Anti-Epilepsy Medicines

tbhallett:


joehcollins:
@tbhallett  - just to say the plots after change 2 dont appear to be visable!

src/tlo/methods/epilepsy.py
```diff
@@ -406,8 +411,14 @@ def do_at_generic_first_appt_emergency(
...
-            event = HSI_Epilepsy_Start_Anti_Epileptic(person_id=person_id, module=self)
-            schedule_hsi_event(event, priority=0, topen=self.sim.date)
+            # Determine if treatment will start - depends on probability of prescribing, which is calibrated to
+            # induce the right proportion of persons with epilepsy receiving treatment.
+
+            prob_start = self.parameters['prob_start_anti_epilep_when_seizures_detected_in_generic_first_appt']
+
+            if self.rng.random_sample() < prob_start:
+                event = HSI_Epilepsy_Start_Anti_Epileptic(person_id=person_id, module=self)
+                schedule_hsi_event(event, priority=0, topen=self.sim.date)
```
joehcollins, line 422: Wondering if, after these changes, cons use is still too high that we could limit the number of times an individual can return to this HSI to start treatment - currently unlimited?

tbhallett: It's a good idea, Joe -- I'll add that in.

I don't think this is a major cause of over-demand as the logic selects the 'best available' of the meds, and only reschedules if non-available. Nevertheless, it's a good thing to do.

tbhallett:
> @tbhallett - just to say the plots after change 2 dont appear to be visable!

My mis

> @tbhallett - just to say the plots after change 2 dont appear to be visable!

my mistake. fixing now.

tbhallett:
fyi @tdm32  and @sakshimohan --- this is the other PR we'll need for the re-runs

src/tlo/methods/epilepsy.py
```diff
@@ -639,8 +662,12 @@ def apply(self, person_id, squeeze_factor):
                 priority=0
             )
 
-        else:
+        elif (
+            self._counter_of_failed_attempts_due_to_unavailable_medicines
+            < self._MAX_NUMBER_OF_FAILED_ATTEMPTS_BEFORE_DEFAULTING
+        ):
```
joehcollins, line 668: (potentially highlighting gaps in my python knowledge) I would have assumed that at every instance of this HSI that self._counter_of... is set at 0 when __init__ runs so following this logic self._counter_of... would never exceed 1? Or is it the case that __init__ is only ran once? If thats the case would the counter not need to be reset to 0 incase the individual, who gives up on seeking care, develops different symptoms again and wants to seek care?

tbhallett: yes, the __init__ is only called when the instance is initiated. 

When it's rescheduled, it's not created again, so __init__ is not called a second time.

joehcollins: Ok perfect!

joehcollins (approved):

tdm32 (approved):

Issue #1509, opened 2024-11-11
Update read cvs files method in util


mnjowe:
- [ ] Update read CSV method in util to be able to return a dictionary when a single file is read and `files` argument is None. This is the expected behaviour in pd.read_excel when one Excel file is read and `sheet_names` argument is `None`. The output should be in the form
`{ sheet_name: DataFrame }`.

- [ ] add another argument for dtypes
- [ ] update read csv files tests

PR #1512, opened 2024-11-12
Updates to read CSV method

mnjowe:
This PR aims to address issue #1509


src/tlo/util.py
```diff
@@ -474,7 +475,7 @@ def convert_excel_files_to_csv(folder: Path, files: Optional[list[str]] = None,
             Path(folder/excel_file_path).unlink()
 
 
-def read_csv_files(folder: Path, files: Optional[list[str]] = None) -> DataFrame | dict[str, DataFrame]:
+def read_csv_files(folder: Path, dtype: DtypeArg | None = None, files: Optional[list[str]] | None | int = 0) -> DataFrame | dict[str, DataFrame]:
```
matt-graham, line 478:
```suggestion
def read_csv_files(folder: Path, dtype: DtypeArg | None = None, files: list[str] | None | int = 0) -> DataFrame | dict[str, DataFrame]:
```
`Optional[T]` is equivalent to `T | None` so is redundant here.

src/tlo/util.py
```diff
@@ -498,15 +500,15 @@ def clean_dataframe(dataframes_dict: dict[str, DataFrame]) -> None:
         for _key, dataframe in dataframes_dict.items():
             all_data[_key] = dataframe.drop(dataframe.filter(like='Unnamed'), axis=1)  # filter and drop Unnamed columns
 
-    if files is None:
+    if files == 0 or files is None:
```
matt-graham, line 503: Is the reasoning for allowing both `files = 0` and `files = None` to be used to select all files in directory so that different output behaviour can be selected? At the moment it looks like `files = 0` if there is a single file in directory will return just a single dataframe (but a dictionary otherwise), while `files = None` will always return a dict. Having this flexibility possibly makes sense, but if this is the intention I think it needs to be made more obvious by explaining in docstring. Alternatively might make more sense to instead have an explicit boolean flag `return_single_file_as_dataframe` to control output behaviour which if `True` and `len(all_data) == 1` will force a single dataframe to be returned, with a dict returned otherwise, as the name would make it explicit to the user what the behaviour is.

src/tlo/util.py
```diff
@@ -484,6 +485,7 @@ def read_csv_files(folder: Path, files: Optional[list[str]] = None) -> DataFrame
     :py:func:`pandas.drop`.
 
     :param folder: Path to folder containing CSV files to read.
+    :param dtype: preferred datatype
```
matt-graham, line 488: The `pandas.read_csv` method allows passing in a dictionary of datatypes (mapping from column names to datatypes) in cases where you want not just a single datatype for whole sheet but different datatypes per column. Are there any cases where we might need this functionality? As we passthrough the `dtype` argument straight to `pandas.read_csv` we could already pass in a dict here, but it might be worth documenting this is an additional option if we do think it will be used.

tests/test_utils.py
```diff
@@ -332,8 +332,44 @@ def copy_files_to_temporal_directory_and_return_path(tmpdir):
     return tmpdir_resource_filepath
 
 
-def test_read_csv_method_with_no_file(tmpdir):
-    """ read csv method when no file name is supplied
+def test_pass_datatypes_to_read_csv_method(tmpdir):
+    """ test passing column datatypes to read csv method. Final column datatype should change to what has been passed """
+    # copy and get resource files path in the temporal directory
+    tmpdir_resource_filepath = copy_files_to_temporal_directory_and_return_path(tmpdir)
```
matt-graham, line 338: As we don't seem to use the files copied to the temporary directory in `copy_files_to_temporal_directory_and_return_path` in this test could we not just use `tmpdir` directly here?

tests/test_utils.py
```diff
@@ -332,8 +332,44 @@ def copy_files_to_temporal_directory_and_return_path(tmpdir):
...
+    sample_data = pd.DataFrame(data={'numbers1': [5,6,8,4,9,6], 'numbers2': [19,27,53,49,75,56]}, dtype=int)
+    sample_data.to_csv(tmpdir_resource_filepath/'sample_data.csv', index=False)
+    # read from the sample data file
+    read_sample_data = read_csv_files(tmpdir_resource_filepath, files=['sample_data'])
+    # confirm column datatype is what was assigned
+    assert read_sample_data.numbers1.dtype and read_sample_data.numbers2.dtype == 'int'
```
matt-graham, line 344:
```suggestion
    assert read_sample_data.numbers1.dtype == 'int' and read_sample_data.numbers2.dtype == 'int'
```
The current condition is equivalent to 

```
(read_sample_data.numbers1.dtype) and (read_sample_data.numbers2.dtype == 'int')
```

that is check if `read_sample_data.numbers1.dtype` evaluates to `True` (which I think will always be the case) and `read_sample_data.numbers2.dtype == 'int'`.

src/tlo/util.py
matt-graham: Actually it seems like you do use explicitly use this ability to pass `dict` in test below so this should definitely be documented and typehint for `dtype` updated to `DtypeArg | dict[str, DtypeArg] | None`.

matt-graham (changes requested):

mnjowe:
Thanks for the review @matt-graham . Applying changes now

src/tlo/util.py
mnjowe: @matt-graham , declaring an explicit flag `return_single_file_as_dataframe`  to control output is good but I'm afraid it will change the behaviour that modellers are used to in pd.read_excel. If we are to implement this now it will mean introducing this flag in several other files as well. Do we want to this in this PR?

src/tlo/util.py
```diff
@@ -498,15 +513,21 @@ def clean_dataframe(dataframes_dict: dict[str, DataFrame]) -> None:
...
+        return_dict = True
+    elif files is None:
+        files = [f_name.stem for f_name in folder.rglob("*.csv")]
+        return_dict = True
+    elif isinstance(files, str):
+        files = [files]
     else:
-        for f_name in files:
-            all_data[f_name] = pd.read_csv((folder / f_name).with_suffix(".csv"))
+        files = [f_name.stem for f_name in folder.rglob("*.csv")]
```
tamuri, line 525: Because of `rglob` here, if you have two files with the same name but in different sub-directories, you'll get one overwriting the other. Should we only read csv files in the given directory, and ignore subdirectories?

src/tlo/util.py
matt-graham: Hi @mnjowe, thanks for the explanation and for expanding the docstring. I think the behaviour with different values of `files` are now clearly defined to marking this as resolved

src/tlo/util.py
```diff
@@ -498,15 +513,21 @@ def clean_dataframe(dataframes_dict: dict[str, DataFrame]) -> None:
...
-
+    return_dict = False # a flag that will determine whether the output should be a dictionary or a DatFrame
+    if isinstance(files, list):
+        return_dict = True
+    elif files is None:
+        files = [f_name.stem for f_name in folder.glob("*.csv")]
+        return_dict = True
+    elif isinstance(files, str):
+        files = [files]
     else:
```
matt-graham, line 524: At the moment this `else` clause will be run whenever `files` is not a instance `list` or `str` and is not `None`. Looking at the type hint / default value I guess this is meant to cover the case when `files` is an instance of an `int`? If so it would be better to make this `elif isinstance(files, int):` and add a separate `else:` clause that raises an error saying `files` is not of an expected type if none of the previous clauses have run. At the moment for example if a user passes a non-list iterable data structure as the `files` argument - for example a tuple or generator (as returned for example by `Path.glob` and `Path.rglob`) then the value passed will be silently overwritten which is probably not what they would expect. Failing loudly if `files` is of an unexpected type (and/or allowing `files` to be of more general iterable types by for example checking if it is an instance of `Iterable` from `collections.abc` instead of `list`) would be better here.

matt-graham: You could also avoid the duplication of the list comprehension `[f_name.stem for f_name in folder.glob("*.csv")]` by having something like

```Python
elif isinstance(files, int) or files is None:
    files = [f_name.stem for f_name in folder.glob("*.csv")]
    return_dict = files is None
else:
    raise TypeError(f"Value passed for files argument {files} is not one of expected types.")
```

mnjowe: The thought of combining those two conditions crossed my mind but I was lacking the smart way of setting `return_dict` to either True or False. thanks for this @matt-graham

mnjowe: I believe using `isinstance(files, collections.abc.Iterable)` is not the best approach here, as it also considers strings to be iterable. If a string is included, it could alter the expected behavior of the return type when modelers specify a file name by passing a string. This would cause the method to return a dictionary instead of a DataFrame, which is likely not the intended outcome.

mnjowe:
Hi @matt-graham . If you're happy with this update, can you please get it merged so I can proceed converting the remaining excel files?

Thanks

src/tlo/util.py
matt-graham: Hi @mnjowe, good point around `str` being an instance of `Iterable`. We could sidestep that issue by having the `elif isinstance(files, str):` clause before the `elif isinstance(files, Iterable):` clause (as this would mean no string `files` reach the latter clause) but the current solution is fine here, the main thing was making sure we get an informative error message when we use a non-list iterable rather than the function doing something unexpected.

matt-graham (approved):

src/tlo/util.py
mnjowe: Ah great! thanks @matt-graham . will definitely consider adding this the next time I'm to updating this method

PR #1516, opened 2024-11-15
update_hrh_data_script

BinglingICL:
The old script cannot run now due to outdated/changed pandas functions, such as append and groupby, and the read_excel function that cannot read the auxiliary data of facility level establishment. Besides, the dropbox path needs to be replaced by the OneDrive path.

Code are now updated and can run now. 

To fix the issue of that auxiliary file cannot be read, I have added a csv file copying the relevant data from the original file to OneDrive.

Checked that the newly resource files are only different with master ones in the data format, not the data numerics.

I have created this pull request so that Marianca can run the script herself. Not necessarily needed to be merged to master for now.

As Mariana will examine the CHAM and MOH data separately, and in the future, we would need to format the coding in more general way to be applied to wider context (e.g. other countries in the region), I expect we would need to update this script more thoroughly.


BinglingICL:
> Thanks @BinglingICL --- Please could you double-check that Mariana has got this working? Then we can merge.
> 
> I see the .csv files are updated, but from the descriptions, I believe that they should not have changed actually, --- is that right?

Thanks for checking this branch. The values of resource files have only changed in the last one or two decimals (12th/13th decimals).

I can run this branch locally; will check with Mariana if she could get it work, too.

tbhallett (approved):

PR #1520, opened 2024-11-19
Reduce viral load testing for PLHIV on ART

tdm32:
Reduce the amount of viral load testing to occur every 2 years, consistent with data showing 400,000-450,000 VL tests occurring annually between 2020-2023.

This also includes the changes from PR #1456 which can be closed - increasing VMMC post-2019

image


tdm32:
@tbhallett This is ready for review

src/tlo/methods/hiv.py
```diff
@@ -2877,13 +2891,15 @@ def do_at_continuation(self, person_id):
...
         person = df.loc[person_id]
+        p = self.module.parameters
 
         # default to person stopping cotrimoxazole
         df.at[person_id, "hv_on_cotrimoxazole"] = False
 
         # Viral Load Monitoring
         # NB. This does not have a direct effect on outcomes for the person.
-        _ = self.get_consumables(item_codes=self.module.item_codes_for_consumables_required['vl_measurement'])
+        if self.module.rng.random_sample(size=1) < p['dispensation_period_months'] / p['interval_for_viral_load_measurement_months']:
```
tbhallett, line 2901: RHS of this equation is confusing me:
 `p['dispensation_period_months'] / p['interval_for_viral_load_measurement_months']` 

This should be a probability of needing a VL, I think.

tbhallett (changes requested):

src/tlo/methods/hiv.py
tdm32: @tbhallett  The probability of needing a VL measurement is the time between ART appts (3 months) divided by the interval for VL measurements which is 24 months, so the prob in this case is 0.25 per ART appt. If we change ART follow-ups to every 6 months, this should still work.

tbhallett: why not just have a parameter saying 'probability of having VL measurement during an ART appointment' then?

tdm32: in case the ART follow-up time is changed to 6-monthly, in which case the probability would have to be rescaled to match. This way we can change the frequency of ART follow-ups and the VL monitoring will still occur at the same frequency

tbhallett: okay, I see. And you want these durations to be 'averages' rather than deterministic.

tbhallett:
@tdm32  -- ok thanks, we can merge this then. With the PR from earlier, there's now a conflict with master though.....

tbhallett (approved):

tdm32:
> @tdm32 -- ok thanks, we can merge this then. With the PR from earlier, there's now a conflict with master though.....

The conflict is down to the move to csv files for the resource files. I've updated now and will keep an eye on the tests.

PR #1507, opened 2024-11-06
Generate publications list from BibTeX file

matt-graham:
Uses [`pybtex`](https://pybtex.org/) to generate a HTML publication list from a BibTeX file containing publication details for inclusion on website.

The `note` field in BibTeX file is used to store section names to organize publications under to reflect [current publications page](https://www.tlomodel.org/publications.html). Compared to the current publications list, this adds a few extra details (abbreviated author list, publication year and journal) - see example output below.

We could potentially make each publication entry expandable with additional details such as full author list, abstract and bibtex entry listed when expanded if that would be useful? @tbhallett tagging you for thoughts on what it would be useful to include!

<details>
<summary>Example output</summary>

<h2>Overview of the model</h2>
<ul>
<li><a href="https://doi.org/10.1101/2024.01.04.24300834">Hallett et al (2024). A <span class="bibtex-protected"><span class="bibtex-protected">Healthcare Service Delivery</span></span> and <span class="bibtex-protected"><span class="bibtex-protected">Epidemiological Model</span></span> for <span class="bibtex-protected"><span class="bibtex-protected">Investigating Resource Allocation</span></span> for <span class="bibtex-protected"><span class="bibtex-protected">Health</span></span>: <span class="bibtex-protected"><span class="bibtex-protected">The Thanzi La Onse Model</span></span>. <em>medRxiv</em>.</a></li>
</ul>

<h2>Analyses using the model</h2>
<ul>
<li><a href="https://doi.org/10.1016/S2214-109X(24)00259-6">Mangal et al (2024). Assessing the effect of health system resources on <span class="bibtex-protected"><span class="bibtex-protected">HIV</span></span> and tuberculosis programmes in <span class="bibtex-protected"><span class="bibtex-protected">Malawi</span></span>: a modelling study. <em>The Lancet Global Health</em>.</a></li>
[...]

PR #1523, opened 2024-11-21
Add Theory of Change Framework preprint

sakshimohan:
Added preprint to publication list.


sakshimohan:
@matt-graham @tbhallett I have added this publication to the Zotero library. But I haven't been succesful with step 3 of the process. 

When I run in the tlo environment
`tox -e update-publications
`
I get the following error - 
`ROOT: HandledError| provided environments not found in configuration file:
update-publications`

matt-graham:
Hi @sakshimohan. Are you running the `tox -e update-publications` command from the root the `TLOmodel` repository and with the most recent version of `master` checked out? The error you got suggests `tox` is not able to find the `update-publications` environment defined in the `tox.ini` file - this was added in #1507 so should be available. You can also try running
```
python /home/matt/projects/TLOmodel/docs/tlo_publications.py --update-from-zotero
```
to directly run the script, but you will also need to run `pip install pybtex requests` first in this case to install the additional dependencies the script needs (`tox` deals with this automatically).

sakshimohan:
> Hi @sakshimohan. Are you running the `tox -e update-publications` command from the root the `TLOmodel` repository and with the most recent version of `master` checked out? The error you got suggests `tox` is not able to find the `update-publications` environment defined in the `tox.ini` file - this was added in #1507 so should be available. You can also try running
> 
> ```
> python /home/matt/projects/TLOmodel/docs/tlo_publications.py --update-from-zotero
> ```
> 
> to directly run the script, but you will also need to run `pip install pybtex requests` first in this case to install the additional dependencies the script needs (`tox` deals with this automatically).

Many thanks, Matt. That problem is resolved. I get the following different error now - 
`  File "/Users/sm2511/PycharmProjects/TLOmodel/docs/tlo_publications.py", line 103
    toplevel[
    ^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?
`

sakshimohan:
> > Many thanks, @matt-graham . That problem is resolved. I get the following different error now - ` File "/Users/sm2511/PycharmProjects/TLOmodel/docs/tlo_publications.py", line 103 toplevel[ ^^^^^^^^^ SyntaxError: invalid syntax. Perhaps you forgot a comma?`
> 
> Oh that's an interesting one! That's suggesting there is a syntax error in
> 
> https://github.com/UCL/TLOmodel/blob/49acef5288ef0c08c49b806472349ebcaacba8aa/docs/tlo_publications.py#L103-L108
> 
> which confusingly does not create any syntax errors when running for me 😕
> 
> Can I check what Python version you have and what the version `pybtex` that got installed was (should be able to check from Python packages tab / dialog in PyCharm or by running `pip list pybtex` in a terminal)?

I've got - 
pybtex 0.24.0
Python version 3.10.13

matt-graham:
Ah this could be an issue with the use of `*(...)` with a generator expression in the indexing operation as this seems to only work on Python 3.11+

As a more minimal example

```Python
import numpy as np
array = np.arange(12).reshape(3, 4)
array[*(i for i in range(1, 3))]
```

runs fine for me on Python 3.11+ but gives a syntax error in Python 3.10.

sakshimohan:
> Ah this could be an issue with the use of `*(...)` with a generator expression in the indexing operation as this seems to only work on Python 3.11+
> 
> As a more minimal example
> 
> ```python
> import numpy as np
> array = np.arange(12).reshape(3, 4)
> array[*(i for i in range(1, 3))]
> ```
> 
> runs fine for me on Python 3.11+ but gives a syntax error in Python 3.10.

Got I will update to Python 11 in this case. Are you OK to merge this PR into master anyway?

matt-graham:
> Got I will update to Python 11 in this case. Are you OK to merge this PR into master anyway? I'm reluctant to update today because I have a deliverable due this evening and I'm afraid my analysis code might run into unexpected issues with the update which I will not have the time to address today.

Sorry missed this while writing previous message! Yes that's fine, once the checks / documentation workflows finish running I'll merge this assuming they pass!

matt-graham (approved):

PR #1524, opened 2024-11-21
Generate `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` at class initialisation

### LINKED ISSUE #1467: Population dataframe accesses in `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` are performance bottleneck
Look at the most recent successful profiling run results

https://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_1226_dbf33d69edadbacdc2dad1fc32c93eb771b03f9b.html

calls to the `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` property are consuming a lot of run time, with a total of 2570s spent in evaluating this property out of a total run time of 27&thinsp;400s (9.3%). There are three functions calling `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` with a significant amount of time recorded in each call path

- `HealthSystem.check_hsi_event_is_valid` (1110s)
- `HSI_Contraception_FamilyPlanningAppt.initialise` (545s)
- `HSI_Contraception_FamilyPlanningAppt.run` (911s)

and in all cases the vast majority of the time in the `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` call is spent in `.loc` indexing operations on the population dataframe.

Looking at the implementation of the property

https://github.com/UCL/TLOmodel/blob/5b4f328e5fc0df469eec40bcf1110a555326aa80/src/tlo/methods/contraception.py#L1150-L1170

we can see this must be arising from the access to the current contraception method property for the individual the HSI event is targetting in the line

```Python
current_method = self.sim.population.props.loc[person_id].co_contraception 
```

<details>
<summary>
This pattern of the `EXPECTED_APPT_FOOTPRINT` being a property rather than an attribute seems to also be used by a few other HSI events
</summary>

`HSI_Hiv_StartOrContinueTreatment`

https://github.com/UCL/TLOmodel/blob/5b4f328e5fc0df469eec40bcf1110a555326aa80/src/tlo/methods/hiv.py#L2968-L2986

`HSI_Tb_StartTreatment`

https://github.com/UCL/TLOmodel/blob/5b4f328e5fc0df469eec40bcf1110a555326aa80/src/tlo/methods/tb.py#L2175-L2183

`HsiBaseVaccine`

https://github.com/UCL/TLOmodel/blob/5b4f328e5fc0df469eec40bcf1110a555326aa80/src/tlo/methods/epi.py#L407-L431
</details>

I'm assuming this has been implemented like this to allow the expected appointment footprint to be based on the target's properties at the point at which it is run rather than when it is scheduled? Tagging @tbhallett, @BinglingICL and @EvaJanouskova as I think this was originally implemented in #826. 

One thing I'm not sure about is how this pattern interacts with this logic in `HSI_Event.initialise`

https://github.com/UCL/TLOmodel/blob/5b4f328e5fc0df469eec40bcf1110a555326aa80/src/tlo/methods/hsi_event.py#L333-L338

which reassigns `self.EXPECTED_APPT_FOOTPRINT` in certain cases, in which case it would end up as a plain (static) attribute rather than a property. As `HSI_Event.initialise` appears to be called in `HealthSystem.schedule_hsi_event`, that is at the time the HSI event is scheduled, this means in some cases even when `EXPECTED_APPT_FOOTPRINT` is defined as a property it will be 'locked-in' as a static attribute with values taken from population dataframe at point at which it is scheduled.

As accessing the population dataframe dynamically when `EXPECTED_APPT_FOOTPRINT` is a property seems to be a performance bottleneck, and given it does seem to be guaranteed this will actually be done at the point the HSI event is run rather than scheduled, it seems that enforcing that `EXPECTED_APPT_FOOTPRINT` is an attribute which is contructed when the HSI event is constructed (when scheduling) might be a good idea to both keep the behaviour consistent and allow us to reduce the population dataframes by just doing once on construction rather than each access of `EXPECTED_APPT_FOOTPRINT`.

matt-graham:
Resolves #1467 

Changes `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` from being a property to a static attribute, with the value evaluated when the class (HSI event) is initialised (scheduled).

This may change the value associated with `EXPECTED_APPT_FOOTPRINT` as it is now depends on the targets properties (specifically the current contraception method indicated by the categorical `co_contraception` property) at the time of scheduling rather than evaluation of `EXPECTED_APPT_FOOTPRINT`. Marking this as draft for now as will need @EvaJanouskova and/or @BinglingICL to verify if this change in behaviour is reasonable.

As discussed in #1467 repeated dataframe accesses when dynamically evaluating `EXPECTED_APPT_FOOTPRINT` for this HSI event where becoming a significant performance bottleneck - from some short profiled runs locally the changes here give around a 20% reduction in run time.

### DIFF
--- a/src/tlo/methods/contraception.py
+++ b/src/tlo/methods/contraception.py
@@ -1146,12 +1146,12 @@ class HSI_Contraception_FamilyPlanningAppt(HSI_Event, IndividualScopeEventMixin)
 
         self.TREATMENT_ID = "Contraception_Routine"
         self.ACCEPTED_FACILITY_LEVEL = _facility_level
-
-    @property
-    def EXPECTED_APPT_FOOTPRINT(self):
-        """Return the expected appt footprint based on contraception method and whether the HSI has been rescheduled."""
-        person_id = self.target
         current_method = self.sim.population.props.loc[person_id].co_contraception
+        self.EXPECTED_APPT_FOOTPRINT = self._get_expected_appt_footprint(current_method)
+
+
+    def _get_expected_appt_footprint(self, current_method: str):
+        """Return the expected appt footprint based on contraception method and whether the HSI has been rescheduled."""
         if self._number_of_times_run > 0:  # if it is to re-schedule due to unavailable consumables
             return self.make_appt_footprint({})
         # if to switch to a method
@@ -1165,9 +1165,11 @@ class HSI_Contraception_FamilyPlanningAppt(HSI_Event, IndividualScopeEventMixin)
         elif self.new_contraceptive in ['male_condom', 'other_modern', 'pill']:
             return self.make_appt_footprint({'PharmDispensing': 1})
         else:
+            warnings.warn(
+                "Assumed empty footprint for Contraception Routine appt because couldn't find actual case.",
+                stacklevel=3,
+            )
             return self.make_appt_footprint({})
-            warnings.warn(UserWarning("Assumed empty footprint for Contraception Routine appt because couldn't find"
-                                      "actual case."))
 
     def apply(self, person_id, squeeze_factor):
         """If the relevant consumable is available, do change in contraception and log it"""
--- a/src/tlo/methods/contraception.py
+++ b/src/tlo/methods/contraception.py
@@ -1147,11 +1147,11 @@ class HSI_Contraception_FamilyPlanningAppt(HSI_Event, IndividualScopeEventMixin)
         self.TREATMENT_ID = "Contraception_Routine"
         self.ACCEPTED_FACILITY_LEVEL = _facility_level
         current_method = self.sim.population.props.loc[person_id].co_contraception
-        self.EXPECTED_APPT_FOOTPRINT = self._get_expected_appt_footprint(current_method)
+        self.EXPECTED_APPT_FOOTPRINT = self._get_appt_footprint(current_method)
 
 
-    def _get_expected_appt_footprint(self, current_method: str):
-        """Return the expected appt footprint based on contraception method and whether the HSI has been rescheduled."""
+    def _get_appt_footprint(self, current_method: str):
+        """Return the appointment footprint based on contraception method and whether the HSI has been rescheduled."""
         if self._number_of_times_run > 0:  # if it is to re-schedule due to unavailable consumables
             return self.make_appt_footprint({})
         # if to switch to a method
@@ -1267,6 +1267,8 @@ class HSI_Contraception_FamilyPlanningAppt(HSI_Event, IndividualScopeEventMixin)
             self._number_of_times_run < self.module.parameters['max_number_of_runs_of_hsi_if_consumable_not_available']
         ):
             self.reschedule()
+
+        return self._get_appt_footprint(current_method)
 
     def post_apply_hook(self):
         self._number_of_times_run += 1


matt-graham:
From running the tests locally it looks like these changes break at least one test in `tests/test_contraception.py`:

```
_______________________________________________________________________________ test_record_of_appt_footprint_for_switching_to_methods[83563095832589325021] ________________________________________________________________________________
tests/test_contraception.py:485: in test_record_of_appt_footprint_for_switching_to_methods
    assert is_list_longer_than_length_of_one_and_with_first_element_nonblank_and_subsequent_blank(
E   AssertionError: assert False
E    +  where False = <function test_record_of_appt_footprint_for_switching_to_methods.<locals>.is_list_longer_than_length_of_one_and_with_first_element_nonblank_and_subsequent_blank at 0x7fb3ac5d45e0>([{'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, ...])
E    +    where [{'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, ...] = <function test_record_of_appt_footprint_for_switching_to_methods.<locals>.get_appt_footprints at 0x7fb3ac614c20>(switch_from='not_using', switch_to='female_sterilization', consumables_available=False)
```

I am not entirely sure what this test is meant to be checking. The full output of the call `get_appt_footprints(switch_from='not_using', switch_to='female_sterilization', consumables_available=False)` which is passed to `is_list_longer_than_length_of_one_and_with_first_element_nonblank_and_subsequent_blank` is

```
[{'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}, {'MinorSurg': 1}]
```

with the last condition (subsequent elements after first blank) failing.

matt-graham:
My understanding of why the test is failing is that the HSI event reschedules itself if a lack of consumables meant the change to contraception method could not be enacted

https://github.com/UCL/TLOmodel/blob/4a67b4881027857609592c0418baf944bed80236/src/tlo/methods/contraception.py#L1266-L1269

with the reschedule function passing in `self` as the HSI event instance 

https://github.com/UCL/TLOmodel/blob/4a67b4881027857609592c0418baf944bed80236/src/tlo/methods/contraception.py#L1276-L1282

This means in any subsequent runs of HSI event `apply` method the value of `EXPECTED_APPT_FOOTPRINT` will remain what is was when event was first scheduled and `__init__` method called (and so be based on the current contraception method at that point), as the rescheduled events do not create a new HSI event instance. 

I believe the intended interface for communicating changes to the appointment footprint from the initial expected footprint at the point of scheduling is for the `apply` method to return the actual appointment footprint. Adding a line 

https://github.com/UCL/TLOmodel/blob/4a67b4881027857609592c0418baf944bed80236/src/tlo/methods/contraception.py#L1271

to the end of the `apply` method to return the appointment footprint based on `current_method` in the call to `apply` appears to fix the previously failing test. 

I initially also added a line `current_method = _new_contraceptive` to the if block

https://github.com/UCL/TLOmodel/blob/4a67b4881027857609592c0418baf944bed80236/src/tlo/methods/contraception.py#L1256-L1262

so that `current_method` at the end of the method when `self._get_appt_footprint` is called reflects the method after a change has been enacted but this led to another test failure

```
FAILED tests/test_contraception.py::test_record_of_appt_footprint_for_switching_to_methods[83563095832589325021] - AssertionError: assert [{'FamPlan': 1}] == [{'PharmDispensing': 1}]
```

so I am assuming the actual appointment footprint should be based on the current contraception method at the point HSI event is being run not after it has run (and the method has potentially changed).

BinglingICL:
Hi Matt, thanks for the neat and clear change and careful checks. I think the changes you've made look fine and agree with your saying that "actual appointment footprint should be based on the current contraception method at the point HSI event is being run not after it has run (and the method has potentially changed)".

Would be grateful if Eva @EvaJanouskova could double check the logic.

EvaJanouskova:
Yes, it seems good. Thank you @matt-graham!

EvaJanouskova (approved):
Yes, it seems good. Thank you @matt-graham!

matt-graham:
From the profiling run looks like this took about 10% off the time of a 5 year run with 50k initial population ([before changes in this PR](https://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/schedule_1415_28643ea7550971f8c191f36c39b472a1264dc3de.html), [before changes in this PR](https://github-pages.ucl.ac.uk/TLOmodel-profiling/_static/profiling_html/issue_comment_1436_4a67b4881027857609592c0418baf944bed80236.html)), with the time spent in evaluating `HSI_Contraception_FamilyPlanningAppt.EXPECTED_APPT_FOOTPRINT` no longer significant enough to be picked up in the profiling output so looks this is working as intended.

BinglingICL:
[like]  She, Bingling reacted to your message:

Issue #1513, opened 2024-11-12
Speed up documentation generation


tamuri:
Improving the time taken to generate the docs would facilitate quicker deployment
Ideas:

- Remove linkcheck as part of doc build because most links are behind paywall and so fail the check (and, we're not actively checking these)
- Instead of generating tables from the resource files, integrate and use an online viewer ([example](https://www.getgrist.com/csv-viewer/)) - this would work for all files that are in CSV format.
- Sphinx seems to be regenerating html files for unchanged source files. Not sure why.

tamuri:
~Not sure why.~

- Remove `-f` argument for sphinx-apidoc
- Remove `{postargs:-E}` for sphinx-build

matt-graham:
Other than online viewer which now has a separate issue (#1526) the remainder of this was resolved by #1525

PR #1525, opened 2024-11-22
Speed up documentation build

matt-graham:
Resolves (some of) #1513 

- Remove `linkcheck` step
- Removes `-f` force option to `sphinx-apidoc` and `-E` fresh env default argument to `sphinx-build`
- Removes `sphinx.ext.viewcode` extension
- Stops last updated date text being added to footers to minimize diff noise
- Changes maximum resource file size before switching to just including a placeholder page linking to file to download rather than rendering table to 32KiB rather than 1MiB - this was a quicker alternative to suggestion in #1513 to use an online viewer. This would still be nice to have but as a lot of resource files are still XLSX can be delayed till we have finished with converting them to CSV.

Times for `tox -e docs` steps before and after changes (in both cases from a clean state):

| Step | Time (`master`) / s | Time (this PR) / s |
|------|-------------------|-------------------|
| Setup | 8.89 | 8.06 |
|  `sphinx-apidoc` | 0.31 | 0.29 |
| `tlo_methods_rst.py` | 7.51 | 7.51 |
|  `tlo_data_sources.py` | 0.34 | 0.37 |
| `tlo_contributors.py` | 0.06 |  0.06 |
| `tlo_publications.py` | 0.38 | 0.42 |
| `tlo_resources.py` | 8.17 | 1.18 |
| `hsi_events.py` (RST) | 6.95 | 7.01 |
| `hsi_events.py` (CSV) | 7.75 | 7.00 |
| `tlo_parameters.py` | 4.95 | 5.10 |
| `sphinx-build` | 216.90 | 39.63 |
| `linkcheck` | 720.06 | (removed) |
| Total | 982.26 | 76.63 | 

Biggest gain is removing `linkcheck` step but `sphinx-build` step is also significantly reduced, mainly because of previously some of the large resource files taking a while to render.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,6 @@ extensions = [
     'sphinx.ext.ifconfig',
     'sphinx.ext.napoleon',
     'sphinx.ext.todo',
-    'sphinx.ext.viewcode',
     'rawfiles'
 ]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,7 +55,6 @@ if not on_rtd:  # only set the theme if we're building docs locally
     html_theme = 'sphinx_rtd_theme'
 
 html_use_smartypants = True
-html_last_updated_fmt = '%b %d, %Y'
 html_split_index = False
 html_show_copyright = False
 html_sidebars = {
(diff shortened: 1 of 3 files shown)


tamuri (approved):
Thanks - much better!

PR #1532, opened 2024-12-03
Sort population by age-group before plotting

### LINKED ISSUE #1506: Groups in population pyramid from calibration analyses not sorted correctly.
From most recent run (2024-10-18)

Image

Looks like it came in with https://github.com/UCL/TLOmodel/pull/1404

Might be as simple as sorting the groups before plotting.

sangeetabhatia03:
This PR aims to fix #1506 
I have sorted the dataframe by pre-defined age categories available via `make_age_grp_types.categories` and checked that population pyramid is now being plotted correctly. 
Incidentally, I wonder if this line contains a typo? Since we are looping over sex,I expected data[_sex][_source].
https://github.com/UCL/TLOmodel/blob/58256ce671ba316ba55534357af5c516ac5faaa5/src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py#L201

### DIFF
--- a/src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
+++ b/src/scripts/calibration_analyses/analysis_scripts/analysis_demography_calibrations.py
@@ -260,6 +260,7 @@ def apply(results_folder: Path, output_folder: Path, resourcefilepath: Path = No
             collapse_columns=True,
             only_mean=True
         )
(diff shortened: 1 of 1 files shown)


tamuri:
> Incidentally, I wonder if this line contains a typo? Since we are looping over sex,I expected data[_sex][_source].

This is because both sexes have the same sources - it doesn't matter which you use.

tamuri (approved):

PR #1536, opened 2024-12-04
Update TB consumables to use item names not packages

tdm32:
PR updates TB consumables for diagnostics following recent runs showing xpert was not being recorded. Items are requested using the item name and not the package name. Optional items (gloves, sputum containers etc are also now added).


tdm32:
> PR updates TB consumables for diagnostics following recent runs showing xpert was not being recorded. Items are requested using the item name and not the package name. Optional items (gloves, sputum containers etc are also now added).

Now confirmed, using consumables availability='all', item number 187 (xpert) is in the consumables logger 'Item_Used'

tbhallett (approved):

PR #1542, opened 2024-12-06
Add CLI command for combining batch runs

matt-graham:
Following question from @BinglingICL on Slack, this PR adds a command to CLI interface for combining the runs from multiple batch jobs, with the assumption that the draws are aligned across the jobs (and different seeds used for runs). It does some basic checks to ensure same draw numbers across different results directories, and that run numbers in base results directory that additional runs are being merged into are consecutively numbered from 0, with the additional runs then being added with further consecutive run numbers.

### DIFF
--- a/src/tlo/cli.py
+++ b/src/tlo/cli.py
@@ -7,6 +7,7 @@ import os
 import tempfile
 from collections import defaultdict
 from pathlib import Path
+from shutil import copytree
 from typing import Dict
 
 import click
@@ -38,6 +39,7 @@ def cli(ctx, config_file, verbose):
     * submit scenarios to batch system
     * query batch system about job and tasks
     * download output results for completed job
+    * combine runs from multiple batch jobs with same draws
     """
     ctx.ensure_object(dict)
     ctx.obj["config_file"] = config_file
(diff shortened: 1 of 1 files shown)


src/tlo/cli.py
```diff
@@ -826,5 +828,68 @@ def add_tasks(batch_service_client, user_identity, job_id,
...
+    if len(additional_result_directories) == 0:
+        msg = "One or more additional results directories to merge must be specified"
+        raise click.UsageError(msg)
+    results_directories = (output_results_directory,) + additional_result_directories
+    draws_per_directory = [
+        sorted(
+            int(draw_directory.name) for draw_directory in results_directory.iterdir()
+        )
+        for results_directory in results_directories
+    ]
```
BinglingICL, line 863: Hi Matt, thanks so much again for this super helpful function of combining the results folders. Would like to let you know that I have run the script <script src="https://gist.github.com/matt-graham/b684a86e21084a5a1c74844e6b98268b.js"></script> you have kindly shared with me and found that the same lines as here would fail because the automatically downloaded .json file is also in the directory. After I deleted the json files in the directories, the code combined the files successfully.

matt-graham: Thanks @BinglingICL - I'd forgot about the JSON files as I was only testing on outputs of local runs, but definitely it will be worth updating the script to allow for there being JSON files present!

tamuri: Matt, will you make this change to this PR?

matt-graham: Yep will work on this now

matt-graham: I've now updated to skip any subpaths which are not directories when iterating over run / draw directories which should hopefully fix issue cause by draws JSON files.

tamuri (approved):

Issue #1538, opened 2024-12-04
Reduce memory usage in Malaria module


tamuri:
I had quick dig into memory usage by module, and found Malaria is holding onto memory during runtime:

```
...snip...
 'SymptomManager': 10.01573371887207,
 'PregnancySupervisor': 10.492152214050293,
 'HealthBurden': 10.63702392578125,
 'Demography': 15.597497940063477,
 'HealthSystem': 73.77490520477295,
 'Malaria': 634.1266536712646
```

The workbooks loaded in [`read_parameters()`](https://github.com/UCL/TLOmodel/blob/58256ce671ba316ba55534357af5c516ac5faaa5/src/tlo/methods/malaria.py#L243) seem to be the issue. Although loaded as parameters for Malaria, they are processed into a few lookup tables and then not accessed again (AFAICT). e.g. `interv`, `itn_district` and others are either never used or, like `irs_district`, processed into something else and then not referenced directly.

Please remove these dataframes from `self.parameters` if they are no longer needed after processing. You can either not load them, set the parameter to None (e.g. `p['itn_district'] = None`) or remove them from the parameters dictionary all together using `del p['itn_district']`.

tamuri:
The parameters to focus on are (size in MB): 

```
inf_inc 195.815411567688
clin_inc 196.7505521774292
sev_inc 195.79530811309814
```

tamuri:
> Please remove these dataframes from self.parameters...

Actually, you can just read the incident worksheets into local variables, and they'll be cleaned up when the function ends.

PR #1545, opened 2024-12-09
Remove storage of incidence data as parameters in malaria.py to reduce memory usage

joehcollins:
@tdm32 @tamuri - had some time to look at this and simply swapped out the use of self.parameters for a local variable in the function where this data is used as per Asif's suggestion. Malaria tests running fine for me locally. Let me know if theres other tweaks I can make to save Tara a job. 

This addresses #1538

### DIFF
--- a/src/tlo/methods/malaria.py
+++ b/src/tlo/methods/malaria.py
@@ -70,7 +70,6 @@ class Malaria(Module, GenericFirstAppointmentsMixin):
     }
 
     PARAMETERS = {
-        'interv': Parameter(Types.REAL, 'data frame of intervention coverage by year'),
         'itn_district': Parameter(
             Types.REAL, 'data frame of ITN usage rates by district'
         ),
@@ -235,7 +234,6 @@ class Malaria(Module, GenericFirstAppointmentsMixin):
         p = self.parameters
 
         # baseline characteristics
-        p['interv'] = workbook['interventions']
         p['itn_district'] = workbook['MAP_ITNrates']
         p['irs_district'] = workbook['MAP_IRSrates']
 
--- a/src/tlo/methods/malaria.py
+++ b/src/tlo/methods/malaria.py
@@ -71,18 +71,6 @@ class Malaria(Module, GenericFirstAppointmentsMixin):
 
     PARAMETERS = {
         'interv': Parameter(Types.REAL, 'data frame of intervention coverage by year'),
-        'clin_inc': Parameter(
-            Types.REAL,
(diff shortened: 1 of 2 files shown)


tamuri:
i think some other worksheets are also not used or needed. Now would be a good time to remove them too!

joehcollins:
@tdm32 - not sure if i've helped or made more work this this but don't want to remove any files as per Asif's comment until you have reviewed.

Noted the failed test as `test_linear_model_for_clinical_malaria()` as its calling for one of the parameters that no longer exists.

tdm32:
> i think some other worksheets are also not used or needed. Now would be a good time to remove them too!

Some of the worksheets are not used in the simulation but are used for plotting / calibration. I could try and combine them into fewer (but larger) worksheets?

tdm32:
> @tdm32 - not sure if i've helped or made more work this this but don't want to remove any files as per Asif's comment until you have reviewed.
> 
> Noted the failed test as `test_linear_model_for_clinical_malaria()` as its calling for one of the parameters that no longer exists.
@joehcollins failing test is now fixed

tdm32 (approved):

tdm32:
Thanks for this @joehcollins

tamuri:
> Some of the worksheets are not used in the simulation but are used for plotting / calibration. I could try and combine them into fewer (but larger) worksheets?

joehcollins:
perfect - thanks @tdm32!! @tamuri this should be ready for your review now

tdm32:
> > Some of the worksheets are not used in the simulation but are used for plotting / calibration. I could try and combine them into fewer (but larger) worksheets?
> 
> Do you need to load them in the malaria module?

no - there are definitely a few there that aren't loaded

tamuri:
Currently, only interested in those worksheets that are being unnecessarily loaded (or held during the simulation run) in parameters in the malaria module but not needed after read_parameters. Joe's PR deals with three of them but are there others?

For example, the line `p['interv'] = workbook['interventions']` - I don't see this dataframe being used anywhere in the malaria module. Can we remove it?

tdm32:
> Currently, only interested in those worksheets that are being unnecessarily loaded (or held during the simulation run) in parameters in the malaria module but not needed after read_parameters. Joe's PR deals with three of them but are there others?
> 
> For example, the line `p['interv'] = workbook['interventions']` - I don't see this dataframe being used anywhere in the malaria module. Can we remove it?

Ok got it! Yes I've removed this now. The others that I was thinking of aren't loaded in the malaria module

PR #1557, opened 2024-12-13
Remove population-wide HSI events

### LINKED ISSUE #1260: Check consistency of population-wide HSI events assumptions

marghe-molaro:
Fixes #1260

This PR resolves Issue #1260 by:
- Adding a check in check_hsi_event_is_valid to ensure an HSI_Event's target cannot be the entire Population.
- Removing all handling of cases where an HSI_Event targets the entire Population.

Note: This PR was already merged with PR #1493 to ensure a known instance of a Population-wide HSI_Event in ChronicSyndrome is removed, and therefore tests can pass.


src/tlo/methods/hsi_event.py
matt-graham, line 1: (Commenting on file here as I can't comment on specific lines as not updated in this PR!)

For the class docstring for `HSI_Event` it might be worth updating 

```
    Concrete subclasses should also inherit from one of the EventMixin classes
    defined in `src/tlo/events.py`, and implement at least an `apply` and
    `did_not_run` method.
```

to something like

```
    Concrete subclasses should also inherit from `IndividualScopeEventMixin`
    defined in `src/tlo/events.py`, and implement at least an `apply` and
    `did_not_run` method.
```

to be explicit that only subclassing from `IndividualScopeEventMixin` and not `PopulationScopeEventMixin` is now allowed.

We could consider going forward just directly inheriting from `IndividualScopeEventMixin` in the `HSI_Event` base class to avoid each subclass needing to individually do this, but given that would also require either updating all of the existing HSI event subclasses or having them all multiply inherit from the same class, I think probably better to leave this as a posisble future todo for now!

src/tlo/methods/healthsystem.py
```diff
@@ -2252,10 +2233,7 @@ def _get_events_due_today(self,) -> Tuple[List, List]:
                       priority=next_event_tuple.priority
                      )
 
-            elif not (
-                isinstance(event.target, tlo.population.Population)
-                or event.target in alive_persons
-            ):
+            elif event.target not in alive_persons:
                 # if individual level event and the person who is the target is no longer alive, do nothing more,
```
matt-graham, line 2237:
```suggestion
                # if the person who is the target is no longer alive, do nothing more,
```

src/tlo/methods/healthsystem.py
```diff
@@ -2359,10 +2326,7 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
                           priority=next_event_tuple.priority
                          )
 
-                elif not (
-                    isinstance(event.target, tlo.population.Population)
-                    or event.target in alive_persons
-                ):
+                elif event.target not in alive_persons:
                     # if individual level event and the person who is the target is no longer alive,
```
matt-graham, line 2330:
```suggestion
                    # if the person who is the target is no longer alive,
```

src/tlo/methods/healthsystem.py
```diff
@@ -2541,10 +2500,7 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
                       priority=next_event_tuple.priority
                      )
 
-            elif not (
-                isinstance(event.target, tlo.population.Population)
-                or event.target in alive_persons
-            ):
+            elif event.target not in alive_persons:
                 # if individual level event and the person who is the target is no longer alive,
```
matt-graham, line 2504:
```suggestion
                # if the person who is the target is no longer alive,
```

matt-graham (approved):
Thanks @marghe-molaro - this looks all good to me, great to see how much it simplifies some of the code paths! I've added some minor suggestions around adjusting comments to remove references to individual events and documenting that `HSI_Event` should only be subclassed from `IndividualScopeEventMixin` but these are minor niggles, so approving as from my perspective this looks good to go in providing tests all pass.

tbhallett (approved):
I agree that this looks GREAT!

Thanks very much @marghe-molaro 

@matt-graham 's comments for additional  / refined commenting looks good to me too.

tbhallett:
As we're approving the additional comments were uncontroversial, I've implemented them and will merge if test pass.

marghe-molaro:
Many thanks @matt-graham for the comments, and @tbhallett for wrapping it up and merging!

PR #1558, opened 2024-12-16
Allow Computing Standard Errors Across Runs in `compute_summary_statistics`

### LINKED ISSUE #1555: `compute_summary_statistics` could have option to use standard-error
Following discussion with @andrew-phillips-1 and @joehcollins it seems that some analysts and summarising uncertainty in model results using the 'standard error'. 
We could implement this in the utility method `compute_summary_statistics` to provide ease of access. An implementation could be as follows (budding on #1457):

```python
    if not use_standard_error:   # <--- `use_standard_error` is new boolean argument, defaulting to False
        lower_quantile = (1. - width_of_range) / 2.
        stats["lower"] = grouped_results.quantile(lower_quantile)
        stats["upper"] = grouped_results.quantile(1 - lower_quantile)
    else:
       #  Use standard error concept whereby we're using the intervals to express a 95% CI on the value of the mean. This will make width of uncertainty become narrower with more runs.
        std_deviation = grouped_results.std()
        std_error = std_deviation / np.sqrt(len(grouped_results))
        z_value = st.norm.ppf(1 - (1. - width_of_range) / 2.)  # (import scipy.stats as st)
        stats["lower"] = stats['central'] - z_value * std_error
        stats["upper"] = stats['central'] + z_value * std_error
```

A question for @andrew-phillips-1 -- I presume it's only appropriate to use this concept in certain circumstances - what would these be? And also, I presume we should raise an error if sometimes tries to use a summary measure other than mean?

tbhallett:
Fixes #1555


src/tlo/analysis/utils.py
```diff
@@ -373,11 +375,18 @@ def compute_summary_statistics(
...
     :param collapse_columns: Whether to simplify the columnar index if there is only one run (cannot be done otherwise).
     :param only_central: Whether to only report the central value (dropping the range).
     :return: A dataframe with computed summary statistics.
-
     """
+
+    if use_standard_error:
+        if not central_measure == 'mean':
+            warnings.warn("When using 'standard-error' the central measure in the summary statistics is always the mean.")
+            central_measure = 'mean'
```
tbhallett, line 388:
```suggestion
    if use_standard_errorr and not central_measure == 'mean':
        warnings.warn("When using 'standard-error' the central measure in the summary statistics is always the mean.")
        central_measure = 'mean'
```

andrew-phillips-1 (approved):
This looks fine as far as I am able to tell.

PR #1563, opened 2025-01-07
Convert Excel files to CSV - Uncategorised files

mnjowe:
This PR converts all excel files that haven't been included in the previous PR's on the same.


mnjowe:
Hi @tamuri or @tbhallett. Do you have an idea of what is causing the documentation fail? you can scroll down [the details page](https://github.com/UCL/TLOmodel/actions/runs/12665068240/job/35294181605?pr=1563) for more details. I can't understand exactly where the error is coming from

tamuri:
The error is `pandas.errors.EmptyDataError: No columns to parse from file` - perhaps one of the new files you've created is empty?

mnjowe:
I see. Thanks @tamuri , indeed there is a file `structure.csv`  in `Resourcefile_Lifestyle_Enhanced` folder that's empty. I looked at the original sheet in the excel file(`Resourcefile_Lifestyle_Enhanced.xlsx`) and shows it has diagrams only(see the attached). May be that's the reason these couldn't be read and added to `structure.csv`? 

Since we have the original excel file in sharepoint, I will delete this file here and see if that resolves the issue.

Screenshot 2025-01-08 at 12 56 39

tbhallett (commented):
This all seems fine to me @mnjowe 

The tricky one is obviously `get_parameters_for_improved_healthsystem_and_healthcare_seeking()`

Please could you confirm that the resulting `dict` in that test is EXACTLY the same when using the csv's as when using the original xlsx?

mnjowe:
@tbhallett, let me check

mnjowe:
@tbhallett. I did the checking by running same test from this brach and master. They output seems similar(see the two logs below)
[output_master.log](https://github.com/user-attachments/files/18374874/output_master.log)
[output.log](https://github.com/user-attachments/files/18374876/output.log)

tamuri (approved):

PR #1570, opened 2025-01-16
Convert Excel files to CSV - Cardiometabolic disorders 

mnjowe:
Seems I missed converting cardio excel files in the previous round of PR's. Converting them here


tbhallett (changes requested):
Seems good to me. 

Only change is that I'd delete the `info.csv` sheets, as we don't need those.

tbhallett (approved):

PR #1573, opened 2025-01-16
Add note about behaviour of asserts in batch jobs

### LINKED ISSUE #1572: Document that asserts disabled by default in batch jobs
Since #1051 was merged, jobs submitted on Azure using batch job submission system will have `assert` statements disabled by default. As pointed out by @EvaJanouskova this might not be entirely clear to users and can lead to asserts which fail only on longer / higher population simulations run as batch jobs being missed. Ideally we should document this behaviour more clearly and indicate that `--asserts-on` flag can be used to enable asserts.

matt-graham:
Resolves #1572 

[Note also added to wiki page on how to run a model analysis](https://github.com/UCL/TLOmodel/wiki/How-to-do-a-model-analysis/17907213ff08e45aac2d0abc903b7201a6cab15f).

### DIFF
--- a/docs/azure_batch.rst
+++ b/docs/azure_batch.rst
@@ -93,10 +93,10 @@ Each of the subcommands of ``tlo`` has a ``--help`` flag for further information
     
 .. note:: **Behaviour of assert statements**
 
-   Scenario runs submitted with tlo batch-submit will be default run with any assert statements disabled.
-   Typically these are only needed for initial checking of correctness of code during local runs but as they add some overhead, 
+   Scenario runs submitted with ``tlo batch-submit`` will by default run with any assert statements disabled.
+   Typically, these are used for checking correctness of code in tests and local runs but as they add some overhead, 
    we can get a useful speed-up by disabling them on longer jobs. 
-   To run jobs with assertions enabled pass the ``--asserts-on flag`` to ``tlo batch-submit``.
+   To run jobs with assertions enabled pass the ``--asserts-on`` flag` to ``tlo batch-submit``.
 
 *List jobs currently on Azure Batch where id contains "tamuri"*::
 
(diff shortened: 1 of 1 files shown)


EvaJanouskova (approved):
This is great, thank you Matt!

docs/azure_batch.rst
```diff
@@ -90,6 +90,13 @@ Each of the subcommands of ``tlo`` has a ``--help`` flag for further information
 *Submit scenario to Azure Batch*::
 
     tlo batch-submit src/scripts/dev/tlo_q1_demo.py
+    
+.. note:: **Behaviour of assert statements**
+
+   Scenario runs submitted with tlo batch-submit will be default run with any assert statements disabled.
```
tamuri, line 96:
```suggestion
   Scenario runs submitted with ``tlo batch-submit`` will by default run with any assert statements disabled.
```

docs/azure_batch.rst
```diff
@@ -90,6 +90,13 @@ Each of the subcommands of ``tlo`` has a ``--help`` flag for further information
...
+   Typically these are only needed for initial checking of correctness of code during local runs but as they add some overhead, 
+   we can get a useful speed-up by disabling them on longer jobs. 
+   To run jobs with assertions enabled pass the ``--asserts-on flag`` to ``tlo batch-submit``.
```
tamuri, line 99:
```suggestion
   To run jobs with assertions enabled pass the ``--asserts-on`` flag` to ``tlo batch-submit``.
```

docs/azure_batch.rst
```diff
@@ -90,6 +90,13 @@ Each of the subcommands of ``tlo`` has a ``--help`` flag for further information
 *Submit scenario to Azure Batch*::
 
     tlo batch-submit src/scripts/dev/tlo_q1_demo.py
+    
+.. note:: **Behaviour of assert statements**
+
+   Scenario runs submitted with tlo batch-submit will be default run with any assert statements disabled.
+   Typically these are only needed for initial checking of correctness of code during local runs but as they add some overhead,
```
tamuri, line 97:
```suggestion
   Typically, these are used for checking correctness of code in tests and local runs but as they add some overhead, 
```

tamuri (approved):
Some suggestions

PR #1586, opened 2025-02-12
Simplify the collection of daily HSI events for mode 1

tamuri:
- Collecting index of alive individuals repeatedly allocates memory and becomes slower than querying directly (esp. on large populations)
- In mode 1, HSI events have the same priority so queue is effectively ordered by date. Stop processing when we reach the events from the next day


marghe-molaro (approved):
This looks good to me!

PR #1588, opened 2025-02-12
Reuse array in `get_squeeze_factors` to avoid excessive memory allocation

tamuri:
Profiling revealed that even under small population (10k) sizes, this method was using several gigabytes of temporary allocations for squeeze factor lists. This PR reuses an array to prevent repeatedly allocating memory. Allocating these small amounts of memory can memory fragmentation.

- Add a numpy array, and grow it as needed to accommodate number of events being processed.
- Reuse the numpy array to store squeeze factors.
- We don't need to overwrite the array, because we always fill items 0 to n, where n is number of events.
- No need to check whether every officer is unavailable. Break on the first match.
- The store is only written to inside the method, transparent to caller.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -1,6 +1,7 @@
 import datetime
 import heapq as hp
 import itertools
+import math
 import re
 import warnings
 from collections import Counter, defaultdict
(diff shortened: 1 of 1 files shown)


marghe-molaro (approved):
Looks good to me. I think typically the required size will be much smaller than 500 (I think (!) usually max 3-4 officers per appt footprint, and max 3-4 appt footprints per event, so something ten times smaller (50) should already be more than enough. But I doubt reducing it would make much difference!).
Interesting to learn that a for loop can be faster than the any() function!

tamuri:
> Looks good to me. I think typically the required size will be much smaller than 500 (I think (!) usually max 3-4 officers per appt footprint, and max 3-4 appt footprints per event

The `_get_squeeze_factors_store` array is used to hold the squeeze factors calculate for every HSI event for the day. These can be quite a few, especially at larger population sizes.

PR #1591, opened 2025-02-12
Delete the availability estimates dataframe once Consumables initialised

### LINKED ISSUE #1584: Remove parameter reference to `availability_estimates` dataframe in HealthSystem
Noticed whilst looking at objects holding memory that unnecessary reference to `availability_estimates`, which is dataframe loaded of `ResourceFile_Consumables_availability_small.csv`, is not needed after `Consumables` has been initialised. 

https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/healthsystem.py#L660-L669

Any reason why this can't be a normal variable in `read_parameters`?

tamuri:
Resolves #1584. The availability estimates dataframe uses quite a lot of memory but we don't need it once Consumables has been initialised. This is a quick fix that simply deletes the key from the parameters dictionary.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -670,6 +670,8 @@ class HealthSystem(Module):
             rng=rng_for_consumables,
             availability=self.get_cons_availability()
         )
+        # We don't need to hold onto this large dataframe
+        del self.parameters['availability_estimates']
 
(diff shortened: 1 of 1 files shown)


tbhallett (approved):
Thanks for doing that one.

PR #1592, opened 2025-02-13
Improve dataframe access in some hotspots

tamuri:
Some population dataframe access patterns showed up during profiling. These changes reduced runtime in these methods in contraception and symptommanager, 6x and 3x respectively.


tbhallett (approved):
amazing

EvaJanouskova:
This is great, thanks!

PR #1598, opened 2025-02-24
Tell isort to skip setuptools-scm's generated `_version.py`

tamuri:
isort does not like change in v8.2.0 of setuptools https://github.com/pypa/setuptools-scm/commit/24d7797e1872aca7936dcd71a1aac81a08ee5326

Quick way to fix failing CI.


tamuri:
isort does have option to [skip files in gitignore](https://pycqa.github.io/isort/docs/configuration/options.html#skip-gitignore) - might be a better solution but need to think through what that means.

Issue #1499, opened 2024-10-29
Tomography appointment is happening at level 2 rather than level 3


BinglingICL:
Scenario run results shows that Tomography appointment is happening at level 2 rather than at level 3. As defined by ResourceFile_Appt_Time_Table.csv, this appointment should happen and only happen at level 3/central hospitals.

Initial search shows that Tomography is only used by RTI module; but not yet know the cause of this inconsistent place of level.

mnjowe:
Hi Bingling. I think [this code](https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/rti.py#L3347-L3352) has to be changed to
```
if  'Tomography' in list(self.EXPECTED_APPT_FOOTPRINT.keys()):
            self.ACCEPTED_FACILITY_LEVEL = '3'
            self.add_equipment({'Computed Tomography (CT machine)', 'CT scanner accessories'})

self.add_equipment(self.healthcare_system.equipment.from_pkg_names('X-ray'))
```
I think with the current implementation, [this condition](https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/rti.py#L3350) will never run since the first block will always be true i.e. `DiagRadio` will always be in the `list(self.EXPECTED_APPT_FOOTPRINT.keys())`. This is according to how the expected footprint is being initialised [here](https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/rti.py#L3339) but also the fact [that this function](https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/rti.py#L3345) is just adding another key/appointment `Tomography` to the existing footprint if there are any injuries requiring tomography appointment and overwriting `DiagRadio` if any injuries requiring its appointment

I think the above is the reason you're having tomography appointment happening at facility level 2. It is because [this block is not running](https://github.com/UCL/TLOmodel/blob/3599cdc477cc195799bc6677753f25729c75664f/src/tlo/methods/rti.py#L3351C12-L3352C15) hence no change of facility level

BinglingICL:
Thanks very much @mnjowe for finding out the reason and providing the solution. Would raise a PR to fix it accordingly. Many thanks and sorry for not getting back on this earlier.

BinglingICL:
[...]

PR #1606, opened 2025-03-19
fix level of Tomography appointment

BinglingICL:
This is to fix issue #1499.


tbhallett (approved):
Thanks very much @BinglingICL  and @mnjowe

Issue #542, opened 2022-04-06
Complete the `Wasting` module


tbhallett:
The `Wasting` module was first proposed in #289, from which `Stunting` was split out (#390), and `Wasting` was left until now.

The work to do is:
- [x] Add the ResourceFile (named here as `ResourceFile_Undernutrition.xlsx`), which was omitted in the #289
- [x] Consider the initial review comments in #289 
- [x] Introduce basic amount of refactoring and updating needed for this to function as module
- [x] Duplicate the refactoring work done for `Stunting` in #390 here.
- [x] Introduce the same level of testing as done for `Stunting` in #390.
- [x] Check that the performance of the module is as expected.

A PR has been opened that brings in the three files that will be changed (#543)

mnjowe:
Hi Everyone

Kindly find attached outputs from the wasting module. I will not be surprised if the outputs don't entirely meet the expectation for there's still some work to be done in the code. This is just a first touch where I wanted to ensure that


1. An analysis file has been added which will give us a picture of the wasting outputs
2. All wasting tests are running(for this, I had to make some minor changes to the wasting module. see my last commits [here](https://github.com/UCL/TLOmodel/pull/543)
3. A wasting resource file has been added

May I kindly request your input on some other outputs we are expecting to see from the wasting module plus any other comment on the expectation on the current outputs. Thanks 

[wasting_outputs.docx](https://github.com/UCL/TLOmodel/files/12828032/wasting_outputs.docx)

mnjowe:
This is noted @timcolbourn , Thanks. 

We will surely get in touch on the interpretation of parameters and the word document as soon as we are happy with  the current work on the programming side of wasting module. Thanks

mnjowe:
Hi @timcolbourn , @tbhallett and @andrew-phillips-1(optional). Myself and @thewati have finished looking at both the code and the document. We have listed below Epi question we need some help from @timcolbourn.  For programming ones @tbhallett, They're many so we would love to work on them, push to the PR and ask for your review. Thanks

Epi issues/questions - @timcolbourn 
1.  It seems there are deaths from MAM(see [this block of code](https://github.com/UCL/TLOmodel/blob/4c089a1702260ce04a1ea2e8289966557eb681a6/src/tlo/methods/wasting.py#L646-L649)) in the module. Shouldn't deaths only occur from SAM? and all MAM cases that haven't recovered progress to SAM? 
[...]

Issue #1431, opened 2024-07-23
Check equipment use logged by the model


sakshimohan:
PR #1341 allowed for the logging of equipment used for service delivery in the model. The `healthsystem.summary` logger now records by `Facility_ID` which pieces of equipment are called upon my modules to run HSIs. Based on an unconstrained 10-year run (100,000 population) from 11 July 2024, 50 pieces of equipment were logged across different levels of care - see .csv file below. This issue requests disease module leads to verify whether this log makes sense, i.e. whether they would like to incorporate the use of any further pieces of equipment at any level or remove any. 

[tlo_equipment_log_summary.csv](https://github.com/user-attachments/files/16349826/tlo_equipment_log_summary.csv)

sakshimohan:
Easily readable version of .csv below

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


  |   | Facility_Level |   |   |   |  
-- | -- | -- | -- | -- | -- | --
Equipment_name | Item_code | 1a | 0 | 2 | 3 | 4 | 5
Height Pole (Stadiometer) | 0 | TRUE | - | TRUE | - | - | -
Weighing scale | 1 | TRUE | - | TRUE | - | - | -
Autoclave | 2 | - | - | TRUE | - | - | -
Blood pressure machine | 6 | TRUE | - | TRUE | - | - | -
Thermometer | 7 | TRUE | - | TRUE | - | - | -
Stethoscope | 13 | - | - | TRUE | - | - | -
Glucometer | 15 | TRUE | - | - | - | - | -
Haemoglobinometer | 17 | - | - | TRUE | - | - | -
[...]

PR #1609, opened 2025-04-09
Wasting Module

EvaJanouskova:
Fixes https://github.com/UCL/TLOmodel/issues/542

Replaces #543, as too many changes were done in order to reach module calibration in a side branch.


EvaJanouskova:
@tamuri, @tbhallett, @timcolbourn, @mnjowe, @thewati, I've created a new PR as I made too many changes aside while calibrating the module. _(I initially expected it would be a few commits, and I would then add them back to the original branch, but it doesn't seem to be worth doing that anymore.)_ So, here is the new PR for the wasting module. Should I now close the original one: #543? (It is noted in the description of this PR that this is replacing #543, to easily access the conversation there.)

EvaJanouskova:
Hi @tamuri and @tbhallett, Since I finally calibrated the module, I'm now preparing the code for the PR. I have a few questions though. I've prepared the Write-up with the version I have in mind to be the final-ish for the study we're working on. However, the code is not there yet. Is there still a rush to have this in ASAP, or not anymore?

If so, what is the acceptable solution:
A. Could the first version be merged into the master without the Write-up?
B. Could the first version be merged in with a Write-up containing a note that it is not in line with the current version of the code, and the code will be updated soon to be in line with the Write-up?
C. Does the code need to be updated first to be in line with the Write-up, and only then can this be merged into the master?

Thank you for your advice!

EvaJanouskova:
@tamuri, Could you please have a look and advise me on why the `test_module_dependencies` is failing? Thank you!

mnjowe:
Hi @EvaJanouskova. I think you should resolve the below error in your code first, and then that test will pass.
```
.tox/py311/lib/python3.11/site-packages/tlo/methods/wasting.py:1320: in apply
    assert wasted_days >= get_min_length(recov_how, person_id, whz),\
E   AssertionError:  The person_id=979 is wasted_days=43 < minimal expected length= 48.0 days when recov_opt='sev_SAM_tx_full_recov'.
```

EvaJanouskova:
> Hi @EvaJanouskova. I think you should resolve the below error in your code first, and then that test will pass.
> 
> ```
> .tox/py311/lib/python3.11/site-packages/tlo/methods/wasting.py:1320: in apply
>     assert wasted_days >= get_min_length(recov_how, person_id, whz),\
> E   AssertionError:  The person_id=979 is wasted_days=43 < minimal expected length= 48.0 days when recov_opt='sev_SAM_tx_full_recov'.
> ```

@mnjowe, @tamuri, It is happening because the simulation in that test is running with default availability of equipment. **Is this intentional?**

I only prepared the asserts for the 'all' availability of equipment but included some code to be used later once the default availability is ready to be used. I shouldn't have include it yet since it is not ready and calibrated. I will fix it.

EvaJanouskova:
@timcolbourn, @tbhallett, @andrew-phillips-1, Would you agree if I rename the Wasting module to Acute Malnutrition module as the wasting is only one of the modelled indicators, but the actual disease modelled is acute malnutrition?

tbhallett:
> @timcolbourn, @tbhallett, @andrew-phillips-1, Would you agree if I rename the Wasting module to Acute Malnutrition module as the wasting is only one of the modelled indicators, but the actual disease modelled is acute malnutrition?
> 
> But probably worth to leave it for the next PR. What do you think?

Yeah, sounds good. Why wait for next PR?

andrew-phillips-1:
That seems fine to me also Eva

timcolbourn:
agree acute malnutrition is a good name, and no need to wait for the next PR

mnjowe:
I would agree that it’s best to wait for the next PR, especially considering the current use(number of places wasting has been mentioned) of the file name `wasting` and the disease module name `Wasting`

EvaJanouskova:
> I would agree that it’s best to wait for the next PR, especially considering the current use(number of places wasting has been mentioned) of the file name `wasting` and the disease module name `Wasting`. Just a quick search on usages/mentions in the project shows 94 matches in 31 files

Thanks Emmanuel, I've discussed it more with Tim C, and he also agrees now to leave it for the next PR.

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+                         'directory': tmpdir,
+                         'custom_levels': {
+                             "*": logging.WARNING,
+                             "tlo.methods.wasting": logging.INFO}
+                     })
+
+    sim.register(demography.Demography(resourcefilepath=resourcefilepath),
+                 enhanced_lifestyle.Lifestyle(resourcefilepath=resourcefilepath),
+                 healthsystem.HealthSystem(resourcefilepath=resourcefilepath,
+                                           disable=False,
```
tbhallett, line 63: faster test performance is disable=True, if you don't need the health system to do anything (except let everything run)...?

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+                 healthseekingbehaviour.HealthSeekingBehaviour(resourcefilepath=resourcefilepath),
+                 healthburden.HealthBurden(resourcefilepath=resourcefilepath),
+                 simplified_births.SimplifiedBirths(resourcefilepath=resourcefilepath),
+                 wasting.Wasting(resourcefilepath=resourcefilepath)
+                 )
+    return sim
+
+
+@pytest.mark.slow
+def test_basic_run(tmpdir):
```
tbhallett, line 76: test could have more informative name --- e.g. `test_integrity_of_properties_of_wasting_module`

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+    sim = get_sim(tmpdir)
+    sim.register(DummyModule())
+    sim.make_initial_population(n=popsize)
+    sim.simulate(end_date=end_date)
+
+
+def test_wasting_incidence(tmpdir):
+    """ Check incidence of wasting is happening as expected. """
+    # get simulation object:
+    dur = pd.DateOffset(days=0)
```
tbhallett, line 181: this is done to ensure simulation doesn't run -- so maybe the assignment to `dur` can be dropped, or its given a better name, like `zero_dur` (?)

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+    # Get person to use
+    df = sim.population.props
+    person_id = df.index[0]
+    df.at[person_id, 'is_alive'] = True
+
+    # 1. Check daly weight for well-nourished person (weight is 0.0)
+    # Reset diagnostic properties
+    df.loc[person_id, 'un_WHZ_category'] = 'WHZ>=-2'
+    df.loc[person_id, 'un_am_nutritional_oedema'] = False
+    df.loc[person_id, 'un_am_MUAC_category'] = '>=125mm'
```
tbhallett, line 242: use `df.at` for consistency when reading/writng to single cell

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+
+    # Verify diagnosis - an individual should be well
+    wmodule.clinical_acute_malnutrition_state(person_id, df)
+    assert df.loc[person_id, 'un_clinical_acute_malnutrition'] == 'well'
+
+    # Report daly weight for this individual
+    daly_weights_reported = wmodule.report_daly_values()
+
+    # Verify that individual has no daly weight
+    assert daly_weights_reported.loc[person_id] == 0.0
```
tbhallett, line 252: use .at (here and elsewhere)

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+    polling.apply(sim.population)
+
+    # Check properties of individuals: should now be moderately wasted
+    under5s = df.loc[df.is_alive & (df['age_years'] < 5)]
+    assert all(under5s['un_ever_wasted'])
+    assert (under5s['un_WHZ_category'].eq('-3<=WHZ<-2')).all()
+    assert (under5s['un_last_wasting_date_of_onset'].eq(sim.date)).all()
+
+
+def test_report_daly_weights(tmpdir):
```
tbhallett, line 213: very good test. 

don't make this change now, but - for next time - not that this would have been even more readable as a for loop across those four conditions.

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+        person = df.loc[person_id]
+        assert person['un_ever_wasted']
+        assert person['un_WHZ_category'] == '-3<=WHZ<-2'
+        assert person['un_last_wasting_date_of_onset'] == sim.date
+        assert pd.isnull(person['un_am_tx_start_date'])
+        assert pd.isnull(person['un_am_recovery_date'])
+        if am_state_expected == 'MAM':
+            assert df.at[person_id, 'un_clinical_acute_malnutrition'] == 'MAM'
+        else:
+            assert df.at[person_id, 'un_clinical_acute_malnutrition'] == 'SAM'
```
tbhallett, line 366: why not make the check be to the `am_state_expected` directly?

tests/test_wasting.py
```diff
@@ -0,0 +1,1062 @@
...
+            assert not person['un_sam_with_complications']
+            assert pd.isnull(person['un_am_recovery_date'])
+            assert pd.isnull(person['un_sam_death_date'])
+
+
+def test_tx_recovery_to_MAM_severe_acute_malnutrition_without_complications(tmpdir):
+    """ Check the onset of symptoms with uncomplicated SAM, check recovery to MAM with tx when
+    the progression to severe wasting is certain, hence no natural recovery from moderate wasting,
+    the natural death due to SAM is certain, hence no natural recovery from severe wasting,
+    and check natural death canceled w\ tx and symptoms resolved when recovered to MAM with tx. """
```
tbhallett, line 411: I really appreciate the docstrings. Very small thing - but could you write out in full to avoid the short hands that some people won't follow (e.g. w/, tx, etc/.)

tbhallett (commented):
So far so (VERY) good. (Sending first batch of comments and commits)

EvaJanouskova:
> @timcolbourn, @tbhallett, Btw, we calibrated the number of direct deaths, i.e. those due to protein-energy malnutrition. However, we didn’t adjust the overall deaths in the TLO model. Since these deaths weren’t previously included, they should have been captured under 'other_causes', or something similar, which accounts for deaths from causes we don’t yet simulate, right?
> 
> So now, we should remove some deaths from 'other_causes', shouldn’t we?

Otherwise, we’re increasing under-5 deaths by approximately 2,000 per year.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        # Dict to hold the DALY weights
+        daly_wts = dict()
+
+        df = self.sim.population.props
+        # Get DALY weights
+        get_daly_weight = self.sim.modules['HealthBurden'].get_daly_weight
+
+        daly_wts['mod_wasting_with_oedema'] = get_daly_weight(sequlae_code=461)
+        daly_wts['sev_wasting_w/o_oedema'] = get_daly_weight(sequlae_code=462)
+        daly_wts['sev_wasting_with_oedema'] = get_daly_weight(sequlae_code=463)
```
tbhallett, line 622: would be lots more efficient to do the lookup of daly-weights once and store the value. (this function is called every month)

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        # will not be assessed during this appt
+        if (individual_properties['age_years'] >= 5) or \
+            (individual_properties['un_last_wasting_date_of_onset'] < individual_properties['un_am_tx_start_date'] <
+             self.sim.date) or \
+            (self.sim.date == individual_properties['un_last_nonemergency_appt_date']):
+            if self.sim.date == individual_properties['un_last_nonemergency_appt_date']:
+                logger.debug(
+                    key="multiple non-emergency appts on same day",
+                    data=f"A non-emerg. appt runs again on the same date {self.sim.date=} for the {person_id=}. "
+                         "All DOs related to wasting are cancelled."
```
tbhallett, line 672: what does this line mean? What is DOs?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+            return
+
+        #or if HSI was already scheduled due to growth monitoring, won't be checked for acute malnutrition again
+        hsi_event_scheduled = [
+            ev
+            for ev in self.sim.modules["HealthSystem"].find_events_for_person(person_id)
+            if isinstance(ev[1], (HSI_Wasting_SupplementaryFeedingProgramme_MAM,
+                                  HSI_Wasting_OutpatientTherapeuticProgramme_SAM,
+                                  HSI_Wasting_InpatientTherapeuticCare_ComplicatedSAM))
+        ]
```
tbhallett, line 683: pleasd don't use find_event_for_person as its soooo slow, as docstring says - it's for testing/debugging only. What are you aiming to do here?

tbhallett: why not just do nonthing if the persons is currently being treated?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+                Wasting_RecoveryToMAM_Event: 'un_recov_to_mam_to_cancel',
+                Wasting_FullRecovery_Event: 'un_full_recov_to_cancel',
+                Wasting_ProgressionToSevere_Event: 'un_progression_to_cancel'
+            }
+
+            # add dates of natural history events to be cancelled due to tx, the health outcome will be driven by the tx
+            for date in dates:
+                df.at[person_id, event_type_map[event_type]].append(date)
+
+        # else: event_type not scheduled, hence no need to cancel any
```
tbhallett, line 832: this is not the way to cancel events. The look of of "find_events_for_person" is ONLY for testing/debugging as it's extremely slow.

What you need to do is -- in the event itself -- check whether it should still run. And if not, then do not do anything.

This check can either be:...
....  a flag stored on the instance of the event itself: this requires you to store pointers to the event.
..... through looking at the population dataframe: e.g. store the date you want the event to run, and if you don't want it to run, change it do pd.NaT... then when the event run, only let it do something if the today's date is the date stored in the datafrmae

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+            if df.at[person_id, 'un_clinical_acute_malnutrition'] == 'MAM':
+                # schedule full recovery after duration of moderate wasting
+                self.sim.schedule_event(event=Wasting_FullRecovery_Event(
+                    module=self.module, person_id=person_id), date=outcome_date)
+            else: # == SAM
+                # schedule recovery to MAM after duration of moderate wasting
+                self.sim.schedule_event(event=Wasting_RecoveryToMAM_Event(
+                    module=self.module, person_id=person_id), date=outcome_date)
+
+class Wasting_ProgressionToSevere_Event(Event, IndividualScopeEventMixin):
```
tbhallett, line 911: why isn't this a population level process?
e.g. you could move people 'en masse' to the next stage in the regular polling event - which would avoid the need for all these individual level events, which would be much more efficient.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        df.at[person_id, 'un_am_nutritional_oedema'] = False
+        df.at[person_id, 'un_sam_with_complications'] = False
+        df.at[person_id, 'un_sam_death_date'] = pd.NaT # death is cancelled if was scheduled
+        df.at[person_id, 'un_am_tx_start_date'] = pd.NaT
+        # Start without treatment, treatment will be applied with HSI if care sought
+        df.at[person_id, 'un_am_treatment_type'] = 'none'
+
+        # this will clear all wasting symptoms (applicable for SAM, not MAM)
+        self.sim.modules["SymptomManager"].clear_symptoms(
+            person_id=person_id, disease_module=self.module
```
tbhallett, line 1170: why not refer to the symptom self.module.symptom_of_wasting (quicker than asking symptom manager to do a lookup)

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+
+        self.TREATMENT_ID = "Undernutrition_GrowthMonitoring"
+        self.ACCEPTED_FACILITY_LEVEL = '1a'
+
+    @property
+    def EXPECTED_APPT_FOOTPRINT(self):
+        """Return the expected appointment footprint based on attendance at the HSI event."""
+        rng = self.module.rng
+        p = self.module.parameters
+        person_age = self.sim.population.props.loc[self.target].age_exact_years
```
tbhallett, line 1253: do this expensive lookup using .at  .at[self.target, 'age_exact_years'] and do it only once (at init) and store the result.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+                    return p['growth_monitoring_frequency_days_agecat'][2]
+
+            person_monitoring_frequency = get_monitoring_frequency_days(df.at[person_id, 'age_exact_years'])
+            if (df.at[person_id, 'age_exact_years'] + (person_monitoring_frequency / 365.25)) < 5:
+                # schedule next growth monitoring
+                self.sim.modules['HealthSystem'].schedule_hsi_event(
+                    hsi_event=HSI_Wasting_GrowthMonitoring(module=self.module, person_id=person_id),
+                    topen=self.sim.date + pd.DateOffset(days=person_monitoring_frequency),
+                    tclose=None,
+                    priority=2
```
tbhallett, line 1312: why create new event when you could re-use `self` event....?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        # or if HSI was already scheduled due to care-seeking, no need to attend the growth monitoring
+        hsi_event_scheduled = [
+            ev
+            for ev in self.sim.modules["HealthSystem"].find_events_for_person(person_id)
+            if isinstance(ev[1], (HSI_Wasting_SupplementaryFeedingProgramme_MAM,
+                                  HSI_Wasting_OutpatientTherapeuticProgramme_SAM,
+                                  HSI_Wasting_InpatientTherapeuticCare_ComplicatedSAM))
+        ]
+        if hsi_event_scheduled:
+            return
```
tbhallett, line 1333: please do not use these testing/debuggging functions

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        def get_attendance_prob(age):
+            if age < 1:
+                return p['growth_monitoring_attendance_prob_agecat'][0]
+            if age <= 2:
+                return p['growth_monitoring_attendance_prob_agecat'][1]
+            else:
+                return p['growth_monitoring_attendance_prob_agecat'][2]
+
+        # perform growth monitoring if attending
+        self.attendance = rng.random_sample() < get_attendance_prob(person_age)
```
tbhallett, line 1266: this will mean if if no "attendance" there the HSI still runs, but just without any cost. That seems odd and probably unintended?
Better to do determine if attendance will happen when scheduling the "next appointment" and not scheduling one if it won't be attended.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        # the person may not attend the appt
+        if not self.attendance:
+            return
+
+        available_equipment = []
+        for equip in ['Height Pole (Stadiometer)', 'Weighing scale', 'MUAC tape']:
+            available = rng.random_sample() < HSI_Event.probability_all_equipment_available(self, equip)
+            if available:
+                available_equipment.append(equip)
+        self.add_equipment(set(available_equipment))
```
tbhallett, line 1344: please declare all equipment -- not conditional on whether it is available or not.

if anything is so essential that the HSI could not proceed without it, declare it in the init.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+            schedule_tx_by_diagnosis(HSI_Wasting_OutpatientTherapeuticProgramme_SAM)
+        else:  # (diagnosis == 'SAM') and complications:
+            # ITC intervention if diagnosed as complicated SAM
+            schedule_tx_by_diagnosis(HSI_Wasting_InpatientTherapeuticCare_ComplicatedSAM)
+
+    def did_not_run(self):
+        logger.debug(key="HSI_Wasting_GrowthMonitoring",
+                     data="HSI_Wasting_GrowthMonitoring: did not run"
+                     )
         pass
```
tbhallett, line 1381: delete 'pass'

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+            # Record that the treatment is provided:
+            df.at[person_id, 'un_am_treatment_type'] = 'CSB++'
+            self.module.do_when_am_treatment(person_id, intervention='SFP')
+        else:
+            logger.debug(key='debug',
+                         data=f"Consumable(s) not available, hence {self.TREATMENT_ID} cannot be provided.")
+
+    def did_not_run(self):
+        logger.debug(key='debug', data=f'{self.TREATMENT_ID}: did not run')
+        pass
```
tbhallett, line 1429: delete `pass`

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+    This is the outpatient management of SAM without any medical complications
+    """
+
+    def __init__(self, module, person_id):
+        super().__init__(module, person_id=person_id)
+
+        # Get a blank footprint and then edit to define call on resources
+        # of this treatment event
+        the_appt_footprint = self.sim.modules["HealthSystem"].get_blank_appt_footprint()
+        the_appt_footprint['U5Malnutr'] = 1
```
tbhallett, line 1443: for foootprints, please use the builit-in function 
`self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({"U5Malnutr": 1})`

(here and elsewhere)

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+
+        # Define the necessary information for an HSI
+        self.TREATMENT_ID = 'Undernutrition_Feeding_Outpatient'
+        self.EXPECTED_APPT_FOOTPRINT = the_appt_footprint
+        self.ACCEPTED_FACILITY_LEVEL = '1a'
+        self.ALERT_OTHER_DISEASES = []
```
tbhallett, line 1449:
```suggestion
```
This has been deprecated for a long time, please remove the 'ALERT_OTHER_DISEASES' declarations.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        pop_sizes_dict['total__under5'] = len(under5s)
+
+        # log wasting prevalence
+        logger.info(key='wasting_prevalence_props', data=wasting_prev_dict)
+
+        # log pop sizes
+        logger.info(key='pop sizes', data=pop_sizes_dict)
+
+
+class Wasting_InitLoggingEvent(Event, PopulationScopeEventMixin):
```
tbhallett, line 1879: this seems to have a lot in common with the other logging class -- could these be combined (or refactored so that logging of prevalence is written once)?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+
+    def did_not_run(self):
+        logger.debug(key="HSI_Wasting_GrowthMonitoring",
+                     data="HSI_Wasting_GrowthMonitoring: did not run"
+                     )
         pass
 
-    def on_birth(self, mother, child):
+
+class HSI_Wasting_SupplementaryFeedingProgramme_MAM(HSI_Event, IndividualScopeEventMixin):
```
tbhallett, line 1384: the two outpatient HSI are very similar -- do they need to be two entirely different classes?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+    """
+
+    def __init__(self, module):
+        """Runs only once, when simulation is initiated.
+        :param module: the module that created this event
+        """
+        super().__init__(module)
+        assert isinstance(module, Wasting)
+
+    def apply(self, population):
```
tbhallett, line 1199: why is this an event, rather  than a function that is called at initialise_simulation?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        df.at[child_id, 'un_am_treatment_type'] = 'not_applicable'
+        df.at[child_id, 'un_recov_to_mam_to_cancel'] = []
+        df.at[child_id, 'un_full_recov_to_cancel'] = []
+        df.at[child_id, 'un_progression_to_cancel'] = []
+        # df.at[child_id, 'un_last_nonemergency_appt_date']= pd.NaT
+
+        # initiate growth monitoring from day 1
+        self.sim.modules['HealthSystem'].schedule_hsi_event(
+            hsi_event=HSI_Wasting_GrowthMonitoring(module=self, person_id=child_id),
+            priority=2, topen=self.sim.date + pd.DateOffset(days=1)
```
tbhallett, line 368: is that what happens?? growth monitoring from day 1?

tbhallett (changes requested):
many thanks for all your work on this @EvaJanouskova 

Lots of really good stuff here (I especially liked the tests)

marghe-molaro:
Hi @EvaJanouskova, it would be really great if we could use this PR to trial the parameter labelling protocol proposed in PR #1610, which will be merged into master soon. Would you be willing to give it a go? Here's [the link](https://docs.google.com/presentation/d/1VvkJiwEUrSxjS9f49IU6KT5aVYFbNqK1qpvKxOSGmuY/edit?usp=sharing) to the presentation discussing it. If you have any questions/concerns please let me know!

timcolbourn:
> @timcolbourn, @tbhallett, Btw, we calibrated the number of direct deaths, i.e. those due to protein-energy malnutrition. However, we didn’t adjust the overall deaths in the TLO model. Since these deaths weren’t previously included, they should have been captured under 'other_causes', or something similar, which accounts for deaths from causes we don’t yet simulate, right?
> 
> So now, we should remove some deaths from 'other_causes', shouldn’t we?

I think 'other_causes' are calculated as the remainder from the total estimated deaths so should be automatically updated? @tbhallett ?

tbhallett:
> > @timcolbourn, @tbhallett, Btw, we calibrated the number of direct deaths, i.e. those due to protein-energy malnutrition. However, we didn’t adjust the overall deaths in the TLO model. Since these deaths weren’t previously included, they should have been captured under 'other_causes', or something similar, which accounts for deaths from causes we don’t yet simulate, right?
> > So now, we should remove some deaths from 'other_causes', shouldn’t we?
> 
> I think 'other_causes' are calculated as the remainder from the total estimated deaths so should be automatically updated? @tbhallett ?

Yes, it's done automatically based on the definitions of causes of deaths/dalys declared. This is why I asked this question:

https://github.com/UCL/TLOmodel/pull/1609#issuecomment-2931276668

EvaJanouskova:
@tbhallett, @tamuri, What should I do about the conflicts? I updated the RF_PriorityRanking_ALLPOLICIES files, but the warning about the conflicts is still there.

EvaJanouskova:
> @tbhallett, @tamuri, What should I do about the conflicts? I updated the RF_PriorityRanking_ALLPOLICIES files, but the warning about the conflicts is still there.

I found out what was wrong.

tests/test_wasting.py
EvaJanouskova: I'm afraid that with disable=True, I won't be able to check if the events were properly scheduled with find_events_for_person().

EvaJanouskova:
> Hi @sakshimohan, I merged the master branch, which required me to update the RFs with Priority Ranking for the policies.
> 
> I noticed that for the 'CVD' policy, most of the rankings for vulnerable groups were changed from -1 to 2, so I applied the same change to the Undernutrition_ treatments. Is that okay?
> 
> Also, there were some priority updates for the 'LCOA_EHP' policy, should I make any changes to the Undernutrition_ treatments as well?

Hi @sakshimohan,

I merged in the master branch again, and it seems that for the 'CVD' policy, the rankings for vulnerable groups were changed back from 2 to -1. Is it correct for me to do the same?

I’m also not sure what happened with the 'LCOA_EHP' policy, perhaps those changes were reverted as well?

Could you please check whether the policies in this branch for `Undernutrition_` treatments make sense? You can find descriptions of these treatments in the [Wasting write-up:](https://github.com/user-attachments/files/21585401/Wasting.docx)
- for `Undernutrition_GrowthMonitoring`: see Routine growth monitoring;
- `Undernutrition_Feeding_Supplementary`: Supplementary feeding programme (SFP),
- `Undernutrition_Feeding_Outpatient`: Outpatient Therapeutic Programme (OTP),
-  `Undernutrition_Feeding_Inpatient`: Intensive inpatient therapeutic care (ITC)).

tests/test_wasting.py
EvaJanouskova: done:commit [test_wast: more informative name of test]

tests/test_wasting.py
EvaJanouskova: done:
commit [test_wast: comment instead of explicit zero-duration definition]

tests/test_wasting.py
EvaJanouskova: done:
commit [test_wast: use df.at for consistency in single-cell read/write]

tests/test_wasting.py
EvaJanouskova: done:
commit [test_wast: use df.at for consistency in single-cell read/write]

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        """Prepares for simulation. Schedules:
+        * the first growth monitoring to happen straight away, scheduled monthly to detect new cases for treatment.
+        * the main incidence polling event.
+        * the main logging event.
+        """
+
+        sim.schedule_event(Wasting_InitLoggingEvent(self), sim.date)
+        sim.schedule_event(Wasting_InitiateGrowthMonitoring(self), sim.date)
+        sim.schedule_event(Wasting_IncidencePoll(self), sim.date + DateOffset(months=1))
+        sim.schedule_event(Wasting_LoggingEvent(self), sim.date + DateOffset(years=1) - DateOffset(days=1))
```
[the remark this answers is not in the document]
EvaJanouskova: done:
commit [wast: logging event updated to run on January 1st each year]

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+
+        # Retrieve the consumables codes and amounts of the consumables used
+        self.cons_codes = self.get_consumables_for_each_treatment()
+
+    def on_birth(self, mother_id, child_id):
```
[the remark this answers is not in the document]
EvaJanouskova: Same as above, these commented-out lines are just to note that those are at this moment set to NaT to better understand what is going on there when reading the code. Do you want me to delete them anyway?

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+                                          p=[prob_normal_whz_with_muac_less_than_115mm,
+                                             prob_normal_whz_with_muac_between_115and125mm,
+                                             prob_normal_whz_with_muac_over_125mm]),
+                axis=1
+            )
+
+    def nutritional_oedema_present(self, idx):
+        """
+        This function applies the probability of nutritional oedema present in wasting and non-wasted cases
+        :param idx: index of children under 5, or person_id
```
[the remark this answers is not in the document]
EvaJanouskova: done:
commit [wast & test_wast: updated treatment parameter naming; docstrings improved]

src/tlo/methods/wasting.py
EvaJanouskova: done:
commit [wast: create a daly-weights lookup for efficiency]

src/tlo/methods/wasting.py
EvaJanouskova: It was because, this doesn't mean they are on treatment, it just means that they were already scheduled for enrolment for the treatment today, but are not enrolled yet. I will do it differently, thanks.

EvaJanouskova: done:
commit [wast: add logic to track acute malnutrition being already assessed today (avoid use of find_events_for_person)]

EvaJanouskova:
TODO:
https://github.com/UCL/TLOmodel/issues/1431#issuecomment-3246222922

src/tlo/methods/wasting.py
EvaJanouskova: done:
commit [wast: determining if patient will attend to growth monitoring moved to __init__]

src/tlo/methods/wasting.py
EvaJanouskova: Can this still be done, given that we now determine the attendance in `__init__` based on age? If I just re-use `self`, it won't update the attendance for the person, will it?

src/tlo/methods/wasting.py
EvaJanouskova: done:
commit [wast: rm redundant pass statements in did_not_run methods]

src/tlo/methods/wasting.py
EvaJanouskova: done:
commit [wast: initialise HSI event footprint using built-in function]

src/tlo/methods/wasting.py
EvaJanouskova: According to `ResourceFile_EPI_WHO_estimates/vaccine_schedule.csv`, the first immunisation appointment is on day one, during which growth monitoring should take place.

src/tlo/methods/wasting.py
EvaJanouskova: done:
commit [wast: join init and annual loggings]

src/tlo/methods/wasting.py
EvaJanouskova: As discussed in the #wasting channel on Slack: 

The reason we implemented it this way was to avoid overwhelming the HS. If we model the progression from moderate to severe wasting at the population level rather than the individual level, allowing everyone to progress 'en masse' means symptoms are recognised simultaneously, which could trigger mass care-seeking and overwhelm the HS.

However, a similar issue might arise from the fact that the incidence of moderate wasting is also modelled at the population level. This might be manageable under the SQ, as relatively few children would become SAM with incidence. But with an intervention that increases symptom awareness, almost all MAM cases will seek care, again leading to a mass care-seeking at the beginning of the month.

In another comment on this PR, you suggested randomising the onset of moderate wasting instead of setting it all on the first day of the month. I understood this was to avoid synchronised progression and care-seeking. But if we change the progression to population level, we’re back to the same issue, everyone progresses on the same date, so the randomised onset would no longer have any effect.

Currently, we decide who and when will progress within the population-level event, but the progression itself is then issued individually, for the reasons outlined above.

Tim C. suggested we leave it as is. Let us know if you strongly feel we should still switch to population-level progression, ie whether the preference is to prioritise model speed over avoiding HS overload.

(I also checked how it’s done in the HIV module, and the onsets are handled at the individual level there as well.)

EvaJanouskova:
@tbhallett, @timcolbourn, I believe I’ve now responded to all comments. When you have a moment, would you mind doing a second round of review?

.gitattributes
tbhallett, line 1: @tamuri - do we want this change to come in?

tests/test_wasting.py
```diff
@@ -0,0 +1,1065 @@
...
+
+    # Run the created instance of HSI_Wasting_OutpatientTherapeuticProgramme_SAM
+    sam_ev = [ev[1] for ev in sim.modules['HealthSystem'].find_events_for_person(person_id)
+              if isinstance(ev[1], HSI_Wasting_OutpatientTherapeuticProgramme_SAM)][0]
+    sam_ev.run(squeeze_factor=0.0)
+
+    # Check death event is scheduled for another day than natural death was scheduled for since there is no recovery
+    # with treatment
+    assert isinstance(sim.find_events_for_person(person_id)[0][1], Wasting_SevereAcuteMalnutritionDeath_Event)
+    assert df.at[person_id, 'un_sam_death_date'] != nat_death_date
```
tbhallett, line 728: should this check that the new date is LATER

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
         df.loc[df.is_alive, 'un_ever_wasted'] = False
+        df.loc[df.is_alive, 'un_WHZ_category'] = 'WHZ>=-2'  # not undernourished
+        # df.loc[df.is_alive, 'un_last_wasting_date_of_onset'] = pd.NaT
+        df.loc[df.is_alive, 'un_clinical_acute_malnutrition'] = 'well'
+        df.loc[df.is_alive, 'un_am_nutritional_oedema'] = False
+        df.loc[df.is_alive, 'un_am_MUAC_category'] = '>=125mm'
+        # df.loc[df.is_alive, 'un_sam_death_date'] = pd.NaT
+        # df.loc[df.is_alive, 'un_am_recovery_date'] = pd.NaT
+        # df.loc[df.is_alive, 'un_am_discharge_date'] = pd.NaT
+        # df.loc[df.is_alive, 'un_am_tx_start_date'] = pd.NaT
```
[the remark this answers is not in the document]
tbhallett: in general --- yes, please remove the commented-out stuff. I think commented-out code is very ambiguous and confusing (unless it's explained what is being done and why).

But, I do not understand what you are saying in your response. You have not declared the value to be taken in intitial population for these properties (e.g. un_last_wasting_date_of_onset). You seem instead to be relying on the defaulting pandas behaviour of the null type when we declare the series to have a certain data type. Elsewhere we always prefer to be EXPLICIT and declare the value that we want. This guards against changes in the pandas default settings and also makes it readable to everyone. This is what we do it in all modules. So I would prefer that all initialising here (and in on_birth) is 100% explicit. So that would be uncommenting these lines.

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        df.loc[df.is_alive, 'un_am_treatment_type'] = 'not_applicable'
+        df.loc[df.is_alive, 'un_recov_to_mam_to_cancel'] = \
+            df.loc[df.is_alive, 'un_recov_to_mam_to_cancel'].apply(lambda x: [])
+        df.loc[df.is_alive, 'un_full_recov_to_cancel'] = \
+            df.loc[df.is_alive, 'un_full_recov_to_cancel'].apply(lambda x: [])
+        df.loc[df.is_alive, 'un_progression_to_cancel'] = \
+            df.loc[df.is_alive, 'un_progression_to_cancel'].apply(lambda x: [])
+        # df.loc[df.is_alive, 'un_last_nonemergency_appt_date']= pd.NaT
```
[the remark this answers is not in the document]
tbhallett: see comment above

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+                # schedule next growth monitoring
+                self.sim.modules['HealthSystem'].schedule_hsi_event(
+                    hsi_event=HSI_Wasting_GrowthMonitoring(module=self.module, person_id=person_id),
+                    topen=self.sim.date + pd.DateOffset(days=person_monitoring_frequency),
+                    tclose=None,
+                    priority=2
+                )
+            # else: no more growth monitoring scheduled as the age will be above 5 at the time
+
+        # TODO: as stated above, for now we schedule next monitoring for all children, even those sent for treatment
```
tbhallett, line 1400: to do resolved now?

tbhallett (commented):
first set of comments

src/tlo/methods/wasting.py
```diff
@@ -1,41 +1,1947 @@
...
+        # TODO: Will they be monitored during the treatment? Can we assume, that after the treatment they will be
+        #  always properly checked (all measurements and oedema checked), or should be the assumed "treatment outcome"
+        #  be also based on equipment availability and probability of checking oedema? Maybe they should be sent for
+        #  after treatment monitoring, where the assumed "treatment outcome" will be determined and follow-up treatment
+        #  based on that? - The easiest way (currently coded) is assuming that after treatment all measurements are
+        #  done, hence correctly diagnosed. The growth monitoring is scheduled for them as usual, ie, for instance, for
+        #  a child 2-5 old, if they were sent for treatment via growth monitoring, they will be on tx for adequate nmb
+        #  of weeks, but next monitoring will be done in ~5 months after the treatment. - Or we could schedule for the
+        #  treated children a monitoring sooner after the treatment.
+        # no
```
tbhallett, line 1362: this looks like long note to self.  Is it covered in the methods document? Can we remove it? It seems odd to leave the thinking-allowed here.

docs/write-ups/Wasting.docx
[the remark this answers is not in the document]
EvaJanouskova: done (hope this is better):
commit [Wast write-up: cleaned and clarified (removed comments, strike throughs, TODOs, and bathtub-related text; improved clarity of calibration section)]

PR #1617, opened 2025-05-08
Implement clinics in TLO

sangeetabhatia03:


sangeetabhatia03:
In the future, we might want to separate the rescaling of fungible/non-fungible capabilities. For now, only fungible time is rescaled.  TODO: Discuss with group.

sangeetabhatia03:
hi @marghe-molaro  as discussed, here is my plan for next steps:
1. Edit HSIEventQueueItem to include a property called clinic_access (boolean).
2. Define a getter function in HealthSystem:
```
def get_clinic_eligibility(str)
 """ Return true if str is one of the keys of the clinics dictionary"""
```
2. Edit _add_hsi_event_queue_item_to_hsi_event_queue as follows: before creating HSIEventQueueItem, check for clinic eligibility using hsi_event.module. That is
```
clinic_access = get_clinic_eligibility(hsi_event.module)
```
and set the value in HSIEvent appropriately.

marghe-molaro:
Looks good @sangeetabhatia03 - in the last step ("and set the value in HSIEvent appropriately.") I think you mean HSIEventQueueItem

sangeetabhatia03:
This PR implements the concept of clinics. This is done by splitting the daily capabilities into fungible and non-fungible buckets according to the split specified in the new resource file `ResourceFile_Clinics.csv`.
Notes
- clinic capabilities are expanded in the same way as fungible capabilities throughout.
- Note that for now we assume that clinic capabilities are only specified at levels below 2; so a level 2 facility in the resource file will trigger an error.
- logger has been amended to also output the clinic at which HSI runs.

All healthsystem tests pass for me locally.

tbhallett:
> All healthsystem tests pass for me locally.

They don't seem to pass on GitHub Actions. Are there uncommitted changes for you locally?

tbhallett (changes requested):

tbhallett:
Thanks @sangeetabhatia03 

Great to see this working. Overall comments:

- [ ] I think we need a place were the design / logic / specification for the new clinics Resource File is carefully explained.
- [ ] You may think it pedantic, but it would really help future readers if the terms we used as 100000% consistent throughout.
- [ ] The changes to HealthSystem basically make there be two tracks of logic (fungible and clinic). But the fungible is really just a special case of clinic (one for which is ring-fenced for all those HSI not allocated specifically to another clinic). So, it would seem simpler and less duplicative to use that observation to make there still be only ONE set of tracking variables. I can't insist on this, as it obviously works as is, but I think out future selves would appreciate that help.
- [ ] The logic that you're aiming for (I believe) is that the "fungible" resources are ring-fenced for all "OTHER" HSIs that are not specifically going to be associated with a "Clinic". In that sense, it's not really "fungible" in the full sense of the word (but it can't be accessed by HSI for which their own clinic time has run out. I think that's the right design choice.... but the wording I think will confuse us). It's just operating as a Clinic with ring-fencing too, but just for "Other"  HSIs.
- [ ] Tests are good, but I think need more explanation and tightening up on the commenting; 
- [ ] Also tests don't seem to address all the upstream processing of the ResourceFile (as the test overwrite the results of that), so I think we need another test to make sure that the RF is read in and processed correctly.

src/tlo/methods/healthsystem.py
```diff
@@ -339,6 +340,10 @@ class HealthSystem(Module):
         'use_funded_or_actual_staffing_postSwitch': Parameter(
             Types.STRING, 'Staffing availability after switch in `year_use_funded_or_actual_staffing_switch`. '
                           'Acceptable values are the same as those for Parameter `use_funded_or_actual_staffing`.'),
+        'include_clinics': Parameter(
```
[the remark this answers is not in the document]
sangeetabhatia03: Included via commit 58d5707

src/tlo/methods/healthsystem.py
```diff
@@ -531,6 +541,7 @@ def __init__(
                 "hsi_event_count_log_period argument should be one of 'day', 'month' "
                 "'year', 'simulation' or None."
             )
+        self.include_clinics = include_clinics
```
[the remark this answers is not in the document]
sangeetabhatia03: Apologies! Deleted.

src/tlo/methods/hsi_event.py
```diff
@@ -362,7 +365,7 @@ def _check_if_appt_footprint_can_run(self) -> bool:
         """Check that event (if individual level) is able to run with this configuration of officers (i.e. check that
         this does not demand officers that are _never_ available), and issue warning if not.
         """
-        if self.healthcare_system._officers_with_availability.issuperset(
+        if self.healthcare_system._officers_with_availability[self.eligible_clinic].issuperset(
```
marghe-molaro: If eligible_clinic were to be moved to HSIEventQueueEvent this check would have to be moved to when item is added to the queue

src/tlo/methods/healthsystem.py
```diff
@@ -2060,7 +2241,8 @@ def run_individual_level_events_in_mode_1(self,
             for ev_num, event in enumerate(_list_of_individual_hsi_event_tuples):
                 _priority = event.priority
                 event = event.hsi_event
-                squeeze_factor = squeeze_factor_per_hsi_event[ev_num]                  # todo use zip here!
+                clinic = self.get_clinic_eligibility(event)
```
marghe-molaro: Here is crux of the matter: it would make sense for clinic eligibility to be an attribute of `event` here (just as `priority` is) as this is where it is needed again and again.
Speaking of: I'm confused as to why the function `self.get_clinic_eligibility` works with `event` here (which is a HSIEventQueueItem) when it was defined as only accepting `HSI_Event` as an input (which here would be `event.hsi_event`)??

src/tlo/methods/healthsystem.py
```diff
@@ -637,6 +668,30 @@ def read_parameters(self, resourcefilepath: Optional[Path] = None):
         # Ensure that a value for the year at the start of the simulation is provided.
         assert all(2010 in sheet['year'].values for sheet in self.parameters['yearly_HR_scaling'].values())
 
+
+    def validate_clinic_configuration(self, clinic_capabilities_df: pd.DataFrame):
```
marghe-molaro: I think we also need to double check that the clinics defined in the two separate resource files match. Is this checked somewhere else atm?

sangeetabhatia03:
> Hi @sangeetabhatia03,
> 
> Great job on this PR, I think looks basically ready! I have pointed out a few things that need tidying up before it can be merged. I would also add that the default ResourceFile_ClinicConfigurations doesn't really seem to require an "OtherClinic" clinic column. (I would also vote to rename "OtherClinic", which sounds a bit strange in a scenario where to other clinics are present, to something like "GenericClinic".
> 
> We could have a chat about these comments over a meeting tomorrow?

Renamed OtherClinic to GenericClinic #e233df1b9bd429b0f5e2769ddcc7a61fccfb1903

src/tlo/methods/healthsystem.py
```diff
@@ -1611,6 +1759,21 @@ def get_appt_footprint_as_time_request(self, facility_info: FacilityInfo, appt_f
     def get_squeeze_factors(self, footprints_per_event, total_footprint, current_capabilities,
                             compute_squeeze_factor_to_district_level: bool
                             ):
+        ## TODO: check if there is a better place to intitalise this store
```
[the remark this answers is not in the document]
sangeetabhatia03: Move it somehwere in pre_initialise_population

src/tlo/methods/healthsystem.py
sangeetabhatia03: i had not set up get_clinic_eligibility to expect HSIEventQueueItem; I think this code works because I only use the TREATMENT_ID property, which the passed object has. I will now amend  get_clinic_eligibility to expect HSIEventQueueItem only.

src/tlo/methods/healthsystem.py
```diff
@@ -1373,8 +1516,9 @@ def _add_hsi_event_queue_item_to_hsi_event_queue(self, priority, topen, tclose,
         else:
             rand_queue = self.hsi_event_queue_counter
 
+        clinic_eligibility = self.get_clinic_eligibility(hsi_event)
```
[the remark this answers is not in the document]
sangeetabhatia03: Done.

sangeetabhatia03:
@tbhallett this PR is now ready for your review.

tests/test_healthsystem.py
```diff
@@ -1906,14 +2036,246 @@ def apply(self, person_id, squeeze_factor):
...
+            self.TREATMENT_ID = treatment_id
+            self.EXPECTED_APPT_FOOTPRINT = self.make_appt_footprint({appt_type: 1})
+            self.ACCEPTED_FACILITY_LEVEL = level
+
+            self.this_hsi_event_ran = False
+
+        def apply(self, person_id, squeeze_factor):
+            self.this_hsi_event_ran = True
+
+    def create_simulation(tmpdir: Path, tot_population) -> dict:
```
tbhallett: function return Sim objects not dict, so type hint should be `-> Simulation`

tests/test_healthsystem.py
```diff
@@ -1906,14 +2036,246 @@ def apply(self, person_id, squeeze_factor):
...
+    other_clinic = sim.modules["HealthSystem"]._daily_capabilities["GenericClinic"]
+    clinic1 = sim.modules["HealthSystem"]._daily_capabilities["Clinic1"]
+
+    # 'FacilityID_20_Officer_DCSA' is coming from the resource file
+    ratio = clinic1["FacilityID_20_Officer_DCSA"] / other_clinic["FacilityID_20_Officer_DCSA"]
+    expect = 0.6 / 0.4
+    assert abs(ratio - expect) < 1e-7, "GenericClinic capabilities are not split correctly"
+
+    # Schedule an identical appointment for all individuals, assigning clinic as follows:
+    # half individuals have clinic_eligibility=GenericClinic and half clinic_eligibility=Hiv
```
tbhallett: Hiv is noted as the eligibility for the clinics, but the options for clinic actually appear to be: "Clinic1", "GenericClinic"

src/tlo/methods/healthsystem.py
```diff
@@ -528,123 +571,205 @@ def __init__(
...
+        self.parameters["clinic_configuration"] = pd.read_csv(filepath)
+        filepath = (
+            path_to_resourcefiles_for_healthsystem
+            / "human_resources"
+            / "clinics"
+            / "ResourceFile_ClinicMappings"
+            / f"{self.parameters['clinic_configuration_name']}.csv"
+        )
+
+        self.parameters["clinic_mapping"] = pd.read_csv(filepath)
```
tbhallett: clnic_mapping isn't declared as a parameter

src/tlo/methods/healthsystem.py
```diff
@@ -1348,11 +1553,24 @@ def schedule_hsi_event(
             # The HSI is allowed and will be added to the HSI_EVENT_QUEUE.
             # Let the HSI gather information about itself (facility_id and appt-footprint time requirements):
             hsi_event.initialise()
+            ## Create a dummy item so that we can query the clinic eligibility
+            clinic_eligibility = "GenericClinic"
+            dummy_item: HSIEventQueueItem = HSIEventQueueItem(
+                clinic_eligibility, priority, topen, 1, 1, tclose, hsi_event
+            )
+            clinic_eligibility = self.get_clinic_eligibility(dummy_item)
```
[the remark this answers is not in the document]
marghe-molaro: Clinic eligibility should be an attribute of HSIEventQueueItem, not HSI_Event, because (like priority) it is something relevant to the HealthSystem and its queuing system, not the HSI itself. It should however be estimated from the HSI_Event and HealthSystem structure, and passed to the HSIEventQueueItem upon initialisation, not the other way around as done here.

tbhallett (approved):

sangeetabhatia03:
Wiki page is here https://github.com/UCL/TLOmodel/wiki/Clinics-implementation-in-TLO

Issue #1601, opened 2025-02-28
Make Fast Version of Symptom Manager


tbhallett:
In profiling, a lot of time is taken in `SymptomManager`. This is because it is designed to do a lot of checking (that the symptom exists, and the disease is allowed to apply it, and when its resolved its only resolved by the disease that caused it). 

Whilst these checks are useful in development, now that we never have problems, we can dispense with these checks. 

We should we create a drop-in replacement for `SymptomManager` that has the same functionality and interface, but internally does the simplest and quickest thing possible. This could be making all the symptoms columns `bools`

PR #1626, opened 2025-06-03
Symptom Bookkeeping

thewati:
Bookkeeping with pandas column or dictionary. Fixes issue #1601


src/tlo/methods/symptommanager.py
```diff
@@ -418,6 +427,16 @@ def change_symptom(self, person_id, symptom_string, add_or_remove, disease_modul
...
 
+            # Update symptom tracker. Remove if no other module is causing this symptom.
+            for pid in person_id:
+                for sym in symptom_string:
+                    symptom_col = self.get_column_name_for_symptom(sym)
+                    if self.bsh.is_empty(pid, columns=symptom_col):
+                        self.symptom_tracker[pid].discard(sym)
+                        # Delete person’s entry entirely if it becomes empty
+                        if not self.symptom_tracker[pid]:
+                            del self.symptom_tracker[pid]
```
tbhallett, line 438: I wonder if we even need to do the delete of the person

src/tlo/methods/symptommanager.py
```diff
@@ -396,6 +400,11 @@ def change_symptom(self, person_id, symptom_string, add_or_remove, disease_modul
             self.bsh.set(person_id, disease_module.name, columns=sy_columns)
             self._persons_with_newly_onset_symptoms = self._persons_with_newly_onset_symptoms.union(person_id)
 
+            # Update symptom tracker
+            for pid in person_id:
+                for sym in symptom_string:
```
tbhallett, line 405: Rather than looping through symptoms, might be that doing the union operation of the current set of symptoms with the new set of symptoms is faster

src/tlo/methods/symptommanager.py
```diff
@@ -331,6 +332,9 @@ def on_birth(self, mother_id, child_id):
         for property in self.PROPERTIES:
             df.at[child_id, property] = 0
 
+        # Clear bookkeeping dictionary
+        self.symptom_tracker.pop(child_id, None)
```
tbhallett, line 336: I don't think we need this. The child id will not be in our tracker when they are born, right?

src/tlo/methods/symptommanager.py
```diff
@@ -583,6 +602,16 @@ def clear_symptoms(self, person_id: Union[int, Sequence[int]], disease_module: M
...
         self.bsh.unset(person_id, disease_module.name, columns=sy_columns)
 
+        # Update bookkeeping
+        for pid in person_id:
+            for sym in self.symptom_names:
+                symptom_col = self.get_column_name_for_symptom(sym)
+                if self.bsh.is_empty(pid, columns=symptom_col):
+                    self.symptom_tracker[pid].discard(sym)
+            # Clean up if person has no symptoms left
+            if not self.symptom_tracker[pid]:
```
tbhallett, line 612: as above, no real need to delete the key for the person when they have no symptoms...?

src/tlo/methods/symptommanager.py
```diff
@@ -601,6 +630,10 @@ def get_persons_with_newly_onset_symptoms(self):
     def reset_persons_with_newly_onset_symptoms(self):
         self._persons_with_newly_onset_symptoms.clear()
 
+    def get_current_symptoms(self, person_id: int) -> Set[str]:
```
tbhallett, line 633: This looks good. 

So we should call this from "has what" for the special case of when it's a request for the symptoms for one person irrespective of disease module.

tbhallett (commented):
Great work. It almost done already!

In addition, we discussed what to do when persons die and whether the demography module should inform this module so that it can delete the key for that person. 

That might not be wholly necessary,depending on whether or not we let this new function be called directly versus after the checking for is alive already included in has-what, or whether we offer no guarantees about alive status, or whether it's very quick and cheap to delete the key for the person. TBC

Interested to see profiling!

src/tlo/methods/symptommanager.py
thewati: I was of the view that we do not need to accumulate empty sets for people with no symptoms. Is there any reason why we would want to keep people without any symptoms in the tracker?

tbhallett: Just thought the delete operation might be time consuming. But agree that it useful. Maybe keep it in and we can see if it's a pinch point?

tamuri: I think it'll be fine - we should remove it

thewati:
Below are approximate improvements after a little profiling:

| Metric  | Master Version | Bookkeeping Version | Improvement
| ------------- | ------------- | ------------- | ------------- |
| Tottime  | 10.9 μs  | 8.9 μs  | ~ 18% faster |
| Cumtime  | 20 μs  | 17.1 μs  | ~ 15% faster |

More in the attached word document below:
[SymptomManager Profiling Visualisation.docx](https://github.com/user-attachments/files/20611536/SymptomManager.Profiling.Visualisation.docx)

tbhallett:
> Below are approximate improvements after a little profiling:
> 
> Metric	Master Version	Bookkeeping Version	Improvement
> Tottime	10.9 μs	8.9 μs	~ 18% faster
> Cumtime	20 μs	17.1 μs	~ 15% faster
> More in the attached word document below: [SymptomManager Profiling Visualisation.docx](https://github.com/user-attachments/files/20611536/SymptomManager.Profiling.Visualisation.docx)

That looks promising, @thewati.

SO, we get a juicy speed-up... though its not mind-blowing.

Looks like you run in on a very small population (given the short times). I wonder if this kind of thing needs a big population and many (50) symptoms to see the benefits of it? What script did you use for it? (Nothing else pushed to this branch so far)

tbhallett:
> I have pushed a script for profiling `scripts/profiling/sympom_manager_bookkeeping_profiling.py`. The results below are after using a population 1,000,000 (its different in the pushed file), a total of 50 symptoms and running the tests for 1,000,000 times:
> 
> SymptomManager Version	Total Time (seconds)	Avg Time per `has_what()` Call (seconds)
> Bookkeeping	5.4675	0.000005
> Original (No Bookkeeping)	272.5963	0.000273

FANTASTIC!!!!

What a great result.

tbhallett:
Are we happy with this to merge now, or do we want any more experiments @thewati  and @tamuri ?

thewati:
> Are we happy with this to merge now, or do we want any more experiments @thewati and @tamuri ?

@tbhallett the code is not clearing symptoms for dead people in the demography module yet. I will make a push that deals with that soon, then it should be ready for review

src/tlo/methods/symptommanager.py
```diff
@@ -600,6 +630,21 @@ def get_persons_with_newly_onset_symptoms(self):
...
+        """Get the current symptoms for a person. Works with bookkeeping dictionary"""
+        return self.symptom_tracker.get(person_id, set())
+
+    def clear_symptoms_for_deceased_person(self, person_id: int):
+        """
+        Clears symptoms by deleting the dead person's ID in the tracker
+        """
+
+        # Remove person from tracker entirely
+        self.symptom_tracker.pop(person_id, None)
```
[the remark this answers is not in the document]
thewati: > 


I tried to remove the parameter "None" as suggested. However, this causes a Key Error in `test_symptommanager.py` when a deceased person has no entry in the symptom_tracker. Since `clear_symptoms_for_deceased_person` can be called even if the person had no symptoms, I was thinking of keeping the parameter "None" to avoid the crash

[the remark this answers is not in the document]
tbhallett: Ah ok. so it's doing the same job as 
```
if person_id in self.symtom_tracker:
    self.symptom_tracker.pop(person_id)
```

I see -- fine to leave it then!

thewati:
@tbhallett and @tamuri, I believe I have handled all the requested changes. You may approve if satisfied

tamuri (approved):
Thanks, Wati, looks good!

Issue #1624, opened 2025-05-28
Integrating DFF into TLO


pchitsulo:
The aim is to assess the impact of the Direct Facility Financing (DFF), specifically in terms of health outcomes, service delivery, access, quality, and efficiency.

Tasks

- [ ] Update TLO parameters using data extracted from DHIS2 as seen in [Table 5](https://docs.google.com/document/d/1a4CDKWD5UpdCD0UrqoDNhptDv32se3afRDK3EXPoYHU/edit?tab=t.0) 
- [ ] Run the 1 intervention (RMCH, vaccines, …) module to assess impact in DALYs averted, HCW time, and PFT with and without DFF changes in ST, MT, LT

thewati:
Project shelved since a different economic analysis was recommended

PR #1627, opened 2025-06-04
DFF Integration into TLO

pchitsulo:
Updating parameters. Fixes issue #1624


thewati:
Project shelved since a different economic analysis was recommended

PR #1633, opened 2025-06-10
Allow consumable availability to be overridden for a given Treatment ID

joehcollins:
This PR adds functionality allowing modellers to define a single/set of Treatment IDs for which consumable availability is set at 100%. I've taken the following approach:

1. Added a new parameter to the health system - `cons_override_treatment_ids` - which can contain treatment ids for which cons. availability should be set at 100%
2. Added an optional parameter to the function `_request_consumables` within `hsi_event.py` to which the `cons_override_treatment_ids` list can be passed
3. Added a boolean parameter to `_lookup_availability_of_consumables` in `consumables.py` if the relevant treatment id is listed (True)
4. Set all cons. to True within `_lookup_availability_of_consumables` is new function parameter set to True - this preserves logging etc and will correctly capture the use of the consumables but ignores probability of availability.
5. Added a couple of lines of testing to `test_use_get_consumables_by_hsi_method_get_consumables` in `test_consumables.py`

Happy for any comments/changes people want to make (e.g. do we also want to be able to set cons availability to 0 for a certain treatment id).

### DIFF
--- a/src/tlo/methods/consumables.py
+++ b/src/tlo/methods/consumables.py
@@ -202,7 +202,8 @@ class Consumables:
                              essential_item_codes: dict,
                              optional_item_codes: Optional[dict] = None,
                              to_log: bool = True,
-                             treatment_id: Optional[str] = None
+                             treatment_id: Optional[str] = None,
+                             override_hsi: Optional[list] = None,
                              ) -> dict:
         """This is a private function called by 'get_consumables` in the `HSI_Event` base class. It queries whether
         item_codes are currently available at a particular Facility_ID and logs the request.
@@ -212,6 +213,7 @@ class Consumables:
         :param optional_item_codes: dict of the form {<item_code>: <quantity>} for the optional items requested
         :param to_log: whether the request is logged.
         :param treatment_id: the TREATMENT_ID of the HSI (which is entered to the log, if provided).
+        :param override_hsi: list of treatment IDs for which consumable availability is set at 100%.
         :return: dict of the form {<item_code>: <bool>} indicating the availability of each item requested.
         """
         # If optional_item_codes is None, treat it as an empty dictionary
@@ -223,8 +225,12 @@ class Consumables:
         if len(not_recognised_item_codes) > 0:
             self._not_recognised_item_codes[treatment_id] |= not_recognised_item_codes
 
-        # Look-up whether each of these items is available in this facility currently:
-        available = self._lookup_availability_of_consumables(item_codes=_all_item_codes, facility_info=facility_info)
+        # Check if the availability of consumables for this treatment id has been overridden
+        avail_overridden = False if override_hsi is None else True if treatment_id in override_hsi else False
+
+        # Look-up whether each of these items is available in this facility currently.:
+        available = self._lookup_availability_of_consumables(item_codes=_all_item_codes, facility_info=facility_info,
+                                                             avail_overridden=avail_overridden)
 
         # Log the request and the outcome:
         if to_log:
@@ -255,7 +261,8 @@ class Consumables:
 
     def _lookup_availability_of_consumables(self,
                                             facility_info: 'FacilityInfo',  # noqa: F821
-                                            item_codes: dict
+                                            item_codes: dict,
+                                            avail_overridden: bool,
                                             ) -> dict:
         """Lookup whether a particular item_code is in the set of available items for that facility (in
         `self._is_available`). If any code is not recognised, use the `_is_unknown_item_available`."""
@@ -270,11 +277,18 @@ class Consumables:
             else:
                 return {_i: False for _i in item_codes}
 
-        for _i in item_codes.keys():
-            if _i in self.item_codes:
-                avail.update({_i: _i in self._is_available[facility_info.id]})
-            else:
-                avail.update({_i: self._is_unknown_item_available[facility_info.id]})
+        # If availability is overridden for this treatment id then all items will be set as available.
+        if avail_overridden:
+           for _i in item_codes.keys():
+               avail.update({_i: True})
+
+        else:
+            for _i in item_codes.keys():
+                if _i in self.item_codes:
+                    avail.update({_i: _i in self._is_available[facility_info.id]})
+                else:
+                    avail.update({_i: self._is_unknown_item_available[facility_info.id]})
+
         return avail
 
     def on_simulation_end(self):
--- a/resources/healthsystem/ResourceFile_HealthSystem_parameters.csv
+++ b/resources/healthsystem/ResourceFile_HealthSystem_parameters.csv
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:f1cb38ba76c5673855e2e17e28ad1d36b5cec07d5d7872a7d6bc8aafde0f7009
-size 828
+oid sha256:3ab2548279c7677d2336c8a17a5a1e6bd0391787a2d60bee2b66e254e4a5f175
+size 862
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -191,6 +191,11 @@ class HealthSystem(Module):
             " When using 'all' or 'none', requests for consumables are not logged. NB. This parameter is over-ridden"
             "if an argument is provided to the module initialiser."
             "Note that other options are also available: see the `Consumables` class."),
+        'cons_override_treatment_ids': Parameter(
+            Types.LIST,
+            "Consumable availability within any treatment ids listed in this parameter will be set at 100%. "
+            "By default this list is empty"),
+
 
         # Infrastructure and Equipment
         'BedCapacity': Parameter(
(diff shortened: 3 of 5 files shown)


tbhallett (changes requested):
Good idea. Implementation seems very sensible. Some suggestions for readability/organisation here.

src/tlo/methods/consumables.py
```diff
@@ -202,7 +202,8 @@ def _request_consumables(self,
                              essential_item_codes: dict,
                              optional_item_codes: Optional[dict] = None,
                              to_log: bool = True,
-                             treatment_id: Optional[str] = None
+                             treatment_id: Optional[str] = None,
+                             override_hsi: Optional[list] = None,
```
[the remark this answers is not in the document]
joehcollins: Can you elaborate? Currently we do pass the whole list of treatment ids to this function in line 231 of hsi_event.py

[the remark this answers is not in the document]
joehcollins: See what you mean. I think i did it that way so that the list of treatment ids can be updated following initialisation (after burn in period)

tbhallett:
I think the pattern to follow is how availability is set and reset.

joehcollins:
@EvaJanouskova did have a go at this but have hit a bit of a wall with some of the test. will pick up next week, happy to chat then

EvaJanouskova (approved):
This looks good to me.

EvaJanouskova (changes requested):
Looks good to me, just needs the tox check to be fixed.

tbhallett (approved):

src/tlo/methods/consumables.py
```diff
@@ -309,51 +287,37 @@ def _request_consumables(self,
...
-                          forced_avail: Optional[bool] = None) -> dict:
-        """Returns a dictionary with availability of item codes when availability probability is being overridden. If
-        set_avail is not predetermined as True/False then availability is determined by
-         _treatment_ids_overridden_avail"""
-
-        if forced_avail is not None:
-            return {_i: forced_avail for _i in item_codes}
-        else:
-            check_avail = self._rng.random_sample() < self._treatment_ids_overridden_avail
-            return {_i: check_avail for _i in item_codes}
```
EvaJanouskova: Hi @tbhallett and @joehcollins, Wasn't it before that the overridden probability was deciding whether all items are available, ie it being the probability of treatment being provided:
```
            check_avail = self._rng.random_sample() < self._treatment_ids_overridden_avail
            return {_i: check_avail for _i in item_codes} 
```

src/tlo/methods/consumables.py
```diff
@@ -309,51 +287,37 @@ def _request_consumables(self,
...
-                return self._return_available(item_codes, forced_avail=True)
+                return {_i: True for _i in item_codes}
             else:
-                # returns false for all item codes
-                return self._return_available(item_codes, forced_avail=False)
+                return {_i: False for _i in item_codes}
+
+        if override_probability is not None:
+            # The probability of these items being available is being overriden
+            return dict(zip(item_codes, self._rng.random_sample(len(item_codes)) < override_probability))
```
EvaJanouskova, line 311: while now it is setting the probability for each item requested for the treatment?
```
        if override_probability is not None:
            # The probability of these items being available is being overriden
            return dict(zip(item_codes, self._rng.random_sample(len(item_codes)) < override_probability))
```

EvaJanouskova (commented):
@tbhallett, seems you have changed the behaviour of the overridden probability, was it intentional?
Old behaviour: The overridden probability determined the availability of the entire treatment as a whole.
New behaviour: The overridden probability determines the availability of each item within the treatment individually.

EvaJanouskova:
@tbhallett, Okay, if this is the intended behaviour for overriding a Treatment ID.

Initially, I had the impression that this would introduce an additional feature, allowing us to set a probability for the treatment as a whole, like a complete package of items ready to be used. That would be different from setting probabilities for each individual item, which is already possible using the `override_availability()` function.

That said, this approach still simplifies the process, so there’s some advantage, even if it’s not a completely new feature.

However, the checks still need to be fixed, btw.

tbhallett:
We could do this to enable use of wildcards to indicate sets of TREATMENT_IDS efficiently, including all:

```python
def is_treatment_id_ovrridden(treatment_id: str, cons_avail_override_treatment_ids: list[str]) -> bool:
    """Determine if a treatment_id (specified as a string) can is subject to consumables availability overriding (i.e., is with the allowable set of treatments, given by `cons_avail_override_treatment_ids`. The rules are as follows:
          * An empty list means nothing is subject to overriding.
          * A list that contains only an asterisk ['*'] means EVERYTHING is subject to overriding
          * An entry in the list of the form "A_B_C" means a treatment_id that matches exactly is allowed
          * An entry in the list of the form "A_B_*" means that a treatment_id that begins "A_B_" or "A_B" is allowed"""

    def _treatment_matches_pattern(_treatment_id, _service_availability):
        """Check if treatment_id matches any services specified with wildcard * patterns"""

            def _matches_this_pattern(_treatment_id, _s):
                """Returns True if this `treatment_id` is consistent with this component of `cons_avail_override_treatment_ids`"""
                if '*' in _s:
                    assert _s[-1] == '*', f"Component of service_availability has an asteriks not at the end: {_s}"
                    _s_split = _s.split('_')  # split the matching pattern at '_' knowing that the last component is '*'
                    _treatment_id_split = _treatment_id.split('_', len(_s_split) - 1)  # split treatment_id at '_' into
                    # as many component as there as non-asteriks component of _s.
                    # Check if all the components (that are not asteriks) are the same:
                    return all(
                        [(a == b) or (b == "*") for a, b in itertools.zip_longest(_treatment_id_split, _s_split)]
                    )
                else:
                    # If not "*", comparison is ordinary match between strings
                    return _treatment_id == _s

                for _t in cons_avail_override_treatment_ids:
                    if _matches_this_pattern(_treatment_id, _t):
                        return True
                    else:
                        return False


        if not cons_avail_override_treatment_ids:
            # Empty list --> nothing is subject to overriding
            return False

        elif cons_avail_override_treatment_ids == ['*']:
            # Wildcard --> everything is subject to overriding
            return True
        
        elif treatment_id in cons_avail_override_treatment_ids:
            # Explicit inclusion of this treatment_id --> subject to overriding
            return True
        
        else:
            if _treatment_matches_pattern(treatment_id, service_availability):
                return True
        
        return False
```

EvaJanouskova:
I think I understand the idea, and it sounds good. It will just need a bit of tidying up:

> ```python
> def is_treatment_id_ovrridden(treatment_id: str, cons_avail_override_treatment_ids: list[str]) -> bool:
>     """Determine if a treatment_id (specified as a string) can is subject to consumables availability overriding (i.e., is with the allowable set of treatments, given by `cons_avail_override_treatment_ids`. The rules are as follows:
>           * An empty list means nothing is subject to overriding.
>           * A list that contains only an asterisk ['*'] means EVERYTHING is subject to overriding
>           * An entry in the list of the form "A_B_C" means a treatment_id that matches exactly is allowed
>           * An entry in the list of the form "A_B_*" means that a treatment_id that begins "A_B_" or "A_B" is allowed"""
> 
>     def _treatment_matches_pattern(_treatment_id, _service_availability):
>         """Check if treatment_id matches any services specified with wildcard * patterns"""
> 
>             def _matches_this_pattern(_treatment_id, _s):
>                 """Returns True if this `treatment_id` is consistent with this component of `cons_avail_override_treatment_ids`"""
>                 if '*' in _s:
>                     assert _s[-1] == '*', f"Component of service_availability has an asteriks not at the end: {_s}"
>                     _s_split = _s.split('_')  # split the matching pattern at '_' knowing that the last component is '*'
>                     _treatment_id_split = _treatment_id.split('_', len(_s_split) - 1)  # split treatment_id at '_' into
>                     # as many component as there as non-asteriks component of _s.
>                     # Check if all the components (that are not asteriks) are the same:
>                     return all(
>                         [(a == b) or (b == "*") for a, b in itertools.zip_longest(_treatment_id_split, _s_split)]
>                     )
>                 else:
>                     # If not "*", comparison is ordinary match between strings
>                     return _treatment_id == _s
> 
>                 for _t in cons_avail_override_treatment_ids:
>                     if _matches_this_pattern(_treatment_id, _t):
>                         return True
>                     else:
>                         return False

I'm not quite sure what the for loop above is supposed to be doing. Also, it's currently placed inside the function it's calling

I believe the lines below should also be indented to the left.

> 
>         if not cons_avail_override_treatment_ids:
>             # Empty list --> nothing is subject to overriding
>             return False
> 
>         elif cons_avail_override_treatment_ids == ['*']:
>             # Wildcard --> everything is subject to overriding
>             return True
>         
>         elif treatment_id in cons_avail_override_treatment_ids:
>             # Explicit inclusion of this treatment_id --> subject to overriding
>             return True
>         
>         else:
>             if _treatment_matches_pattern(treatment_id, service_availability):
>                 return True
>         
>         return False
> ```

The variable service_availability isn't defined anywhere. I feel like the else block might actually be intended to be replaced by the for loop from above, is that the case?

src/tlo/methods/consumables.py
tamuri: Tim has answered this below - resolving.

Issue #1646, opened 2025-07-13
resources/healthsystem/consumables/README.md -- not a pointer?


EvaJanouskova:
Hi @tamuri and @tbhallett, since I merged in the latest master, pycharm is telling me there is a change in `resources/healthsystem/consumables/README.md`, but if I open the Show Diff, it says 'No Difference'. When I try to use git difftool in my Terminal (not in the pycharm), it says:
```
Failed to register: Timeout was reached
Encountered 1 file that should have been a pointer, but wasn't:
	resources/healthsystem/consumables/README.md
```

I wonder if it could be because someone pushed the `resources/healthsystem/consumables/README.md` file without using Git LFS but on my side it is expecting LFS pointer instead of the full file?

Or is it something else, would you know?

PR #1647, opened 2025-07-14
gitattributes: excl resources/**/*.md from LFS

EvaJanouskova:
This update excludes all `resources/**/*.md` files from being tracked by Git LFS, ensuring that Markdown files under the `resources/` directory are treated as regular text files.

Fixes: #1646.


EvaJanouskova:
> That way, we don’t need to make any changes to .gitattributes.

Isn’t that exactly what was done, and why we were experiencing the issue described here: https://github.com/UCL/TLOmodel/issues/1646?

Also, when you put it on LFS, you can’t see what changes have been made to the README directly on GitHub, which is why I thought this might be a better solution

PR #1653, opened 2025-08-04
Paramater Revamp - Measles

mmsuarezcosta:
Updating measles parameters (removing module hardcoding and adding labels) 

**To Note:** 
- Deleted legacy _beta.csv_ as it was not used anywhere in the code 

**Parameters to Review labeling:** 
- Prob_severe: Marked as undetermined, but unsure if should be updated
- Universal parameters (all): all are universal properties linked to references in the write-up, but 1) I did not go into the references to specifically determine a non-vague prior range, 2) I cannot confirm that they were not adjusted for calibration purposes. 

**Reference Notes:** 
No access to resource to get priors for beta and phase (h[ttps://pubmed.ncbi.nlm.nih.gov/18256664/](https://pubmed.ncbi.nlm.nih.gov/18256664/) )
No access to resource to get priors for vaccine efficacy (Strebel, P. M., M. J. Papania, P. A. Gastañaduy and J. L. Goodson (2018). 37 - Measles Vaccines. Plotkin's Vaccines (Seventh Edition). S. A. Plotkin, W. A. Orenstein, P. A. Offit and K. M. Edwards, Elsevier: 579-618.e521)


mmsuarezcosta:
@marghe-molaro ready for review

mmsuarezcosta:
Thanks so much @marghe-molaro for the review and feedback 

Agree
- `beta_baseline`: Reviewed and agree. 
- `prob_severe` and `risk_death_on_treatment` : Agree to remove. Moving forward, will watch out for cases like this
-  `symptoms`: Reviewed and agree 

Open discussion
- `treatment_followup_days`: This is the days until follow up appointment after treatment. I deemed this to be guided by Malawi healthcare guidelines, and thus labeled it as local.  Is local the correct label in this case? Would a more descriptive name such as `followup_appt_schedule_days_post_treatment` serve better?
- `reduction_in_death_risk_treated_measles`: So in this case, I labeled it as undetermined primarily because i decided that if the modeler chose a specific value, they probably chose the value that most served the calibration purposes. However, should the logic be as follows: if developer reviews evidence for a variable, then automatically it cannot be defined as undetermined. Instead, it is 'local' if it came from a local source, or 'universal' if it came from a universal source. 
Pending update: `reduction_in_death_risk_treated_measles`, `treatment_followup_days`

- All other parameters related to transmission dynamics switch to 'local': Makes sense. I honestly wasn't exactly sure what the source of these values were, and I deemed it to be the CDC, which was universal [HERE](https://physiciansforinformedconsent.org/epidemiology-and-prevention-of-vaccine-preventable-diseases-13th-edition-pink-book-2015.pdf), but I believe they may come more from [HERE](https://www.nature.com/articles/nrmicro1550) ?

**Completed updated to local**: death_timing_min_days, death_timing_max_days, natural_resolution_min_days, natural_resolution_max_days, symptom_onset_min_days, symptom_onset_max_days, treatment_followup_days
**Kept as universal**: `maternal_immunity_age_threshold`,` age_min_symptoms_constant_rate`

-

mmsuarezcosta:
Updates: 
- `treatment_followup_days`: this one has definition that is not accurate in the code. Need to update the definition and rename the parameter

marghe-molaro:
Hi @mmsuarezcosta, 
I think since it is describing the period over which we'd expect a single infection peak it should be labelled as local (there might be contexts in which we expect two infection peaks in a year, in which case the period should be 6 months). So this is not a structural assumption.
Notice that for the same reason the phase_shift should not be a structural assumption, since it describes in what month of the year we'd expect that peak (for Malawi it is assumed to be May, but it could be on another month in a different context)

mmsuarezcosta:
@marghe-molaro Understood. I've changed the values as follows (please confirm this is OK)
- `period`: prior_min = 6 months, prior_max = 24 months, prior_note = vague prior
- `phase_shift`: prior_min = 0 months, prior_max = 23 months, prior_note = vague prior

mmsuarezcosta:
Ready for @mnjowe review

tbhallett (approved):
with that change, I can approve --  if you're happy with it.

Note that prior_min and prior_max for a dataframe could be overwritten to but it cumbersome!!

A mrore manual  (but ultimately neater approach is really to simplify everthing into a few additional parameters in paraeter.csv (e.g. CFR_age_0_to_1, prob_rash_0_to_5, etc.)

src/tlo/core.py
```diff
@@ -397,6 +398,21 @@ def load_parameters_from_dataframe(self, resource: pd.DataFrame) -> None:
...
+                                   prior_min: Optional = None,
+                                   priox_max: Optional = None) -> None:
+        """Declare metadata for a parameter"""
+        if parameter_name not in self.PARAMETERS:
+            raise ValueError(f"Parameter {parameter_name} not declared in PARAMETERS dictionary")
+
+        if self.PARAMETERS[parameter_name].metadata != {}:
+            warnings.warn(f"Parameter {parameter_name} already has metadata declared")
+
+        self.PARAMETERS[parameter_name].metadata.update(param_label=param_label, prior_min=prior_min, prior_max=priox_max)
```
mnjowe:
```suggestion
        self.PARAMETERS[parameter_name].metadata.update(param_label=param_label, prior_min=prior_min, prior_max=prior_max)
```

src/tlo/methods/measles.py
```diff
@@ -105,9 +142,14 @@ def read_parameters(self, resourcefilepath: Optional[Path] = None):
         self.load_parameters_from_dataframe(workbook["parameters"])
 
         self.parameters["symptom_prob"] = workbook["symptoms"]
+        self.declare_parameter_metadata(parameter_name="symptom_prob", param_label="local")
```
mnjowe: I feel like we can easily miss a parameter with this approach. Better assign metadata inside `load_parameters_from_dataframe` method where we have access to all declared parameters via `self.PARAMETERS` dictionary?

src/tlo/methods/measles.py
```diff
@@ -196,12 +240,13 @@ def process_parameters(self):
...
         Make `self.symptom_probs` to be a dictionary keyed by age, with values of dictionaries keyed by symptoms and
         the probability of symptom onset."""
+        p = self.parameters
         probs = self.parameters["symptom_prob"].set_index(["age", "symptom"])["probability"]
         self.symptom_probs = {level: probs.loc[(level, slice(None))].to_dict() for level in probs.index.levels[0]}
 
         # Check that a sensible value for a probability of symptom onset is declared for each symptom and for each age
-        # up to and including age 30
-        for _age in range(30 + 1):
+        # up to and including age for which symptom probabilities stabilizes
```
mnjowe: now that we have taken out 30, I feel like the comment is not clear enough. up-to which age? or which age is age for which symptom probabilities stabilizes?

src/tlo/methods/measles.py
```diff
@@ -225,7 +270,8 @@ class MeaslesEvent(RegularEvent, PopulationScopeEventMixin):
     """
 
     def __init__(self, module):
-        super().__init__(module, frequency=DateOffset(months=1))
+        p = module.parameters
+        super().__init__(module, frequency=DateOffset(months=p['main_polling_event_frequency_months']))
```
mnjowe: I have asked a question before and I will also put it here. is the frequency for events such as this a local behavior that we expect it to change with different context?

src/tlo/methods/measles.py
```diff
@@ -64,14 +64,51 @@ class Measles(Module, GenericFirstAppointmentsMixin):
...
+            Types.REAL, "Minimum days from infection to symptom onset (incubation period)"
+        ),
+        "symptom_onset_max_days": Parameter(
+            Types.REAL, "Maximum days from infection to symptom onset (incubation period)"
+        ),
+        "sympt_resolution_after_treatment": Parameter(
+            Types.REAL, "Days until symptom resolution after treatment"
+        ),
+        "main_polling_event_frequency_months": Parameter(
+            Types.REAL, "Measles main polling event frequency months"
```
mnjowe: I think all these can be change to Types.INT?

src/tlo/core.py
```diff
@@ -397,6 +398,21 @@ def load_parameters_from_dataframe(self, resource: pd.DataFrame) -> None:
                 prior_max=prior_max
                 )
 
+    def declare_parameter_metadata(self,
```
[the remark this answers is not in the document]
mmsuarezcosta: Agree this can be improved for other parameter files. Going to leave as-is in this PR, but will be beneficial to create this PR when addressing other parameter files. Emmanuel, would you be able to take the lead on this?

src/tlo/methods/measles.py
mmsuarezcosta: It is a module design decision. We are also adding these as parameters in the CSV file. They are not to change by context, but rather they may change if the module lead decides to update the design or other parametres within the module.

mmsuarezcosta:
> Thanks Mariana and Tim for your great work on this. Kindly find my comments below.
> 
> My question regarding `parameters.csv` concerns the relationship between the parameter label and its reference. I’ve noticed some parameters labeled as “local” but linked to references that are not Malawi-specific. Could you please clarify if this is intentional or acceptable?
> 
> As a minor concern, I’m a bit worried about having too many parameters even for smaller details such as dates or event frequencies in the resource file as this could potentially complicate the model’s adoption for new analysts

Thank you @mnjowe  for reviewing. Your review and feedback are extremely helpful. Thanks always for your level of detail in review. 

 - **Local references for universal params**: This is OK. The goal in param labeling is to capture the essence of what the parameter is to be. So for example, a death rate may be expected to be universal and consistent regardless of context, yet it is cited from a local source. It is a grey area! If there are some that you believe should indeed be local that I have labeled as universal, please bring these forward so that we can discuss. 
 - Too many parameters: Yes, this is definitely a big debate, and other developers have expressed similar concerns. For this project, we have erred on the side of 'over paramaterizing'. Can open this discussion again. 

Please see complete documentation on guide for labeling [HERE](https://docs.google.com/document/d/1DkQrjpUoCVKuDGygGelOirRePh-NopBe49f90-uYUyg/edit?tab=t.0) in tab **Guide for Modelers** that may support in ongoing revisions.

src/tlo/methods/measles.py
mnjowe: Sure, I can open a PR addressing this

src/tlo/methods/measles.py
mnjowe: or you can say;

`# up to and including age for which symptom probabilities stabilizes( age >= put parameter prior_min value here)`

mnjowe:
On point number one I was asking about universal references for local params. Not sure if the explanation is the same
> Thank you @mnjowe for reviewing. Your review and feedback are extremely helpful. Thanks always for your level of detail in review.
> 
> * **Local references for universal params**: This is OK. The goal in param labeling is to capture the essence of what the parameter is to be. So for example, a death rate may be expected to be universal and consistent regardless of context, yet it is cited from a local source. It is a grey area! If there are some that you believe should indeed be local that I have labeled as universal, please bring these forward so that we can discuss.
> * Too many parameters: Yes, this is definitely a big debate, and other developers have expressed similar concerns. For this project, we have erred on the side of 'over paramaterizing'. Can open this discussion again.
> 
> Please see complete documentation on guide for labeling [HERE](https://docs.google.com/document/d/1DkQrjpUoCVKuDGygGelOirRePh-NopBe49f90-uYUyg/edit?tab=t.0) in tab **Guide for Modelers** that may support in ongoing revisions.

mnjowe (approved):

tbhallett:
@tdm32  -- does this look OK to you? 
It seems good to me.

tdm32:
> @tdm32 -- does this look OK to you? It seems good to me.

Sorry @tbhallett I'm late to respond, yes looks fine thank you.

PR #1660, opened 2025-08-05
Param Revamp - Bladder Cancer 

mmsuarezcosta:
Updating bladder cancer parameters (removing module hardcoding and adding labels) 

**To Note:** 
NA

**Parameters to Review labeling:** 
- All init_prop: Currently labeled as 'undetermined', but unsure if should be labeled as 'local' given they were based on references, but likely adjusted for calibration
- _local_ and _universal_ parameters with _Missing_ references to review. Need confirmation to know if they are 'undetermined'  

**Reference Notes:** 
NA

### DIFF
--- a/resources/ResourceFile_Bladder_Cancer/parameter_values.csv
+++ b/resources/ResourceFile_Bladder_Cancer/parameter_values.csv
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:bdee94acfce7d6f790a1048cf313c5d48ee35de04ef0603cc8bc40b0ef635042
-size 4982
+oid sha256:2e00332091bfb52e123add10c0ff68cdb6646ee8b256a521be4a2ce242874a37
+size 5058
--- a/src/tlo/methods/bladder_cancer.py
+++ b/src/tlo/methods/bladder_cancer.py
@@ -196,6 +196,9 @@ class BladderCancer(Module, GenericFirstAppointmentsMixin):
         "main_polling_frequency_months": Parameter(
             Types.INT, "frequency in months for main polling event"
         ),
+        "delay_initial_main_polling_event_days": Parameter(
+            Types.INT, "delay in months for initial main polling event"
+        ),
         "beddays_treatment": Parameter(
             Types.INT, "number of bed days required for bladder cancer treatment"
         ),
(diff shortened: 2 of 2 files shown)


mmsuarezcosta:
ready for review @marghe-molaro

mmsuarezcosta:
@marghe-molaro 

**- Local parameters with 'missing' reference:** My bad. I notice that when I switched over the init probabilities from 'undetermined' to 'local', I forgot to change the reference column from 'N/A' to 'Missing'. This is the reason why these are like this. 
- **Flagging code discrepances:** Will be sure to keep in the comments. Regarding flagging to developers, what is the best way that you recommend I do this? Is it to make a comment and tag the respective developer to address? 
- **Analysis file**: So in the beginning, out of habit when I was developing the cervical cancer module, I was running the 'analysis' file rather than the 'test' file to make sure my changes were OK, and then ran the test file. I realize I did not have to do that, just the test file would suffice. 
The analysis file for these cancer modules will not run if they do not have the updates that I have added to them (changes in requirements since developers initially wrote them). This has no implication on the test files. I considered reverting it, but I kept it because it is an improvement. Let me know if you think I should revert this because it may be confusing why it is with this PR.

andrew-phillips-1:
This looks all OK to me thanks, although I could not view the parameter file.  I think there is a clear error with modules that are implementing a 12 year follow up  - the comment is right this should be months and not years in all cases.

mmsuarezcosta:
> I wasn't sure about these being viewed as universal: rp_bladder_cancer_tobacco,3.5,universal,0.01,100,vague prior,"Cumberbatch MG et al. The Role of Tobacco Smoke in Bladder and Kidney Carcinogenesis: A Comparison of Exposures and Meta-analysis of Incidence and Mortality Risks. DOI: 10.1016/j.eururo.2015.06.042 European urology , 2016, Vol.70(3), p.458-466" rp_bladder_cancer_schisto_h,5,universal,0.01,100,vague prior,Missing

- `rp_bladder_cancer_tobacco`: My logic was as follows -- given that this is a variable based on a biologically grounded driver of the disease AND I deem this source to be universal (a meta-analysis, not specific to African or LIC context), I believe this variable is universal. @andrew-phillips-1, can you further elaborate on why you believe this variable should be local? Thanks. 
- `rp_bladder_cancer_schisto_h`: Agree, should be local based on source (Parkin M, Hämmer L, Ferlay J, Kantelhardt EJ.   Cancer in Africa 2018: The role of infections.   Int. J. Cancer   2019). I will change this.

mmsuarezcosta:
> It was just that it is a relative prevalence rather than a relative incidence which feels like it would perhaps depend on local prevalences. I'm not strongly arguing for a change though.

Ahh I see, that is a good point and I agree. Will aim to be consistent with the labeling of relative prevalence

PR #1680, opened 2025-08-22
Correct measles transmission function

marghe-molaro:
Typo in transmission function (2+pi instead of 2*pi) means there currently isn't a yearly peak in transmission in May as referenced in the write-up. 

Suggest also removing comment on effect of HIV on cfr, since it's currently not accounted for.


tdm32 (approved):

matt-graham:
Actually looking at the calibration workflow logs it looks like the runs for the last week haven't completed successfully - looks like this is possible [an issue with runners running out of disk space](https://github.com/UCL/TLOmodel/actions/runs/17087295596), so this may need to be resolved first before setting off a run.

tamuri:
It's this one: https://github.com/UCL/TLOmodel/actions/runs/17267958593

tamuri:
The workflow successfully completed. Results are on the dev website (which we _will_ make public 🙈)

PR #1684, opened 2025-08-26
Param Revamp - Newborn Outcomes

mmsuarezcosta:
**To note** 

**Parameters to review**
- Some hard coded values remain 
- The PDF is very well documented in terms of the parameters that have been reviewed for calibration and the references for each parameter. I did not go one-by-one to determine the best label, this work is still required. Any that are 'undetermined' are because I have explicitly found that they were used for calibration.  
- If truly no data exists, and we create a 50/50 draw, I am calling this parameter 'local' as it is the 'safest' option and requires revision. Does that sound OK? This is specifically for all the variables differentiating between moderate and mild disabilities.


joehcollins (approved):
All looks good here as well. just a couple of updates to the parameter labels

PR #1689, opened 2025-09-01
Remove reference to squeeze factors in test and mode 1 check

marghe-molaro:
This PR is related to PR #1662 and addresses the following issues:

- The reference to squeeze has been removed from `test_rescaling_capabilities_based_on_squeeze_factors`, which has been renamed to `test_rescaling_capabilities_based_on_load_factors`.

- The reference to squeeze has also been removed from the check determining whether HSI in mode 1 can proceed. The check now verifies if any required officer has a non-zero capability, rather than checking if any squeeze factor is inf.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -1606,6 +1606,66 @@ class HealthSystem(Module):
 
         return appt_footprint_times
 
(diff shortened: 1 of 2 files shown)


tbhallett:
Closing in favour of #1781

Issue #1681, opened 2025-08-22
Discrepancy between comment and code for PostTreatmentCheck in Cancer modules


marghe-molaro:
@mmsuarezcosta spotted several instances (in Bladder, Oesophageal, Prostate, and Other cancers at least) where there are discrepancies between the stated and coded delay in scheduling `HSI_PostTreatmentCheck`, e.g. 

https://github.com/UCL/TLOmodel/blob/c616315c3c0668b18f665d92e62a9268c65d108d/src/tlo/methods/oesophagealcancer.py#L800-L809

We are flagging this in case the error is in the code, rather than the comment

marghe-molaro:
I think possibly, yes; what might have happened is that this was an error in one of the cancer modules which was then accidentally copied/pasted in the other ones as well.

If the error is in the code rather than the comment (i.e. the follow-up should take place on a timescale of months rather than years) then it would be good to address this asap. Do you think this would have had much impact on dalys/deaths?

andrew-phillips-1:
I doubt it would have much (any) impact on dalys or deaths as the person is either cured or not at that point.  I would need help from someone to change the code and do the pr as I don't know the process now.

marghe-molaro:
Great, if it is unlikely to impact DALYs/deaths (and, I assume, healthcare resources available) then at least it's not urgent. We can leave it as an issue for now and then find someone to fix it.

marghe-molaro:
This issue was addressed by merged PR #1691, and therefore can be closed.

PR #1691, opened 2025-09-02
Fix typo in cancer PostTreatmentCheck scheduling

marghe-molaro:
This PR solves issue #1681 

Hi @andrew-phillips-1, I have corrected the delay to PostTreatmentCheck from units of "year" to "month" in the cases where there was a discrepancy between comment and code. 

There are just a few additional cases where things are unclear, e.g. a few cases where the comment suggests there should be an interval of 1 month until the PostTreatmentCheck, but it is coded as 3 months, e.g.:
https://github.com/UCL/TLOmodel/blob/42a56c40c9650e112ef11413cc6bfdd3dbddeb17/src/tlo/methods/breast_cancer.py#L868-L877
and cases where the comment says the interval should be 12 or 1 month, but it is coded as 3 months, e.g.:
https://github.com/UCL/TLOmodel/blob/42a56c40c9650e112ef11413cc6bfdd3dbddeb17/src/tlo/methods/other_adult_cancers.py#L798-L807

Could you please advise on these cases? Many thanks in advance!

### DIFF
--- a/src/tlo/methods/bladder_cancer.py
+++ b/src/tlo/methods/bladder_cancer.py
@@ -909,7 +909,7 @@ class HSI_BladderCancer_StartTreatment(HSI_Event, IndividualScopeEventMixin):
                     module=self.module,
                     person_id=person_id,
                 ),
-                topen=self.sim.date + DateOffset(years=12),
+                topen=self.sim.date + DateOffset(months=12),
                 tclose=None,
                 priority=0
             )
@@ -961,7 +961,7 @@ class HSI_BladderCancer_PostTreatmentCheck(HSI_Event, IndividualScopeEventMixin)
                     module=self.module,
                     person_id=person_id
                 ),
-                topen=self.sim.date + DateOffset(years=1),
+                topen=self.sim.date + DateOffset(months=1),
                 tclose=None,
                 priority=0
             )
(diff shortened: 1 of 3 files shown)


marghe-molaro:
I see, thanks for clarifying @andrew-phillips-1!

In terms of this PR, aside from the fix of the typo months -> years, what do you think we should do about these remaining discrepancies in the number of months to PostTreatment?

andrew-phillips-1:
I think after those corrections we probably don't have enough evidence to worry about changing it

tbhallett (approved):
Thanks both!
This all looks good to me.

I'll just push a commit modifying ever so slightly the comment and then merge.

tbhallett:
> I think after those corrections we probably don't have enough evidence to worry about changing it

@andrew-phillips-1 --- this will mean that person will have post-treatment checks (for prostate cancer) every month FOR EVER (until they did, or the cancer becomes metastatic). This sounds a bit extreme (and makes more work for the HCW than it reasonable, so will knock off the calibrations). Probably, it was right originally: monitoring every YEAR rather than every month.....? Shall we make that the case for all the cancer modules?

andrew-phillips-1:
yes sorry you are right - i thought we were doing 6 months, 1 year and then annual.

PR #1701, opened 2025-09-08
Param Revamp - Depression

mmsuarezcosta:
Variables to review
-------------------
### Undetermined, 'Calibration'
- `base_3m_prob_depr`: Believe this is a calibration variable given that it has been set to the ratio that is output by the model. However, it does reference local sources, so there is an argument for it to be local. Full developer note is as follows: 
> Rate is derived indirectly based on prevalence of depression that model produces as an output (Abas et al, 1997; Marwick et al, 2010; Udedi et al, 2014; This is much higher than incidence in Todd et al 1999 (approx 0.16 in 2 months) but that is inconsistent with prevalence unless extremely short duration. )

### Local 
- Most of the relative rate variables (e.g. `rr_depr_wealth45` and `rr_depr_pregnancy` ) are deemed to be local rather than universal in this module. This is due to the reference source (many derived from local context) and the understanding that depression is affected more by social norms/constructs than other diseases. This can be revised if not the case. 
- `antidepr_refill_window_days`: `antidepr_refill_interval_months` is a local variable, and the window of 7 days for which it can be scheduled has been set as a local variable as well. Uncertain if it may be a design decision as it may not be inherent to the MW healthcare system practices. 

### Hard coded values still in code
-  Wealth bracket 4,5 relative to 1,2,3 (categorical numbers)
- Medication dosage calculations


mmsuarezcosta:
Ready for review @marghe-molaro

marghe-molaro:
Hi @mmsuarezcosta, given that @andrew-phillips-1 is happy with this as module lead I am also labelling as ready to be merged on the tracker file

tbhallett (approved):

PR #1706, opened 2025-09-17
Add comprehensive equipment to disease modules with facility-level de…

andrew-phillips-1:
As you will guess, I interacted with Github co-pilot quite a bit in preparing this - so this is partly an exploration of whether / how we can use AI to help with this kind of thing - I'm sure we will need some iterations.

- Cancer modules (breast, prostate, bladder, oesophageal, other adult cancers):
  * Added facility-level dependent equipment based on column H specifications
  * Equipment includes imaging (X-ray, ultrasound), laboratory analysis, and cancer-specific tools
  * Probabilities vary by facility level (1a=community, 1b+=hospitals)

- COPD module:
  * Added moderate exacerbation treatment HSI with respiratory equipment
  * Incentive spirometers (100% availability) and blood gas analyzers (facility-dependent)
  * Enhanced severe exacerbation equipment with respiratory support tools

- Epilepsy module:
  * Added comprehensive rehabilitation equipment for severe epilepsy cases
  * Pediatric-specific equipment for children with epilepsy
  * Assessment tools, mobility aids, and physiotherapy equipment

- Fixed equipment codes:
  * 454  202 (X-ray machine)
  * 453  161 (Ultrasound scanning machine)

All equipment codes validated against ResourceFile_EquipmentCatalogue.csv

### DIFF
--- a/src/tlo/methods/bladder_cancer.py
+++ b/src/tlo/methods/bladder_cancer.py
@@ -734,6 +734,26 @@ class HSI_BladderCancer_Investigation_Following_Blood_Urine(HSI_Event, Individua
             # Use a biopsy to diagnose whether the person has bladder Cancer
             # If consumables are available update the use of equipment and run the dx_test representing the biopsy
             self.add_equipment({'Cystoscope', 'Ordinary Microscope', 'Ultrasound scanning machine'})
+            
+            # Additional cancer equipment based on column H comments
+            # Laboratory equipment for cancer screening and diagnosis
+            if self.module.rng.random() < 0.80:  # 80% probability at level 1b
+                self.add_equipment({221})  # Analyser, Haematology
+            
+            if self.module.rng.random() < 0.70:  # 70% probability at level 1b  
+                self.add_equipment({220})  # Analyser, Chemistry
+            
+            # Urological imaging equipment - relevant for bladder cancer
+            if self.module.rng.random() < 0.60:  # 60% probability at level 1b
+                self.add_equipment({202})  # X-ray machine
+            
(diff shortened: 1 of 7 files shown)


tbhallett:
Thanks @andrew-phillips-1  - this is great. Two quick comments which I think will lead to simplifications:
* We've got the big operation underway currently to remove all hard-coded parameters values, so if you want to include those probabilities, I think we'd need to include them formally as parameters.
* The probability of use seems unknown in practice (right?), and the equipment use we need to capture is really "ever use" (if its ever used, then it need to be there and paid for).


So, I would suggest dropping the `if... ` causes, and just adding the equipment directly.
You can also add more than one item at a time, by separating the codes them with commas, so that should make it neater and shorter.
Lastly, I think most of the time we use name for the equipment to make it make it more 'readable'. Maybe you prefer the codes though, or the names didn't work? Its not a dealbreaker and the utility functions are designed to work either way.

andrew-phillips-1:
Happy to take out the new HSI but we feel the new HSI class is justified because:
Different facility levels: Moderate (1a) vs Severe (1a→1b→2)
Different equipment needs: Basic respiratory support vs full oxygen therapy
Previously moderate exacerbations only got inhalers in the generic first appointment

tbhallett:
> Happy to take out the new HSI but we feel the new HSI class is justified because: Different facility levels: Moderate (1a) vs Severe (1a→1b→2) Different equipment needs: Basic respiratory support vs full oxygen therapy Previously moderate exacerbations only got inhalers in the generic first appointment

Ok, if you think justified and its intentional then that's fine.

PR #1721, opened 2025-09-29
Param Revamp - COPD

mmsuarezcosta:
### File 
Resource file placement to be updated. COPD not under folder. 

### Hard coded parameters
The names of the parameters with age groups have been kept as-is (e.g. `idx_notover15`) 

### Test File 
Had to updated `def test_ch_lungfunction` because module because I have moved it to be a method of the CopdModels class, so the way it is called in the test function must be updated accordingly. 

```
def test_ch_lungfunction():
    """ test everyone ends up in lung function category 6 if high progression rate is set """
    sim = get_simulation(10)
    df = sim.population.props

    copd_module = sim.modules['Copd']

    # make all individuals qualify for progressing to the next lung function
    df.loc[df.index, 'is_alive'] = True
    df.loc[df.index, 'ch_lungfunction'] = 0
    df.loc[df.index, 'age_years'] = np.random.choice(range(20, 50), len(df))

    # check they're all eligible to progress to the next lung function
    assert all(copd_module.models.eligible_to_progress_to_next_lung_function(df)), 'some are still not eligible to progress to ' \
                                                                     'next lung function'
```

### DIFF
--- a/resources/ResourceFile_COPD/parameter_values.csv
+++ b/resources/ResourceFile_COPD/parameter_values.csv
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5f3742a751610ab80c218854f0e2fb38de94cf6844efe8a534efcaac0b61431a
+size 2355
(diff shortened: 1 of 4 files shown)


andrew-phillips-1:
seems ok to me thanks

marghe-molaro:
Hi @mmsuarezcosta, if @andrew-phillips-1 as module leader is happy with this I am happy with it too! Marked as ready to merge on the tracker file

PR #1722, opened 2025-09-29
Param Revamp - Stunting

mmsuarezcosta:
### Local v Universal
- `effectiveness_of_complementary_feeding_education_in_stunting_reduction`: marked as _local_ because believe education is location-specific 
- `effectiveness_of_food_supplementation_in_stunting_reduction`: marked as _universal_ because believe this can be compared to a 'treatment' 
- _Risk factors for stunting prevalence at initiation_: In conversation with @marghe-molaro , stated most local except those relating to term. I have also labeled `or_stunting_male` as universal. 
- _The incidence of stunting_: In conversation with @marghe-molaro , stated most local except maybe `rr_stunting_per_diarrhoeal_episode`. I would argue that this one should also be local as it is in the parameter category of local incidence rate. However, there may be an argument for this variable and others (e.g. `rr_stunting_preterm_and_AGA` and `rr_stunting_SGA_and_preterm` to be deemed as universal) 
- _Progresso to severe stunting_: In conversation with @marghe-molaro , stated these may be considered universal, however I have marked them as local given that the sources are MW based and this can have local elements. 

### Hard coding
Keep age group values hard coded (e.g. 1 to 2, 2 to 3, ...). Only parameterized the max age of stunting (age 5) 

## Files

- Delete file 'Cover Sheet.csv', not used anywhere in module 
- Rename file Parameter_values to parameter_values

### DIFF
--- a/resources/ResourceFile_Stunting/parameter_values.csv
+++ b/resources/ResourceFile_Stunting/parameter_values.csv
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:681f0869a8d01d9edd4a9a7b1024cfa5acce44bd80c0cf92baa5b589b6c1fe37
-size 3180
+oid sha256:b306a546a4680e4459cc878ef8e04475cb29a9d2022edad9bcea240dbf76f4e1
+size 3117


marghe-molaro:
Thanks @mmsuarezcosta, 

This looks good to me, I think all hardcoded parameters have been removed so updating this in the sheet file.

All other labelling decisions - specifically which parameters classify as universals and which don't - will need to be reviewed by the module developer.

@EvaJanouskova , ready for you to review

PR #1726, opened 2025-10-14
"Partial death" when collecting mnh events

marghe-molaro:
Hi @joehcollins,
 
Quick update:
- All mni variable changes are now captured by the functions detecting whether an "event" has occurred for an individual;
- I have implemented the "partial survival"/"partial death" approach in your modules* (but only when collecting event chains).   
This should hopefully really help you get your statistics up for your interventions project!
- This is now all compatible with your cohort model;

Just sharing this now in case you wanted to have a quick look before our in person-meeting on the 21/10/25, but we can discuss this in more detail then. If you decide you'd prefer not to use the partial death approach, the event collection including mni variables is of course still compatible. 

*Across pregnancy_supervisor.py, labour.py, and postnatal_supervisor.py, I found 8 instances where do_death was invoked, if I missed anything please let me know!


marghe-molaro:
Close to revisit in future

PR #1730, opened 2025-10-16
Param Revamp - ALRI 

mmsuarezcosta:
### Deleted
- _Calculations.csv_, _Lazzerini CFR.csv,_ and _Pathogen_specific.csv_ were not in use and have been deleted 

## Parameters 
Note: There is discrepancy between the write-up and the parameter values on the CSV file (parameters names, list, values, ...) 

### Dummy/Assumed
I am documenting the following parameters in the CSV file that have been either marked as _dummy_ or _assumed_ : `prob_lung_abscess_in_pulmonary_complicated_pneumonia`, `prob_pneumothorax_in_pulmonary_complicated_pneumonia`, `prob_cough_in_other_alri`, `prob_difficult_breathing_in_other_alri`, `prob_chest_indrawing_in_other_alri`, `prob_danger_signs_in_sepsis`, `prob_iCCM_severe_pneumonia_treated_as_fast_breathing_pneumonia`, `prob_IMCI_severe_pneumonia_treated_as_non_severe_pneumonia`

### Local v. Universal 
- Any proportion has been defined as local, and any probability (parameter with prefix with `prob`)  has been defined as universal. A recommended exercise will be to revisit to `prob` parameters and determine if any are local-specific
- All treatment failure parameters (prefixed with `tf`) are assumed to be based on the true effect of the drug treatment, not on any location-specific factors (eg drug adherence) and thus are deemed to be universal.

### DIFF
--- a/resources/ResourceFile_Alri/parameter_values.csv
+++ b/resources/ResourceFile_Alri/parameter_values.csv
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:0fb2df4b88c2ef765146fa3250d8d972f76ebe34f9f56234a3acb987f0620b80
-size 13366
+oid sha256:e2b417a870b56a1301f13c0fa9e0624a395e4e5e598b506052db7b741d42ea16
+size 13358


marghe-molaro:
Hi @mmsuarezcosta, 

Many thanks for this!

I believe all relevant hardcoded parameters have been removed (the few standing ones are more "by definition" anyway, so I'm flagging it as such in the overview file.

For the parameters flagged as "undetermined"/"calibration", was this established via the write-up? 

I am not sure I am in a position to judge whether all those universal parameters can be labelled as such, I think this is @tbhallett's call!

PR #1743, opened 2025-10-30
Update Level 1b consumable availability to reflect 1b-2 merge

sakshimohan:
This PR updates the RF for consumable availability for level 1b facilities to reflect the merging of level 1b and 2 in simulations.

### DIFF
--- a/src/scripts/data_file_processing/healthsystem/consumables/generating_consumable_scenarios/generate_consumable_availability_scenarios_for_impact_analysis.py
+++ b/src/scripts/data_file_processing/healthsystem/consumables/generating_consumable_scenarios/generate_consumable_availability_scenarios_for_impact_analysis.py
@@ -37,7 +37,6 @@ extracted from the Harmonised Health Facility Assessment 2018/19.
 Consumable availability is measured as probability of consumable being available at any point in time.
 """
 import datetime
-import os
 from collections import defaultdict
 from pathlib import Path
 
@@ -80,7 +79,7 @@ def generate_alternative_availability_scenarios(tlo_availability_df: pd.DataFram
     # Get TLO Facility_ID for each district and facility level
     mfl = pd.read_csv(resourcefilepath / "healthsystem" / "organisation" / "ResourceFile_Master_Facilities_List.csv")
     districts = set(pd.read_csv(resourcefilepath / 'demography' / 'ResourceFile_Population_2010.csv')['District'])
-    fac_levels = {'0', '1a', '1b', '2', '3', '4'}
+    fac_levels = {'0', '1a', '1b', '2', '3', '4'}  # noqa: F841
     tlo_availability_df = tlo_availability_df.merge(mfl[['District', 'Facility_Level', 'Facility_ID']],
                         on = ['Facility_ID'], how='left')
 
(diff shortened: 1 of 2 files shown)


sakshimohan:
Availability at level 1b is updated with average of availability (by item_code and month) for levels 1b and 2. Three strategies of weighting are adopted to test calibration - 
	1.	level2 – fully replace Level 1b availability with Level 2 values. `ResourceFile_Consumables_availability_small_level2.csv`
	2.	national_1b_to_2_ratio – apply a single national-level 1b:2 facility ratio for weighting. `ResourceFile_Consumables_availability_small_national_1b_to_2_ratio.csv`
	3.	district_1b_to_2_ratio – apply district-specific 1b:2 ratios for weighting. `ResourceFile_Consumables_availability_small_district_1b_to_2_ratio.csv`

`ResourceFile_Consumables_availability_small` contains the original level 1b and 2 availability.

sakshimohan:
Hi @tbhallett. As discussed earlier this week, I've added three new versions of the consumable availability RF on this branch to see how each of these performs in comparison with data. 
The average availability by level in the 4 cases is as follows - 

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/sm2511/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">


</head>

<body link="#0563C1" vlink="#954F72">


Level | Actual | level2 | national_1b_to_2_ratio | district_1b_to_2_ratio
-- | -- | -- | -- | --
0 | 0.22 | 0.22 | 0.22 | 0.22
1a | 0.44 | 0.44 | 0.44 | 0.44
1b | 0.58 | 0.70 | 0.61 | 0.63
2 | 0.70 | 0.70 | 0.70 | 0.70
3 | 0.79 | 0.79 | 0.79 | 0.79
4 | 0.38 | 0.38 | 0.38 | 0.38



</body>

</html>

tbhallett:
> > @sakshimohan -- I've looked at the calibration runs for this, and nothing disastrous has happened -- so I think it's OK to proceed.
> 
> Many thanks, @tbhallett! I assume any of the four weighting options are OK? In this case, I'll use `district_1b_to_2_ratio` which weights availability by the ratio of 1b to 2 facilities within each district, because this feels most sound conceptually.

Ok, that's fine to proceed then.

I also made this graph to help (me) remember the difference between that scenario and what's in master:

facility_scatter_plots

sakshimohan:
> > > @sakshimohan -- I've looked at the calibration runs for this, and nothing disastrous has happened -- so I think it's OK to proceed.
> > 
> > 
> > Many thanks, @tbhallett! I assume any of the four weighting options are OK? In this case, I'll use `district_1b_to_2_ratio` which weights availability by the ratio of 1b to 2 facilities within each district, because this feels most sound conceptually.
> 
> Ok, that's fine to proceed then.
> 
> I also made this graph to help (me) remember the difference between that scenario and what's in master:

This is really helpful. Just to note the reason that the availability is so vastly different for Salima, Ntchisi, Mwanza, Machinga, and Balaka (Facility IDs 2, 54, 70 102, 114) is that no facilities at level 1b were recorded in these districts as per HHFA. This means that the availability at level 1b is equated to that at level 2. 
If we want to avoid this extreme case, I can just use the `national_1b_to_2_ratio` for the weighted average.

tbhallett:
here's the comparison of how availability changes for the level demarked as '1b' (which are used for '1b' and for '2')

facility_scatter_plots

PR #1753, opened 2025-11-17
Generate chain of events for individuals inc. for mni dataframe

marghe-molaro:
This PR updates previous PR draft #1468

Now includes changes in the mother_and_newborn_info dataframe.

### DIFF
--- a/src/tlo/methods/consumables.py
+++ b/src/tlo/methods/consumables.py
@@ -6,6 +6,7 @@ from typing import Dict, List, Optional
 
 import numpy as np
 import pandas as pd
+
 from tlo import logging
 from tlo.notify import notifier
 
--- a/src/tlo/notify.py
+++ b/src/tlo/notify.py
@@ -66,7 +66,7 @@ class Notifier:
         e.g. if you are running multiple tests or simulations in the same process.
         """
         self.listeners.clear()
-        
+
     def has_listeners(self, notification_key):
         """
         Check if there are any listeners registered for a specific notification.


marghe-molaro:
Tidied up the logging of chains so that this is now in EAV format.

Awaiting broadcasting module before finalising this.

src/tlo/methods/collect_event_chains.py
```diff
@@ -0,0 +1,330 @@
...
+    def on_notification_event_about_to_run(self, data):
+        """Do this when notified that an event is about to run. This function checks whether this event should be logged as part of the event chains, and if so stored required information before the event has occurred. """
+
+        p = self.parameters
+
+        # Only log event if
+        # 1) generate_event_chains is set to True
+        # 2) the event belongs to modules of interest and
+        # 3) the event is not in the list of events to ignore
+        if p['generate_event_chains'] and (data['module'] in p['modules_of_interest']) and (data['link_info']['EventName'] not in p['events_to_ignore']):
```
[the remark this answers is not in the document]
marghe-molaro: Thanks, and yes good point!

src/tlo/methods/collect_event_chains.py
```diff
@@ -0,0 +1,370 @@
...
+        pass
+        
+    def on_notification_pop_has_been_initialised(self, data):
+
+        # When logging events for each individual to reconstruct chains,
+        # only the changes in individual properties will be logged.
+        # At the start of the simulation + when a new individual is born,
+        # we therefore want to store all of their properties
+        # at the start.
+        if self.generate_event_chains:
```
tamuri: I'd always start these with the negating condition, and exit. Then you get an unindented block of code for the interesting logic. i.e. 

```
if not self.generate_event_chains:
    return

ednav = ...
```

`ednav` is not a great variable name (EAV is a well-known structure for data)

src/tlo/methods/collect_event_chains.py
```diff
@@ -0,0 +1,370 @@
...
+        
+        # Initialise these variables
+        self.print_chains = False
+        self.df_before = []
+        self.row_before = pd.Series()
+        self.mni_instances_before = False
+        self.mni_row_before = {}
+        self.entire_mni_before = {}
+        
+        self.print_chains = True
```
tamuri: You just set print_chains to False above

src/tlo/methods/hsi_event.py
```diff
@@ -193,13 +194,41 @@ def _run_after_hsi_event(self) -> None:
                 item_codes=self._EQUIPMENT,
                 facility_id=self.facility_info.id
             )
+        
 
     def run(self, squeeze_factor):
         """Make the event happen."""
+        
+        # Dispatch notification that HSI event is about to run
+        notifier.dispatch("event.about_to_run",
```
tamuri: A notification key should only come from one place. As suggested elsewhere, change this to `HSI_Event.pre-run` and the one after to `HSI_Event.post-run`

src/tlo/methods/hsi_event.py
```diff
@@ -193,13 +194,41 @@ def _run_after_hsi_event(self) -> None:
...
         self.post_apply_hook()
         self._run_after_hsi_event()
+
+        # Dispatch notification that HSI event has just ran
+        if updated_appt_footprint is not None:
+            footprint = updated_appt_footprint
+        else:
+            footprint = self.EXPECTED_APPT_FOOTPRINT
+            
+        if self.facility_info:
```
tamuri: This logic could potentially go in the callback

src/tlo/methods/hsi_event.py
```diff
@@ -193,13 +194,41 @@ def _run_after_hsi_event(self) -> None:
...
+        
 
     def run(self, squeeze_factor):
         """Make the event happen."""
+        
+        # Dispatch notification that HSI event is about to run
+        notifier.dispatch("event.about_to_run",
+                          data={"target": self.target,
+                                "module" : self.module.name,
+                                "link_info" : {"EventName": type(self).__name__}})
```
tamuri: Use `self.__class__.__name__` to get the class name

src/tlo/methods/rti.py
```diff
@@ -2776,7 +2776,7 @@ class RTIPollingEvent(RegularEvent, PopulationScopeEventMixin):
     def __init__(self, module):
         """Schedule to take place every month
         """
-        super().__init__(module, frequency=DateOffset(months=1))
+        super().__init__(module, frequency=DateOffset(months=1000)) # Single polling event
```
tamuri: Revert this back to the original value of 1...?

src/tlo/simulation.py
```diff
@@ -298,9 +300,17 @@ def initialise(self, *, end_date: Date) -> None:
...
         for module in self.modules.values():
             module.initialise_simulation(self)
         self._initialised = True
+        
+        # Since CollectEventChains listeners are added to notified upon module initialisation,
+        # this can only be dispatched here.
+        # Otherwise, would have to add listener outside of CollectEventChains initialisation
+        
+        # Dispatch notification that pop has been initialised
+        notifier.dispatch("simulation.pop_has_been_initialised", data={"EventName" : "StartOfSimulation"})
```
tamuri: Make notification key `Simulation.post-initialise`. You don't need to return any data here. The registered listener should know it's being triggered by the notification it's listening for.

tamuri:
> There is one outstanding issue, which is that when running a scenario it runs fine but it fails to create a log file. No error message suggesting why this might be failing, but I was wondering whether this could be related to the issue previously raised around saving and loading sim outputs in PR #1759 ?

I don't think so. Logging should work as previously. Is there no output at all, or only the event history isn't being logged?There are quite a few settings turning logging on/off. Might need to step through the code to see what's going on.

src/tlo/methods/collect_event_chains.py
```diff
@@ -0,0 +1,370 @@
...
+        # only the changes in individual properties will be logged.
+        # At the start of the simulation + when a new individual is born,
+        # we therefore want to store all of their properties
+        # at the start.
+        if self.generate_event_chains:
+
+            # EDNAV structure to capture status of individuals at the start of the simulation
+            ednav = df_to_EAV(self.sim.population.props, self.sim.date, 'StartOfSimulation')
+
+            logger.info(key='event_chains',
```
tamuri: Looking at the output, it's not in the right format. Currently, each cell of the logged entry is a dict. 

What we want is a flat table of entries with columns: E, A, V, EventName. And one single entity+attribute+value+EventName per row. So, I think you need to loop over each of the changed attributes and log them one at a time. Otherwise, you'll have the same problem of unknown varying number of items in each logged row.

tamuri (commented):
Another thing I noticed. It's late so I might not be making sense -easier to speak on Friday.

tamuri:
> Are the copies of pop dataframe/rows "deep" enough? Do columns containing series and lists need to be copied iteratively?

The following population dataframe columns have object type i.e. the ones to check are copied/checked properly:

```
>>> [col for col in df.columns if df[col].dtype == 'object']
['rt_injuries_to_cast', 'rt_injuries_for_minor_surgery', 'rt_injuries_for_major_surgery', 'rt_injuries_to_heal_with_time', 'rt_injuries_for_open_fracture_treatment', 'rt_injuries_left_untreated']
```

src/tlo/methods/consumables.py
```diff
@@ -249,7 +251,11 @@ def _request_consumables(self,
                              essential_item_codes: dict,
                              optional_item_codes: Optional[dict] = None,
                              to_log: bool = True,
+                             to_broadcast: bool = True,
```
tamuri: I think you don't need the `to_broadcast` argument here (I see that you check elsewhere where the individual tracker is registered) and simply use the existing `to_log? The notifier should always broadcast messages. Then the listeners register themselves to receive the message. If nothing is listening, there is no error. The obvious example is if a different module wants to listen for notifications.

src/tlo/methods/individual_history_tracker.py
```diff
@@ -0,0 +1,476 @@
...
+        "modules_of_interest": Parameter(
+            Types.LIST, "Restrict the events collected to specific modules. If *, print for all modules"
+        ),
+        "events_to_ignore": Parameter(
+            Types.LIST, "Events to be ignored when collecting chains"
+        ),
+    }
+    
+    PROPERTIES = {
+        "track_history": Property(Types.BOOL, "Whether the individual should be tracked by the individual history tracker or not")
```
tamuri: Needs a prefix - `iht_`?

src/tlo/methods/consumables.py
```diff
@@ -282,28 +288,31 @@ def _request_consumables(self,
...
+                    },
+                    description="Record of requested and used consumable items."
+                )
+                self._summary_counter.record_availability(
+                    items_available=items_available,
+                    items_not_available=items_not_available,
+                    items_used=items_used,
+                )
+            
+            notifier.dispatch("consumables._request_consumables", data={'target' : target, 'event_name' : event_name, 'Item_Available': str(items_available),'Item_NotAvailable': str(items_not_available), 'Item_Used': str(items_used)})
```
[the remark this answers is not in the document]
marghe-molaro: So unlike elsewhere, where the broadcasting happens before or after the named function is called, in this case the broadcasting occurs inside the `_request_consumable` fnc...so I opted to rename it `consumables.on_request-consumables`, but please let me know if you disagree

marghe-molaro:
> A few suggestions below...
> 
> Looking at the logs, I don't know how common is the following scenario: same individual, repeated same event on the same dat i.e. you would get a number of entries that were potentially of a different execution of the same event. If so, you could add an incrementing counter that gets logged as an identifier to group the entries in the EAV table. Might not be necessary if this doesn't happen.

Thanks @tamuri, I have now included a sim-wide counter of events being ran that is appended as a unique identifier to the event name when logged (this doesn't apply to birth events, which I think we can safely assume only occur once per individual!)

I have additionally included a similar counter - which in this case is however reset for each event - which tracks the instance of consumable requests within that HSI, ensuring that if multiple consumable calls occur within the same HSI_Event these can be logged separately.

src/scripts/track_individual_histories/analysis_extract_data.py
```diff
@@ -0,0 +1,82 @@
...
+    print(filtered)
+
+
+def apply(results_folder: Path, output_folder: Path, resourcefilepath: Path = None, ):
+    """Extract event chains
+    """
+    pd.set_option('display.max_rows', None)
+    pd.set_option('display.max_colwidth', None)
+
+    individual_individual_histories = extract_individual_histories(results_folder)
```
tamuri: One of the linting errors is that this variable is unused. Will you put something here, or remove capturing result of `extract_individual_histories`?

PR #1756, opened 2025-11-20
Include option to use emulator in RTI module

marghe-molaro:
This PR offers users the option to use an emulated version of the RTI module. It will require #1758 to run.

@tamuri, a couple of "conceptual" issues:
- One of the advantages of using an emulator is the need for much fewer module-specific properties in the population dataframe. However, because properties are declared _before_ parameters are initialised, it is not possible for the module to know a priori whether the user will opt to use the emulator or not, and hence how many properties to declare. Do you have any suggestions? Could I initially declare all properties and then remove redundant ones if parameter 'use_RTI_emulator' is initialised to `True`? (Not very elegant)
- As discussed yesterday, I wanted to include a one-time warning for the user if the emulator is being used under the "wrong" HealthSystem assumptions. This would however require knowing how the HealthSystem parameters have been initialised, but this would require relying on a known order of parameter initialisation. Where can I issue this warning while avoid issuing this repeatedly during runtime? 
- There is some RTI-specific logic in the HealthSystem; in the future I think we would instead iterate over all modules to check if the use_{module}_emulator parameter is set to True for any of them. But is this ok for now?

Many thanks, 
Margherita


src/tlo/methods/demography.py
```diff
@@ -758,7 +759,10 @@ def apply(self, population):
         mort_risk = self.mort_risk_per_poll.loc[
             self.mort_risk_per_poll.fallbackyear == fallbackyear, [
                 'age_years', 'sex', 'prob_of_dying_before_next_poll']].copy()
-
+        
+        # Artificially increase risk for men under 50 by 50
```
tamuri: This should probably be removed for the merge...?

src/tlo/methods/healthsystem.py
```diff
@@ -2739,6 +2768,7 @@ def record_hs_status(
     def write_to_log_and_reset_counters(self):
         """Log summary statistics reset the data structures. This usually occurs at the end of the year."""
 
+
```
tamuri: Remove newline

src/tlo/methods/rti.py
```diff
@@ -1020,70 +1047,85 @@ def __init__(self, name=None):
         'maximum_number_of_times_HSI_events_should_run': Parameter(
             Types.INT,
             "limit on the number of times an HSI event can run"
+        ),
+        'use_RTI_emulator': Parameter(
```
tamuri: Rename this to `use_emulator` - you're already in RTI

tamuri (changes requested):
Thanks, Marghertia. We could tidy this up by having a new EmulatedRTI class that is separate from the RTI class. If there are things that need to be shared, you could make a base class (e.g. BaseRTI) and have both RTI and EmulatedRTI inherit that. But you may not need to because most of the code is not run in the EmulatedRTI.

Each on these classes could have their own polling event class or, if they too share code (looks like it does) the common bits could be extracted into a shared helper function or, similarly, make a BaseRTIPollingEvent and have the simulated and emulated event class use/inherit.

This would avoid the long if-else blocks and keeps the emulated RTI stuff together and separate. You might not need the 'use_emulated' in the end because a different module will be registered at simulation start.

marghe-molaro:
Following Q4 discussion:
- Add warning about parameters in emulator - i.e. which you can change and which not
- Workflow so that they can be retrained automatically
- Add link between emulator and git commit from which training data was generated

marghe-molaro:
Close for future development

PR #1758, opened 2025-11-20
Add SDV as dependency and update requirements file

### LINKED ISSUE #1755: Add SDV dependency to requirements
Add https://sdv.dev/ to get ready for Margherita's RTI emulator.

tamuri:
Resolves #1755. I have avoided updating the other dependencies, especially numpy and pandas as these have big effect on the epi modules. However, had to update virtualenv because sdv requires a more recent version of platformdirs.

Also, `dev.txt` is also constrained by `base.txt` when using pip-compile to ensure consistency.

@marghe-molaro - the resolved version of sdv (1.20.1) was the most recent that doesn't disturb the other packages. Can you check these requirement still work for your emulator, please? You can create a new conda environment to test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "azure-storage-file-share",
     # For saving and loading simulation state
     "dill",
-    "sdv==1.24.1",
+    "sdv",
 ]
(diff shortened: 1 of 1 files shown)


tamuri:
Thanks for checking - please approve!

Issue #1770, opened 2025-12-18
Computation of rescaling factors is incorrect


sangeetabhatia03:
The current computation of rescaling factors is incorrect, as it does not take into account clinic split of capabilities.
To fix this, we need to edit the function `frac_time_used_by_facID_and_officer`

marghe-molaro:
This Issue was a addressed by PR #1781 and therefore can be close.

PR #1773, opened 2026-01-07
Fix computation of fraction time used by facility ID and officer type for clinic specific rescaling factors

sangeetabhatia03:
This PR fixes issue #1770.
In the implementation of clinics, we did not pass clinic information to the function `frac_time_used_by_facID_and_officer`, and therefore the computation  therein was not clinic-specific, leading to erroneous calculations.
We have now fixed this and updated other necessary objects (namely `running_total_footprint`) to be clinic-specific.
This PR also brings in changes from PR #1662  and PR #1689 because in the absence of these, we would have to make the whole squeeze factor computation clinic-specific, when it is intended to be removed shortly.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -2271,6 +2271,28 @@ class HealthSystem(Module):
                         priority=_priority,
                         clinic=event_clinic,
                     )
+                                   # if not ok_to_run
+                else:
+                    # Do not run,
+                    # Call did_not_run for the hsi_event
+                    rtn_from_did_not_run = event.did_not_run()
+                    # If received no response from the call to did_not_run, or a True signal, then
+                    # add to the hold-over queue.
+                    # Otherwise (disease module returns "FALSE") the event is not rescheduled and will not run.
+
+                    if rtn_from_did_not_run is not False:
+                        # reschedule event
+                        hp.heappush(_to_be_held_over, _list_of_individual_hsi_event_tuples[ev_num])
+
+                    # Log that the event did not run
+                    self.record_hsi_event(
+                        hsi_event=event,
+                        actual_appt_footprint=event.EXPECTED_APPT_FOOTPRINT,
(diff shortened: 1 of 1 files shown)


tests/test_healthsystem.py
```diff
@@ -2892,3 +2828,176 @@ def initialise_simulation(self, sim):
...
+        )
+        sim.modules["HealthSystem"]._clinic_names = ["Clinic1", "GenericClinic"]
+        sim.modules["HealthSystem"].setup_daily_capabilities("funded_plus")
+
+        # Assign the entire population to the first district, so that all events are run in the same district
+        col = "district_of_residence"
+        s = sim.population.props[col]
+        ## Not specifying the dtype explicitly here made the col a string rather than a category
+        ## and that caused problems later on.
+        sim.population.props[col] = pd.Series(s.cat.categories[0], index=s.index, dtype=s.dtype)
```
marghe-molaro: The district of residence should be derived from the location associated with the selected Facility_ID, so that we can be sure the district to which the population is assigned matches the district of the facility

tests/test_healthsystem.py
```diff
@@ -2892,3 +2828,176 @@ def initialise_simulation(self, sim):
...
+                policy_name="",
+                use_funded_or_actual_staffing="funded_plus",
+            ),
+            DummyModuleGenericClinic(),
+            DummyModuleClinic1(),
+        )
+        sim.make_initial_population(n=tot_population)
+
+        sim.modules["HealthSystem"]._clinic_configuration = pd.DataFrame(
+            [{"Facility_ID": 20.0, "Officer_Type_Code": "DCSA", "Clinic1": 0.6, "GenericClinic": 0.4}]
```
marghe-molaro: Since this 0.6/0.4 split is arbitrary/never used, I would move the initialisation of the capabilities and clinics split here for facility 20 here (based on your intended plan of number of appts that you expect to be delivered in each).

PR #1785, opened 2026-01-21
Adding herbal medication to lifestyle module

### LINKED ISSUE #1784: Incorporating herbal medication use into the Lifestyle module
**Background:** 

Herbal medication use is a common health-related behaviour in Malawi and has been identified as a risk factor for several conditions already represented in the TLO model, particularly chronic kidney disease, as well as liver disease, cardiovascular disease, and bleeding disorders. Currently, herbal medication use is not explicitly captured in the Lifestyle module, which may lead to underestimation of risk in relevant disease pathways.

**Proposed change:**
Introduce herbal medication use as a lifestyle exposure within the Lifestyle module, represented as a categorical or binary variable (e.g. current use / no use), with stratification by rural vs urban residence.

**Intended model linkages**
The herbal medication exposure would modify transition probabilities or risk parameters in the cmd_chronic_kidney_disease module and in the enhanced lifestyle module

**Population stratification**
•	Rural vs urban residence

**Data considerations (placeholders)**
•	Prevalence of herbal medication use: DHS, local surveys, or published Malawian studies
•	Relative risks / hazard ratios: regional or international literature where local data are unavailable

upile-n:
Fixes #1784 
Introduce herbal medication use as a lifestyle exposure within the Lifestyle module, represented as a categorical or binary variable (e.g. current use / no use), with stratification by rural vs urban residence.


tbhallett:
Hi @upile-n  --- that sounds a  great idea. I think you've haven't committed/pushed your changes onto this branch though, as the only change we see for now is a comment in the cardio-metabolic-disorders module.

Please ping up when all the changes are here and ready for review.

andrew-phillips-1:
This seems fine but I was thinking you could have it even simpler than this and have herbal medication use as a fixed personal property, like sex or urban/rural, so you just assign each person in 2010 or at birth with a herbal medication status that applies for life.

src/tlo/methods/enhanced_lifestyle.py
```diff
@@ -337,14 +359,16 @@ def __init__(self, name=None):
         'li_date_acquire_clean_drinking_water': Property(Types.DATE, 'date acquire clean drinking water'),
         'li_date_acquire_non_wood_burn_stove': Property(Types.DATE, 'date acquire non-wood burning stove'),
         "li_is_sexworker": Property(Types.BOOL, "Is the person a sex worker"),
-        "li_is_circ": Property(Types.BOOL, "Is the person circumcised if they are male (False for all females)"),
+        "li_is_circ": Property(Types.BOOL, "Is the person circumcised if they are male (False for all females)"
+                               ),
+        'li_herbal_medication': Property(Types.BOOL, 'whether someone uses herbal medication or not'),
```
tbhallett: I would add the word 'currently' in here (if we're to keep the current structure, of whether use can start and stop)

tbhallett (commented):
Agreed with Andrew!
* This is nicely coded and looks like it will work fine. WELL DONE! 
* I'm intersetd in the decison to make it a property that can change versus one that is fixed for an individual. Is there a good reason to go one way or the other? Let's have a think and discuss
* For the PR, we'd also need to :
    * have a look at plots showing what this property does - you've made a good start here. (prevalence of it, by urban/rural over time)
    * have a justifcation for any new parameters included in the write-up document.

andrew-phillips-1:
I think one thing to consider here is whether we are thinking of this property as "use of the types of herbal medications that cause increased renal problems" or something more general (I would vote for the former I think).  I was seeing that there are various herbal medications used and not all are thought to contribute to renal problems.  Once that is defined then perhaps Upile you can do another search (and great that you have access to all articles now through UCL if you need them) and look for estimates of prevalence by urban/rural, but I think you may well find none.  It seems to be clear that use is highly differential by urban/rural but perhaps you can find some sources that state that, even if not quantitative.  I think we are probably going to have to say that we have gone with target of 80% in rural areas because it is clear that use is widespread although not universal.  Whilst the 10% in urban areas reflect that use is much more limited.  We have assumed that it tends to be a behaviour associated with a person over their hole like and effects are long term so we elected to make it a fixed property of an individual ?  Let me know if you disagree though.

andrew-phillips-1:
Yes if we think this is mainly about renal disease then perhaps the property can be in the kidney transplant module.

upile-n:
> I think one thing to consider here is whether we are thinking of this property as "use of the types of herbal medications that cause increased renal problems" or something more general (I would vote for the former I think). I was seeing that there are various herbal medications used and not all are thought to contribute to renal problems. Once that is defined then perhaps Upile you can do another search (and great that you have access to all articles now through UCL if you need them) and look for estimates of prevalence by urban/rural, but I think you may well find none. It seems to be clear that use is highly differential by urban/rural but perhaps you can find some sources that state that, even if not quantitative. I think we are probably going to have to say that we have gone with target of 80% in rural areas because it is clear that use is widespread although not universal. Whilst the 10% in urban areas reflect that use is much more limited. We have assumed that it tends to be a behaviour associated with a person over their hole like and effects are long term so we elected to make it a fixed property of an individual ? Let me know if you disagree though.

I would opt for the more general approach so that this property can be used in other modules, and not just chronic kidney disease, whether now or in future. I am, however, open to discuss that further, if having specific types of herbal medication could have an added benefit. 

I will be sure to do a deep dive search for the urban/rural estimates.

tbhallett:
Nice plots. I'm sure we could live with this implementation as a first pass. (a future PR could resample upon moving; or upon reaching maturity, etc.).

Next step would be to convince us that these are the right levels of herbal medicine use: i.e., compare the model outputs to data.

PR #1786, opened 2026-01-21
Pass module name with consumables when tracking individual histories

marghe-molaro:
Fix dispatching of consumable access when tracking individual histories: it should be module's name, not module, that is passed to the dispatcher.

Additionally include small style fixes to the consumable logging (no need to log HSI name and module during cons accessed as this is redundant, and create a dictionary of useful info for each cons access rather than attaching cons access counter to info's name).


marghe-molaro:
@tamuri tests are passing locally

PR #1790, opened 2026-01-29
Weather Disruptions Module

RachelMurray-Watson:
This PR aims to develop a separate "WeatherDisruptions" module. This modularizes code developed in rmw/climate_change_switch and rmw/climate_troubleshoot_LHS. 

The module aims to do the following: 

1. Assign individuals facilities based on their districts*
2. Read in data on the weather disruptions**
3. Determine whether a disruption will occur (and consequence e.g. reseeking healthcare)
4. Feed this back to the health system
5. Log disrupted HSIs by district

* This is a necessary but, also, distinct functionality that could instead be included in the demography module.
** Currently these are csv files with the proportion of ANC appointments disrupted for each facility by month. @marghe-molaro previously suggested saving the coefficients of the ANC model instead and doing the LM each time. This will probably save us space (as we would only need the details for each facility to be loaded in) but it seems more computationally intensive (as would need to look up all details for each facility and do calculation each time  a HSI is processed). 

To Do: 

- [x] Framework of WeatherDisruptions module
- [x] Parameter file separated from health system file
- [x] Tests of functionality 
- [x] Decide on points * and ** above 
- [x] Add new logging to HealthBurden module


RachelMurray-Watson:
I think this is finally ready for its initial review @marghe-molaro !

src/tlo/methods/demography.py
```diff
@@ -100,8 +102,6 @@ def __init__(self, name=None, equal_allocation_by_district: bool = False):
     # We should have 21 age range categories
     assert len(AGE_RANGE_CATEGORIES) == 21
 
-    # Here we declare parameters for this module. Each parameter has a name, data type,
```
marghe-molaro, line 103: Wondering why this description was removed in this PR?

src/tlo/methods/demography.py
```diff
@@ -143,13 +143,30 @@ def __init__(self, name=None, equal_allocation_by_district: bool = False):
...
             categories=['SET_AT_RUNTIME']
         ),
-
         'region_of_residence': Property(
             Types.CATEGORICAL,
             'The region of residence (mapped from district_num_of_residence).',
             categories=['SET_AT_RUNTIME']
         ),
 
+        # Facility level assignments; categories are set in `pre_initialise_population`
```
marghe-molaro, line 152: In an ideal world, we would look up facility levels declared in the HS module and then declare this dynamically. e.g. if in the HS facility levels declared are `facility_levels = ["0", "1a", "1b", "2", "3"]`, you would create an array of properties, called something like `assigned_real_facility_for_facility_level`, and populate it later based on of length f`acility_levels`, where `assigned_real_facility_for_facility_level[0]` refers to facility level `facility_level[0]`. Might not be the time to do this now, but could you add a note on this on the HealthSystem review doc?

src/tlo/methods/demography.py
```diff
@@ -241,6 +270,31 @@ def pre_initialise_population(self):
             categories=self.parameters['pop_2010']['Region'].unique().tolist()
         )
 
+        # Map each TLO facility level to the facility types it encompasses
+        facility_levels_types = {
```
marghe-molaro, line 274: Similarly to comment above, this should come from the HS, i.e. not be redefined here. Not sure why you need this info?

src/tlo/methods/demography.py
```diff
@@ -366,6 +420,13 @@ def on_birth(self, mother_id, child_id):
         _district_of_residence = df.at[_id_inherit_from, 'district_of_residence']
         _region_of_residence = df.at[_id_inherit_from, 'region_of_residence']
 
+        # Inherit facility assignments from parent
```
marghe-molaro, line 423: This seems a bit verbose/unnecessary if only used here, you can just call directly?

src/tlo/methods/demography.py
```diff
@@ -527,6 +765,7 @@ def do_death(self, individual_id: int, cause: str, originating_module: Module):
             'cause': str(cause),
             'label': self.causes_of_death[cause].label,
             'person_id': individual_id,
+            'district_of_residence': person['district_of_residence'],
```
marghe-molaro, line 768: Is this a change that we want to bring into master for sure? Or more relevant to your particular analysis?

src/tlo/methods/fullmodel.py
```diff
@@ -116,6 +118,8 @@ def fullmodel(
         copd.Copd,
         depression.Depression,
         epilepsy.Epilepsy,
+        # Climate-change related modules
+        weather_disruptions.WeatherDisruptions
```
marghe-molaro, line 122: Similarly to simplified births, if you are declaring a variable for inclusion (`use_weather_disruption`), you could consider only including the relevant module if this is true.

src/tlo/methods/fullmodel.py
```diff
@@ -36,11 +36,13 @@
     symptommanager,
     tb,
     wasting,
+    weather_disruptions,
```
marghe-molaro, line 39: I'm not actually 100% sure we want the weather_disruptions to be included in the full model for now. I think a user interested in including them could more simply invoke the fullmodel and add the `weather_disruptions` module in the scenario file, unless we want to make weather disruptions a standard component of the full model. @tbhallett is better placed to advise.

If we do want to include this module in the full model, I suggest always including it and then declaring a parameter in that module e.g. `include_weather_disruptions` to decide whether they are considered or not. I think `use_simplified_births` option when invoking the full model is mainly used to ensure mutually exclusive sets of modules (simplified births vs everything else maternity/neonatal related) are not included at the same time

src/tlo/methods/healthsystem.py
```diff
@@ -404,44 +404,6 @@ def __init__(
         disable_and_reject_all: bool = False,
         hsi_event_count_log_period: Optional[str] = "month",
     ):
-        """
```
marghe-molaro, line 407: I'm a bit confused, it looks like you are removing all the HealthSystem parameters here, but not re-adding them? It must be a weird git glitch

src/tlo/methods/healthsystem.py
```diff
@@ -544,20 +506,12 @@ def __init__(
 
         self._hsi_event_count_log_period = hsi_event_count_log_period
         if hsi_event_count_log_period in {"day", "month", "year", "simulation"}:
-            # Counters for binning HSI events run (by unique integer keys) over
```
marghe-molaro, line 547: It looks like there are a lot of comments that are being removed from the HealthSystem module...just wondering why?

src/tlo/methods/healthsystem.py
```diff
@@ -2373,97 +1949,75 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
-                    # Run event immediately.
-
-                    # Retrieve officers&facility required for HSI
+                    # The event is due today and the person is alive.
                     original_call = next_event_tuple.hsi_event.expected_time_requests
                     _priority = next_event_tuple.priority
 
-                    # Check if any of the officers required have run out.
+                    climate_disrupted = False
+                    if 'WeatherDisruptions' in self.sim.modules:
```
marghe-molaro, line 1988: Going back to the 'fullmodel' comment: this would be suboptimal because this would always be true by default. But if following my first suggestion (i.e. we don't need the WeatherDisruptions in the fullmodel) then absolutely fine

src/tlo/methods/healthsystem.py
```diff
@@ -2373,97 +1949,75 @@ def process_events_mode_2(self, hold_over: List[HSIEventQueueItem]) -> None:
...
+                        climate_disrupted, is_supply_side = (
+                            self.sim.modules['WeatherDisruptions'].check_hsi_for_disruption(
+                                hsi_event_item=next_event_tuple,
+                                current_date=self.sim.date
+                            ) if 'WeatherDisruptions' in self.sim.modules else (False, False)
```
marghe-molaro, line 1993: Same as above

marghe-molaro (changes requested):
I'm just including a few comments for now.

Main thing to note at present is that this PR seems to be introducing a large number of potentially unwanted changes to the HealthSystem (e.g. removing all module parameters)

PR #1802, opened 2026-02-05
Enable switching service availability in healthsystem module

sangeetabhatia03:
This pull request enables switching of service availability in the healthsystem module. Key changes include the addition of new parameters for service availability switching.

I have also created a scenario file for our initial analysis; would be great to get your feedback on this.


tbhallett (commented):
Nice logic and nice test.

Personal taste but I would suggest simplifying the scenario file so that we have everything in one file. (The style of having it split across multiple files is useful when we have lots of scenario file drawing from the same 'base', but it feel a bit confusing here).

src/scripts/lcoa_inputs_from_tlo_analyses/scenarios_definitions.py
```diff
@@ -0,0 +1,45 @@
...
+
+                    "mode_appt_constraints": 1,
+                    "year_service_availability_switch": self.YEAR_OF_SERVICE_AVAILABILITY_SWITCH,
+
+                    # allow historical HRH scaling to occur 2018-2024
+                    # 'year_HR_scaling_by_level_and_officer_type': self.YEAR_OF_SERVICE_AVAILABILITY_SWITCH,
+                    'yearly_HR_scaling_mode': 'historical_scaling',
+                },
+
+                "ImprovedHealthSystemAndCareSeekingScenarioSwitcher": {
```
marghe-molaro: Just checking - did you check that these are compatible with the suspend/resume on Azure?

marghe-molaro (changes requested):
Hi @sangeetabhatia03, this is looking great, I've just added a few comments!
Summary:
- I agree with Tim about having a single scenario file
- I've suggested tightening the test just a little bit, it shouldn't take too long!

src/tlo/methods/healthsystem.py
```diff
@@ -1250,17 +1263,28 @@ def format_clinic_capabilities(self) -> pd.DataFrame:
...
+        Computation of these factors is split from the actual rescaling to facilitate
+        capturing them even when running the model in mode 1."""
+        self._rescaling_factors = defaultdict(dict)
+        for clinic, clinic_cl in self._daily_capabilities.items():
+            for facID_and_officer in clinic_cl.keys():
+                self._rescaling_factors[clinic][facID_and_officer] = self._summary_counter.frac_time_used_by_facID_and_officer(
+                    facID_and_officer=facID_and_officer, clinic=clinic
+                )
+        self._summary_counter._rescaling_factors = self._rescaling_factors
+
```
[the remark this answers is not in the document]
sangeetabhatia03: yes, so that we can log rescaling factors.

src/scripts/lcoa_inputs_from_tlo_analyses/scenarios_definitions.py
sangeetabhatia03: As discussed with @marghe-molaro  parking this check for now because switching to `max_healthsystem_function` and `max_healthcare_seeking` is the same across all draws for our current analysis.

tamuri:
Split HS+HS tests into its own PR & LCOA analyses into another PR.

Issue #1618, opened 2025-05-08
Tidying up _daily_capabilities/capabilities_today in HealthSystem


marghe-molaro:
Currently, in the HealthSystem, the daily HR capabilities available are:
- Initialised via _daily_capabilities
- Accessed via the capabilities_today property, which multiplies the _daily_capabilities by the capabilities_coefficient every time it is accessed.

Originally, the capabilities_coefficient was supposed to represent some productivity scaler, which can be initialised through the module parameter. Currently, however, this capabilities_coefficient is only responsible for rescaling assumed capabilities by the simulated to real 2010 population size ratio. Given that this could/should be done once when the simulation is initialised, and that the discussion of HCW productivity has moved quite significantly from a single coefficient since this was first implemented, suggest merging _daily_capabilities and capabilities_today to avoid confusion.

marghe-molaro:
This will be addressed in the future in [HealthSystem Review](https://docs.google.com/document/d/1IGi-jZRk2pbQCVKxHOTed266i44fLNfPBCYdxskOLvc/edit?usp=sharing), so can close.

PR #1806, opened 2026-02-06
Correct staff count log

marghe-molaro:
When logging the number of staff, the `log_number_of_staff` function compares `capabilities_today` - which rescales the `_daily_capabilities` by the `capabilities_coefficient` - with  `_daily_capabilities_per_staff` - which does not. Unless I am missing something obvious, this would therefore lead to an incorrect estimation of the staff count.

As raised in Issue #1618, it seems dangerous to dynamically incorporate the `capabilities_coefficient` via `capabilities_today`; the rescaling should occur once, upon initialisation, to avoid issues such as this.


marghe-molaro:
Could we merge this for now so that this fix is in, and then I'll get to Issue #1618 asap?

PR #1807, opened 2026-02-06
Split `test_healthsystem.py` into two files to enable parallel running and so speed up testing.

tbhallett:
Here I propose a split of the `test_healthsystem.py` into:

1) `test_healthsystem_general_and_mode1.py`: This test file is focussed on the overall function of the module and its behaviour in Mode 1.
2) `test_healthsystem_mode2.py`: This test file is focussed on behaviour in Mode 2 (including rescaling and clinics).

This feels like a logical way to split the tests. However, the first of these still contains more tests, so may still take much longer. If necessary, a further cut could be made!


tamuri (approved):
Thanks, Tim. This cuts the current test runtime in about half - merging!

PR #1847, opened 2026-03-18
Add monthly daly reporting to individual history tracker

marghe-molaro:
PR to ensure that the monthly dalys reported are included in the individual history tracker, broken down by cuase.

Note: In the healthburden module, we do not automatically dispatch the raw reported dalys dataframe in all instances, but first:
- Check if relevant listener is included;
- Pre-process dispatched data to only include non-zero entries;
This should avoid unnecessarily dispatching large dataframes during runtime.


tbhallett (approved):
All seems fine to me.

Two comments both thinking about speed, but do not feel very attached to either. 

Happy to approve now

tamuri (approved):
Mentioned some small changes, otherwise looks good.

Issue #1595, opened 2025-02-24
Effects of integrating diabetic retinopathy into the wider health system


thewati:
The aim of this is to assess the value and effect of integrating Diabetic Retinopathy (DR) services into Diabetes Mellitus (DM) services and the wider health system. In the TLO model, DM has been modelled in the `cardiometabolicdisorders` module. A separate DR module is created to get all people with DM from the `cardiometabolicdisorders` module and then model/apply all epidemiological effects and health system interactions on them in the separate DR module.

Tasks:
- [ ] Create the method file for DR and capture the symptoms, onset, progression and events (HSIs or otherwise) that occur to people with DR 
- [ ] Create an analysis file for DR to estimate the prevalence and risk factors of DR
- [ ] Create a test file that ensures everything is working
- [ ] Create a resource file to store all parameters
- [ ] Assess current health system for delivery of DR and model the inclusion of DR services within DM services and the wider health system
- [ ] Create a write-up file to document the DR work

PR #1849, opened 2026-03-19
Create date_onset Property in cardio_metabolic_disorders.py

thewati:
Creating date_onset property for conditions in the cardio_metabolic_disorders module. Fixes issue #1845. The nc_diabetes_date_onset property will be used in issue #1595


thewati:
@mnjowe this is now using a dictionary. I'm done

src/tlo/methods/cardio_metabolic_disorders.py
```diff
@@ -1058,6 +1061,12 @@ def schedule_death_to_occur_before_next_poll(p_id, cond):
             idx_acquires_condition = acquires_condition[acquires_condition].index
             df.loc[idx_acquires_condition, f'nc_{condition}'] = True
 
+            # Store onset dates only for diabetes
+            if condition == 'nc_diabetes':
+                for person_id in idx_acquires_condition:
+                    if person_id not in self.module.diabetes_onset_dates:
+                        self.module.diabetes_onset_dates[person_id] = self.sim.date
+
```
tbhallett, line 1069: This is totally fine and very readable. I was thinking that a more pythonic version would be as follows, but this is perhaps less readable (now that I read it after writing it out)

```python
if condition == 'nc_diabetes':
    self.module.diabetes_onset_dates |= {
        person_id: self.sim.date
        for person_id in idx_acquires_condition
        if person_id not in self.module.diabetes_onset_dates
    }
```

tbhallett (approved):

thewati:
Hi @tbhallett,
Sorry I was a bit clumsy. I noticed I missed 2 things in this PR.
1. At line 1065, I accidentally used `if condition == 'nc_diabetes':` instead of `if condition == 'diabetes':`. This is something that changed with one of the rollbacks I did.
2. Somewhere around line 450 (inside initialise_pupulation), I need to add the code below to capture onset dates for initial diabetes prevalence. There was a mismatch between those with diabetes and those with onset dates and this is coming from the the part where I missed this in prevalence:

            `if condition == 'diabetes':
                has_diabetes = df.index[df.is_alive & df.nc_diabetes]

                for person_id in has_diabetes:
                    if person_id not in self.diabetes_onset_dates:
                        self.diabetes_onset_dates[person_id] = self.sim.date`

Can this be undone?

tbhallett:
You could raise a new PR to make amendments if needed.

Sent from Outlook for iOS<https://aka.ms/o0ukef>

PR #1860, opened 2026-05-06
Xin hpv infection process

tdm32:


src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        # Modifiers
+        # "rr_hpv_hiv": Parameter(
+        #     Types.REAL,
+        #     "Relative risk for HPV infection among HIV positive people",
+        # ),
+        "rr_hpv_hiv_no_art": Parameter(
+            Types.REAL,
+            "Relative risk for HPV acquisition among HIV positive people not on ART",
+        ),
+        "rr_hpv_hiv_art_unsuppressed": Parameter(
```
tdm32: is there strong evidence that this value is different to rr_hpb_no_art? If not, then just use the same parameter for both untreated and treated but not virally suppressed

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            Types.REAL,
+            "Weibull shape parameter for HPV clearance duration",
+        ),
+
+        # Modifiers
+        "rr_clear_hiv_no_art": Parameter(
+            Types.REAL,
+            "Rate ratio for HPV clearance among PLWH not on ART",
+        ),
+        "rr_clear_hiv_art_unsuppressed": Parameter(
```
tdm32: as above, would this not be the same value for untreated?

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        #     "Relative risk for reinfection with hr2 if previously infected",
+        # ),
+        # "rr_immunity_hr3": Parameter(
+        #     Types.REAL,
+        #     "Relative risk for reinfection with hr3 if previously infected",
+        # ),
+    }
+
+    PROPERTIES = {
+        'hp_is_infected': Property(
```
tdm32: I don't think you need this property as well as the 3 below

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            Types.BOOL, 'Current infected with hr2'),
+        'hp_infected_hr3': Property(
+            Types.BOOL, 'Current infected with hr3'),
+        'hp_date_infected_hr1': Property(
+            Types.DATE, 'Date of infection of hr1'),
+        'hp_date_infected_hr2': Property(
+            Types.DATE, 'Date of infection of hr2'),
+        'hp_date_infected_hr3': Property(
+            Types.DATE, 'Date of infection of hr3'),
+        'hp_date_first_infected': Property(
```
tdm32: Instead of this property, you can read the earliest start date from the 3 properties above

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            Types.BOOL, 'Ever infected with hr3'),
+
+        'hp_persistent_hr1': Property(
+            Types.BOOL, 'Persistent hr1 infection, duration >= 12 months'),
+        'hp_persistent_hr2': Property(
+            Types.BOOL, 'Persistent hr2 infection, duration >= 12 months'),
+        'hp_persistent_hr3': Property(
+            Types.BOOL, 'Persistent hr3 infection, duration >= 12 months'),
+
+        # "va_hpv": Property(Types.INT, "number of doses of hpv vaccine received"),
```
tdm32: I assume these are here as a reminder for useful properties created by other modules?

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            Types.DATE, 'Date of infection of hr3'),
+        'hp_date_first_infected': Property(
+            Types.DATE, 'Start date of current HPV infection'),
+        'hp_duration_hr1': Property(
+            Types.INT, 'Duration for current hr1 infection'),
+        'hp_duration_hr2': Property(
+            Types.INT, 'Duration for current hr2 infection'),
+        'hp_duration_hr3': Property(
+            Types.INT, 'Duration for current hr3 infection'),
+        'hp_duration_all_clear': Property(
```
tdm32, line 196: I'm not sure I see the use for this property?

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        df.loc[df.is_alive, 'hp_persistent_hr1'] = False
+        df.loc[df.is_alive, 'hp_persistent_hr2'] = False
+        df.loc[df.is_alive, 'hp_persistent_hr3'] = False
+
+        for group in self.HPV_GROUPS:
+            p_init = self.parameters[f'init_prev_hpv_{group}']
+            u = self.rng.random(size=len(eligible))
+            infected_this_group = eligible[u < p_init]
+
+            for person_id in infected_this_group:
```
tdm32: you could remove this loop, indexing properties using df.loc[infected_in_this_group,...] and creating an array of previous infection dates with size=infected_in_this_group etc

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+                # randomly select an infection date for initial population
+                previous_infection = int(self.rng.randint(0, 24))  # 0-23month
+                infection_date = self.sim.date - DateOffset(months=previous_infection)
+
+                df.at[person_id, f'hp_date_infected_{group}'] = infection_date
+                # df.at[person_id, f'hp_date_clear_{group}'] = pd.NaT
+                df.at[person_id, f'hp_duration_{group}'] = previous_infection
+                df.at[person_id, f'hp_persistent_{group}'] = previous_infection >= 12
+
+        initially_infected = df.index[df.is_alive & df.hp_is_infected]
```
tdm32: this part of the code is redundant, you could find the date of first infection at any point by looking up the earliest infection date for an individual when needed. That would avoid creating a new property.

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            labels=self.AGE_LABELS,
+            right=False  # right side not included
+        )
+    def _get_age_group(self, age_years):
+        for i in range(len(self.AGE_BINS)-1):
+            if self.AGE_BINS[i] <= age_years < self.AGE_BINS[i + 1]:
+                return self.AGE_LABELS[i]
+        return None
+
+    def _build_age_mixing_matrix(self, within=0.80, adjacent=0.15, distant=0.05):
```
tdm32: do you have data on mixing patterns for Malawi? We might have it for Zimbabwe but be careful making assumptions on mixing patterns without data

tdm32: all numbers should be in resource file and stored as parameters

tdm32: all values should be in resource file and stored as parameters

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+                for dg in distant_groups:
+                    row[dg] = share_dist
+
+            # normalize to exactly
+            row = row / row.sum()
+            M.loc[label] = row
+        return M
+
+
+    def _months_since(self,start_date,end_date=None):
```
tdm32, line 358: if you have the full dates why use year and then round to 12 months? you can do (date2 - date1).days / 30.5 for example

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            end_date = self.sim.date
+
+        months = (end_date.year - start_date.year) * 12 + (end_date.month - start_date.month)
+        return max(0, int(months))
+
+    def _get_clearance_rr(self, person_id):
+        df = self.sim.population.props
+        p = self.parameters
+
+        if 'hv_inf' not in df.columns:
```
tdm32, line 405: simpler logic might be to check if hiv registered in sim.modules, if not return 1.0 else....

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        # df.at[person_id, f'hp_date_clear_{group}'] = pd.NaT
+        df.at[person_id, f'hp_duration_{group}'] = -1
+        df.at[person_id, f'hp_persistent_{group}'] = False
+
+        still_infected = any(
+            df.at[person_id, f'hp_infected_{group}'] for group in self.HPV_GROUPS
+        )
+        df.at[person_id, 'hp_is_infected'] = still_infected
+
+        if not still_infected:
```
tdm32, line 477: why do we need to track the duration of time clear of infection?

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+                    p['rr_hpv_hiv_art_unsuppressed']
+                ),
+            )
+
+        # add the basic event
+        event = HpvInfectionEvent(self)
+        sim.schedule_event(event, sim.date + DateOffset(months=6))
+
+        # add an event to log to screen
+        sim.schedule_event(HpvLoggingEvent(self), sim.date + DateOffset(months=6, days=1))
```
tdm32: why add another day to this? If scheduled after HpVInfectionEvent it will run immediately after it

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        # df.at[child_id, 'hp_date_clear_hr2'] = pd.NaT
+        # df.at[child_id, 'hp_date_clear_hr3'] = pd.NaT
+        df.at[child_id, 'hp_ever_infected_hr1'] = False
+        df.at[child_id, 'hp_ever_infected_hr2'] = False
+        df.at[child_id, 'hp_ever_infected_hr3'] = False
+        df.at[child_id, 'hp_persistent_hr1'] = False
+        df.at[child_id, 'hp_persistent_hr2'] = False
+        df.at[child_id, 'hp_persistent_hr3'] = False
+
+    def report_daly_values(self):
```
tdm32, line 559: todo - look up DALY values associated with HPV infection (if any)

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        df = population.props
+        module = self.module
+        now = self.sim.date
+
+        # 1. define eligible population
+        eligible = df.index[df.is_alive & (df.age_years >= 15)]
+        if len(eligible) == 0:
+            return
+
+        # 2. self-clearance
```
tdm32, line 631: this feels like a very inefficient way to do this, given the properties for each group, you could do a sweep of all those infected with each group for a certain length of time then determine whether they are cleared or not...?

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            new_group = set()
+
+            if sex == 'F':
+                source_prev_by_age = prev_by_age_male
+            elif sex == 'M':
+                source_prev_by_age = prev_by_age_female
+            else:
+                continue
+
+            my_age_group = module._get_age_group(df.at[person_id,'age_years'])
```
tdm32, line 713: should this be at the top of this class? You use it before this point (line 668) so it will be using the age group from 6 months ago

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+            )
+
+            # male_group_inf = df.loc[male_idx, f'hp_infected_{group}'].sum()
+            # female_group_inf = df.loc[female_idx, f'hp_infected_{group}'].sum()
+            #
+            # prev_male[group] = male_group_inf/len(male_idx) if len(male_idx) > 0 else 0
+            # prev_female[group] = female_group_inf / len(female_idx) if len(female_idx) > 0 else 0
+
+        # 4. new infection
+        interval_years = 0.5
```
tdm32: try to avoid hard-coding, if you change the frequency of the event this will no longer be correct

src/tlo/methods/hpv.py
```diff
@@ -0,0 +1,934 @@
...
+        super().__init__(module, frequency=DateOffset(months=self.repeat))
+        assert isinstance(module, HPV)
+
+    def apply(self, population):
+        # get some summary statistics
+        df = population.props
+        module = self.module
+
+        eligible = df.index[df.is_alive & (df.age_years >= 15)]
+        log_data = {
```
tdm32: the logger will automatically output the date so this is not needed

tdm32 (commented):
@Liuxinac  I've added some comments so please have a look through and let me know anything you want to discuss further. It looks great so far and when you've finished the tests, this will be good confirmation that all is working as expected.

PR #1862, opened 2026-05-15
TLO output vs TLM data: patient mix

BinglingICL:
Here we aim to compare patient mix output by TLO with TLM data, including four aspects: 
(1) patient load per HCW. --> This is the fundamental step, making sure that TLO is capturing the patient volumes appropriately. We begin with subgroup comparison, then plan to build hierarchical model using the survey data to model district and level effects for a statistically sound validation for TLO model.
(2) patient care seeking persistence
(3) patient mix across age groups, wealth levels, health conditions, and facility types. District attribute may not be considered due to limited districts selected in TLM data.
(4) patient contact time.

### DIFF
--- a/src/tlo/methods/hsi_event.py
+++ b/src/tlo/methods/hsi_event.py
@@ -276,7 +276,8 @@ class HSI_Event:
             treatment_id=self.TREATMENT_ID,
             target=self.target,
             event_name=self.__class__.__name__,
-            module = self.module
+            module = self.module,
+            appt_footprint=self.EXPECTED_APPT_FOOTPRINT,
         )
 
         # Return result in expected format:
--- a/src/scripts/calibration_analyses/analysis_scripts/analysis_patient_mix.py
+++ b/src/scripts/calibration_analyses/analysis_scripts/analysis_patient_mix.py
@@ -1,0 +1,166 @@
+"""
+This file analyses and plots the patient load and mix in service areas including outpatient (including ANC) and
+emergency care.
+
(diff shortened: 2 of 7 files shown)


BinglingICL:
Here is the latest analysis by TLO Q2 meeting 2026.
[TLO output vs TLM data - TLO Q2 2026 - Feedback.pptx](https://github.com/user-attachments/files/29376977/TLO.output.vs.TLM.data.-.TLO.Q2.2026.-.Feedback.pptx)

BinglingICL:
This branch can be put in personal TLO, no new function proposed.

src/tlo/methods/hsi_event.py
```diff
@@ -276,7 +276,8 @@ def get_consumables(
             treatment_id=self.TREATMENT_ID,
             target=self.target,
             event_name=self.__class__.__name__,
-            module = self.module
+            module = self.module,
+            appt_footprint=self.EXPECTED_APPT_FOOTPRINT,
```
BinglingICL, line 280: Hi @marghe-molaro, here (together with line 352 in cosumable.py above) I want to add the expected appt footprint of the HSI event to the Consumable logger. I have done a small local run and it is working.

marghe-molaro: Hi Bingling, the coding looks fine to me, I guess my more fundamental question is whether the expected footprint is the right information to capture. I can think of a few HSIs that will set the *actual* appt footprint to none if the consumables are not available, could it be that that's the information you want to capture?

PR #1863, opened 2026-05-22
bs/historical_hrh_growth_extended

BinglingICL:
Here we explore the health impact of historical HRH growth between 2020 and 2024, under variations of scale factors across cadres, multiple health system parameters incl. cons. availability, policy and health system performance, and the absorption rate of expanded HRH. [This is for a master student project.]

A new class of HRH scaling by year and officer type has been created for the purpose to enable a dynamic cadre-mix scaling. Initial checks of HRH growth in the scenarios using this new class have been conducted, and the results are expected as in the following figure:
image


BinglingICL:
Hi Tim @tbhallett , here is the branch I have prepared to run Izzy's job, including the scenarios discussed with her. With some small runs looking normal to me, I think it is fine to submit a full run to extract outputs for Izzy's analysis. Do you want to have a quick look before I submit? And let me know if you would prefer a short talk about this. Many thanks.

BinglingICL:
Hi Tim @tbhallett , below are the results from the last job (pop = 20_000, runs per draw = 5). The HRH growth trends are correct and DALYS and Deaths in all scenarios but the "max health system" are expected and can be explained. I have also added another scenario "Best settings assembled" that assembles the best possible parameter settings in the scenario file, which is not yet run. I would appreciate your advice on the following questions:

(1) How shall we interpret the worse results of "max health system" scenario

--> By switching to the perfect health system, the limited health resources have to provide care for both vulnerable and less vulnerable people, "crowding out" services that could lead to better health outcomes. Izzy could check the HSIs delivered to interpret this.

_TH: Not such about vulnerable/non-vulnerable people, but high-impact and low-impact services. By default, the things that do tend to happen are the most impactful things. When we turn on 'Max health system", we do "crowd out" some of those high-impact services with a deluge of lower-value services (that "should" occur too of course, but there is not enough capacity for everything.)_

_Bingling: Thanks for the helpful clarification._


(2) For the "lower bound" scenario you mentioned in Slack, I think the existing "no historical growth" will be it?

--> The "lower bound" scenario will be: Historical growth (uniform) + Cons. availability (default) + Absorption rate (0.5) + Policy (default/naive)
--> The "upper bound" scenario will be: Historical growth (uniform) + Cons. availability (maximal) + Absorption rate (default/1.0) + Policy (LCOA)


_TH: Yes, although note: (i) that these have the same cadre-mix assumption.  I would suggest that the upper one has the best cadre mix assumption and the lower one has the worst. (ii) That each of these estimate will need to be associated with  its own counterfactual wherein there is no Historical growth but the other settings remain. _ 

(3) If current results all look good, I think we will just need to run the "Best settings assembled" scenario, also with pop = 20_000 and 5 runs. No need to run all scenarios with bigger pop such as 100_000, and more runs per draw?

--> We will create a baseline/no historical growth scenario for every alternative scenario (including the lower bound and upper bound scenarios with HRH growth), so that we will be able to check the isolated effect of HRH growth under each setting.
--> For now, we set pop = 20_000, runs per draw = 5; for publication, we will set pop = 100_000 at least, and see if need to set runs per draw = 10.

_TH: yes_

Thanks very much.

Number_of_healthcare_workers
DALYs_2020-2030
DALYs_Averted_vs_Historical_growth_(uniform)_2020-2030
Deaths_2020-2030
Deaths_Averted_vs_Historical_growth_(uniform)_2020-2030

BinglingICL:
Hi Tim @tbhallett, thanks again for the helpful discussions. In the comment above, I have added the answers to my questions as discussed. Could you please check if those are what you meant?

BinglingICL:
Hi Tim @tbhallett, following our discussions, I have updated the scenario list based on Izzy's [here](https://docs.google.com/spreadsheets/d/1sC7F3Ckfcx56A2CvciIBVnwQpRtzgnFO3FTBTj31ZeU/edit?gid=373880970#gid=373880970). How they will be used to answer the proposed research questions are also explained below the scenarios. Could you please have a check and let me know any concern before I submit additional runs? Or if necessary, we could meet Izzy next Tuesday to discuss this instead and then submit the job.

tbhallett:
> Hi Tim @tbhallett, thanks again for the helpful discussions. In the comment above, I have added the answers to my questions as discussed. Could you please check if those are what you meant?

Thanks Bingling. I have commented in the above.

BinglingICL:
> > Hi Tim @tbhallett, thanks again for the helpful discussions. In the comment above, I have added the answers to my questions as discussed. Could you please check if those are what you meant?
> 
> Thanks Bingling. I have commented in the above.

Hi Tim @tbhallett, thanks very much for the helpful comments above and the great edits in the scenario list. It looks much clearer now, and I like the suggestion on comparing each alternative scenario with its sister counterfactual, so that we are investigating the impact of HRH growth under different contexts as reflected by policy/cons./hs performance settings.

I will submit the additional scenarios; the full results shall be ready when Izzy is back next Tuesday.

BinglingICL:
The overall results for [the new scenario list](https://docs.google.com/spreadsheets/d/1sC7F3Ckfcx56A2CvciIBVnwQpRtzgnFO3FTBTj31ZeU/edit?gid=373880970#gid=373880970) look good.
image
image

BinglingICL:
This branch can be put in personal TLO; the HRH scaling by year and officer type function can be included in a general scaling up function within a new HRH module for the next version of TLO.

PR #1866, opened 2026-06-19
Fix to HSI Event Queue following introduction of clinic eligibility attribute

marghe-molaro:
Clinics PR introduced error in order of attributes of `HSIEventQueueItem`, ordering events by clinic types first. This meant that `_get_events_due_today` erroneously exited the `HSI_EVENT_QUEUE` at the first not-due event for the first clinic type encountered, missing all events due on the day for other clinic types. This PR fixes this, and additionally ensures clinic type is logged when logging HSI events ran in mode 1.
Notice that this bug should not have impacted master's default configuration, where only one clinic type is considered.

### DIFF
--- a/src/tlo/methods/healthsystem.py
+++ b/src/tlo/methods/healthsystem.py
@@ -1582,7 +1582,7 @@ class HealthSystem(Module):
             rand_queue = self.hsi_event_queue_counter
 
         _new_item: HSIEventQueueItem = HSIEventQueueItem(
-            clinic_eligibility, priority, topen, rand_queue, self.hsi_event_queue_counter, tclose, hsi_event
+            priority, topen, rand_queue, self.hsi_event_queue_counter, clinic_eligibility, tclose, hsi_event
         )
 
         # Add to queue:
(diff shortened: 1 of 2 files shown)


marghe-molaro:
@tamuri, failing tests are passing locally, is there a way to resubmit them?

tamuri (approved):

marghe-molaro:
Hi @sangeetabhatia03, with this PR it would be good to tighten your clinics test to ensure that such bugs are not missed in the future.

A very easy approach is to sprinkle a few appts not due on the day on the queue for both clinics. They won't affect any of the logic of your tests, but they will ensure that we have "future" HSI events in the queue when testing. Would you be able to add this?

sangeetabhatia03:
yes absolutely, will do. While we are here, I was wondering if we should add a test for mode1 with clinics to ensure clinic specification does not interfere with mode 1.

marghe-molaro:
Thanks @sangeetabhatia03, and great idea about adding the mode 1 test!

sangeetabhatia03:
Reminder to myself to add a test for checking clinic configuration is specified via parameters.

PR #1870, opened 2026-06-29
Added weather disruption framework into TLO 

RachelMurray-Watson:
Expanded from PR #1604. 

Major: 

Added facility locations and assigned each individual to facility.
Log facility ID, monthly DALs, weather disrupted HSIs.


tamuri:
Tag for publication. Doesn't need to come into master

PR #1879, opened 2026-07-31
nfuller/add_characteristics_DALYs_and_Pop_loggers

nmfuller:
Editing the population logger in the demography module to include sex, wealth, urban/rural and region of residence.

Editing the healthburden module to record DALYs by sex, wealth, urban/rural and region of residence. This required a small edit to the demography module so that live years lost were also recorded for these characteristics.

### DIFF
--- a/src/tlo/methods/demography.py
+++ b/src/tlo/methods/demography.py
@@ -556,6 +556,9 @@ class Demography(Module):
                                                                     date_of_birth=person['date_of_birth'],
                                                                     age_range=person['age_range'],
                                                                     cause_of_death=cause,
+                                                                    urban=person['li_urban'],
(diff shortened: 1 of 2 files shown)


tbhallett:
This all looks good to me -- but we'll wait Naomi's confirmation that the runs using this worked out as we wanted,

tbhallett (approved):

tbhallett:
Merging now. I've put a little functionality in so that we can shunt this into the DEBUG level logger if needed (if this one becomes to big and unwieldy for usual outputs)

For the future, we should (more simply) allow the user to specify what they want the outputs broken down by, and just log that, rather than making all these permutations.

Issue #1872, opened 2026-07-01
Log DALYS and Deaths by district in HealthBurden Logger


RachelMurray-Watson:
As used in PR 1870 and 1749, add a logger to the HealthBurden module that logs DALYs by district.

RachelMurray-Watson:
Addressed in PR 1870, which will come into TLO V1.0

PR #1888, opened 2026-09-11
Adding Death and DALY District Loggers

thewati:
Adds Death and DALY loggers to allow extraction at district level. Fixes issue #1872


thewati:
Hi @BinglingICL and @RachelMurray-Watson ,
This is ready for review

src/tlo/methods/healthburden.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, name=None):
         self.years_life_lost_stacked_age_and_time = None
         self.years_lived_with_disability = None
         self.multi_index_for_age_and_wealth_and_time_ur = None
+        self.multi_index_for_age_and_wealth_and_time_and_district = None
```
BinglingICL: Hi @thewati, just a small question here: is this parameter redundant? It seems you've directly update the original multi_index_for_age_and_wealth_and_time_ur below. If so, you may just add the district info in the existing parameter: such as multi_index_for_age_and_wealth_and_time_and_ur_and_district?

tests/test_demography.py
```diff
@@ -156,6 +156,83 @@ def test_cause_of_death_being_registered(tmpdir, seed):
...
+    # District deaths by cause, summed across all districts.
+    district_deaths = (
+        deaths
+        .groupby(["district_of_residence", "label"])
+        .size()
+        .groupby("label")
+        .sum()
+        .sort_index()
+    )
+
```
BinglingICL: Could add a small test here to check: national deaths > 0. Just to avoid the risk that if deaths are not correctly logged by district, leading to zero death by district, which would still pass the test as "0=0".

tests/test_healthburden.py
```diff
@@ -75,6 +137,27 @@ def test_run_with_healthburden_with_dummy_diseases(tmpdir, seed):
     # Do the checks
     # correctly configured index (outputs on 31st december in each year of simulation for each age/sex group)
     dalys = output['tlo.methods.healthburden']['dalys']
+
+    # This captures national DALYs
+    dalys_stacked_by_age_and_time = output['tlo.methods.healthburden'][
+        'dalys_stacked_by_age_and_time'
+    ]
+
```
BinglingICL: Add a similar check here: total national DALYs aggregated from dalys_stacked_by_age_and_time > 0?

BinglingICL:
> Hi @BinglingICL and @RachelMurray-Watson , This is ready for review

Hi @thewati, thanks very much. This looks clear and good. The tests are quite helpful, so that we do not need to do such check in personal analysis. Just small comments. Please see if helpful.

src/tlo/methods/healthburden.py
RachelMurray-Watson: Yes I don't actually think that it's used, so it can be dropped, and I think the other attribute should be renamed

tests/test_demography.py
RachelMurray-Watson: Yes reading through I had a similar concern. I would actually consider making a new mockitis with a 100% fatality rate (or similar) so the test is more deterministic, or using the logic laid out in test_airthmetic_of_lifeyearslost to test (where someone is assigned a death).

tests/test_demography.py
```diff
@@ -156,6 +156,83 @@ def test_cause_of_death_being_registered(tmpdir, seed):
                                                     )
 
 
+@pytest.mark.slow
+def test_death_log_by_district_equals_national(tmpdir, seed):
```
RachelMurray-Watson: I'm a bit confused by this test. From what I understand, you 
(1) Take the deaths from the deaths log 
(2) Assign the death a district from the population log 
(3) Check the sum of all the district-deaths is equal to the total, nationally, in the deaths log

This feels a bit unnecessary, as it will match by construction. I am also unsure how it tests the logger that you've added, which is focussed on DALYs - I think at present the deaths don't have a district log.

RachelMurray-Watson:
Hi Wati, 

This looks great! I think the coding looks good to me, my major question is more of a conceptual one. Bingling has already noted there was some redundancy in the parameters, and given that the region of residence is a product of the district of residence, I was wondering if we could slim down the loggers by only reporting district. I am unsure how much this would gain us, realistically, but it could be a small efficiency gain. 

I know that this would "break" any current scripts that use region, so I understand if it would not be feasible. 

My second question is that, at present, I don't think deaths actually have a district logged. You could add it to data_to_log_for_each_death? 

Relatedly, but again this may go farther than what is within the PR's remit, it may be helpful to log the population per district for calculating rates.

tests/test_demography.py
BinglingICL: Good spot.

tests/test_demography.py
BinglingICL: If the national deaths and district death are from the same logger, then as Rachel suggested, this whole test is not needed.

